PR #98, opened 2014-02-19
Run drone daemon as unprivileged user

notandy:
Hi again :)

Running the drone daemon as root is not such a good idea, especially for a service that exposes a web front-end.
As far as I tested, drone works flawlessly as a normal user if it can access the docker socket (group docker) and has write access to /var/lib/drone.

But update will break setups that use a web front-end port under 1024, since drone cannot acquire such a privileged port. :/ 

I changed the default port to 8080 and adapted the readme, a reverse proxy is needed in front of drone to allow access to privileged ports (and imho heavily recommended for TLS). At very last I added apt-get --purge code for total cleanup.

Also tried to implement a "drop to unprivileged user" feature, but for now Go developer discourage [using it](https://code.google.com/p/go/issues/detail?id=1435#c19)

PS: I'am very sorry if you got some mails from my drone, completely forgot that the default .drone.yml includes your mail address :/ should not happen again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,8 @@ install:
 	cd bin && install -t /usr/local/bin drone
 	cd bin && install -t /usr/local/bin droned
 	mkdir -p /var/lib/drone
+	getent passwd drone >/dev/null || useradd -r -M -d /var/lib/drone -s /bin/false -g docker drone
+	chown drone:docker /var/lib/drone
 
 clean:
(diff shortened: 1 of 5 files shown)


bradrydzewski:
0.4 is now only distributed as a Docker image. This is more of a docker issue now, but should be solved with 1.9 and their initial user namespace support.

PR #361, opened 2014-06-23
Add docker.io Drone plugin which creates a docker image for each successful build

kvigen:
This change adds a new Drone plugin for creating a docker image. The plugin
creates a docker image for the test build and pushes the image to images.docker.io.


bmoyles0117:
Can anyone help me determine what I should be using as the default docker host in a drone.io open source environment?

PR #423, opened 2014-08-15
add remote plugin stash

reinbach:
the Stash stuct has a PrivateKey attribute that is currently set in the stash init.go file and expects there to be a "drone.pem" file in the server dir. Not pretty! But I'm not sure how you would like to handle config options for plugins. I did not want to add this field for all remotes without checking with you first.

This private key is needed for oauth, so not necessarily related to sslkey/sslcert options, as we will actually not be running our drone instance under ssl.

Also note that this stash implementation works against self hosted stash instances, so path not always stash.atlassian.com

### DIFF
--- a/plugin/remote/stash/stash.go
+++ b/plugin/remote/stash/stash.go
@@ -76,27 +76,27 @@ func (r *Stash) Authorize(w http.ResponseWriter, req *http.Request) (*model.Logi
 		return nil, err
 	}
 
-	// Stash has no API endpoint to get current user
 	// create the Stash client
-	// client := stash.New(
-	// 	s.Client,
-	// 	s.Secret,
-	// 	accessToken.Token(),
-	// 	accessToken.Secret(),
-	// )
+	var client = stash.New(
+		r.URL,
+		r.Secret,
+		accessToken.Token(),
+		accessToken.Secret(),
+		r.PrivateKey,
+	)
 
 	// get the currently authenticated Stash User
-	// user, err := client.Users.Current()
-	// if err != nil {
-	// 	return nil, err
-	// }
+	user, err := client.Users.Current()
+	if err != nil {
+		return nil, err
+	}
 
 	// put the user data in the common format
 	login := model.Login{
-		//Login:  user.User.Username,
+		Login:  user.Username,
 		Access: accessToken.Token(),
 		Secret: accessToken.Secret(),
-		//Name:   user.User.DisplayName,
+		//Name:   user.DisplayName,
 	}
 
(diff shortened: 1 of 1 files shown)


jamie-stackhouse:
@bradrydzewski, I was looking at possible solutions like building a Java plugin for Stash, and based on a comment I found on [Stack Overflow](http://stackoverflow.com/questions/26096882/atlassian-stash-how-to-get-the-identity-of-the-connected-user-via-the-rest-api) I'm wondering if this is now possible?

> You can retrieve the username of the currently authenticated user by doing a GET to the following resource (relative to the base URL of your Stash server)

> /plugins/servlet/applinks/whoami

However, looking through the Stash JIRA instance, I found a ticket [(RESTful Resource to retrieve the authenticated user)](https://jira.atlassian.com/browse/STASH-5354) by the same author.

> It would be very helpful to have a resource in the RESTful API to retrieve the currently authenticated user.
The "whoami" servlet seems only to work with a username&password session, and not with an OAuth Token.
I'm currently working on a Continuous Delivery solution that connects to Stash. And not knowing which user is authenticated is really blocking.

bradrydzewski:
@jamie-stackhouse you can look at this code and reverse engineer:
https://github.com/reinbach/go-stash/blob/master/stash/http.go#L123

The username is extracted from an auth header (X-Ausername)

preflightsiren:
@jamie-stackhouse can confirm /plugins/servlet/applinks/whoami does return a text/plain document with the current authenticated username. You can also get the username from an additional header (X-AUSERNAME)

if you need anyone to test stash, let me know :)

jamie-stackhouse:
Sorry, it looks like you guys are already well on your way to have Stash support in Drone.

@preflightsiren Sorry, it may have looked like I'm working on a different service, I wasn't looking at working on it right now, I thought this was blocked and was considering making a Stash plugin to assist what was needed to get Drone working. However, it looks like work has continued in go-stash. Also, the header appears to work even for OAuth sessions which wasn't true of the whoami servlet.

aegaas:
+1 for Stash support!

apriendeau:
@bradrydzewski what is necessary to get this merged since stash has  `/plugins/servlet/applinks/whoami` which can then use `/rest/api/1.0/users/{{username}}`. I really need an alternative to Jenkins at this point :)

josephpage:
+1 for Stash support !

suisho:
+1

ramonskie:
any news?

bradrydzewski:
Stash support added in `:latest` build

Issue #439, opened 2014-09-03
GitHub Release Plugin


rgarcia:
We've started to do more and more publishing of (mostly Go binaries) via Github releases.

Right now the process is `make release` to put a bunch of release artifacts in a folder (e.g. `release/`) and then run a script that creates a Github release and uploads the contents of this directory.

I think we could turn this into a fairly generic drone publish plugin. Configuration options:
* `script`: list of commands (or a single command?) to run to build a release
* `dir`: directory to find release artifacts after script is run
* `token`: Github token to use to interact with the Releases API.
* `version_file`: defaults to VERSION at the root of the repository
* `branch`: restrict to this branch

Example: if we were etcd and we wanted to produce their releases (https://github.com/coreos/etcd/releases) the config would look like:
```
publish:
  github:
    script: scripts/build_release
    dir: release
```

Thoughts?

bradrydzewski:
github release plugin recently added by @tboerger http://addons.drone.io/github_release/

PR #457, opened 2014-09-12
github release plugin

rgarcia:
addresses #439. Have done some manual testing over here: https://github.com/rgarcia/test/blob/master/.drone.yml e.g. https://github.com/rgarcia/test/releases/tag/0.3.7


mephux:
@rgarcia just a note - this works well for duplicates if the release is not a draft. However, if the draft flag is set it will error out when a dup is found. Just fyi.

PR #645, opened 2014-10-29
Enable to use slash and hash in brach name

muddydixon:
Purpose
------
This PR enables to use slash and hash in brach name.

Currently drone (0.3) cannot use branch name containing control chars such as slash, hash.
But we use "feature/somethingnew_#333" as branch name.


bradrydzewski:
0.4 remove branch name from the URL, so no longer necessary

muddydixon:
> 0.4 remove branch name from the URL, so no longer necessary

That's nice!

PR #647, opened 2014-10-29
add flat and flat-square badge styles

rkusa:
Shields.io has two additional styles: `flat` and `flat-square`. It would be awesome if drone supports them, too.

This PR adds them.
Usage: `/api/badge/:host/:owner/:name/status.svg?style=flat`.

I have no problem with adjusting the implementation - If you'd like to have it implemented differently. E.g. using a struct for  the `success`, `failure`, `started`, `error`, `none` map.

### DIFF
--- a/server/handler/badge.go
+++ b/server/handler/badge.go
@@ -13,13 +13,34 @@ import (
 
 // badges that indicate the current build status for a repository
 // and branch combination.
-var (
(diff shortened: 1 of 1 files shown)


rkusa:
@bradrydzewski Would you accept a similar PR (adding badge styles) for drone `0.4.0`? Just wanted to ask before I am going to create it.

bradrydzewski:
@rkusa I've got a local patch that I'd like to submit first, that stores the the files in bindata instead of inline. I think this will work a bit better as we add support for more badge styles. Once merged, yes, let's get all styles (I believe there are 3 now) included

PR #673, opened 2014-11-06
Set container user and working dir from docker inspect information

Bugagazavr:
Alternative solution for https://github.com/drone/drone/pull/671


bradrydzewski:
merged 0.4 into master. Because 0.4 plugins rely on a shared `/drone` volume across all containers, and the plugins (including git plugin) run as root we need the build container to run as root. This is a technical limitation of the 0.4 design.

PR #702, opened 2014-11-14
add --ssh-config option to drone cli

wuub:
Allows you to inject a .ssh/config file similarly to how -i allows you to insert id_rsa key. 

In our use case we need to inject .ssh/config to force `go get` to use correct ssh user when fetching/cloning dependencies.


SobiborTreblinka:
+1

bradrydzewski:
the client was moved to https://github.com/drone/drone-cli and also drone has moved to a netrc instead of ssh keys for cloning

Issue #690, opened 2014-11-11
Multiple ssh commands


nils-werner:
Current, you can only run one command in an SSH deployment scenario. This should be changed to allow a list of commands, similar to the `script` section. e.g.

    deploy:
        ssh:
            target: foo@server.de:var/www
            cmd: git pull origin master; python manage.py syncdb

Instead it would be awesome to do something like

    deploy:
        ssh:
            target: foo@server.de:var/www
            cmd: 
                - git pull origin master
                - python manage.py syncdb

PR #706, opened 2014-11-16
Refactor ssh deployment to use rsync

### LINKED ISSUE #689: deploying multiple artifacts via SSH
Deploying multiple files via SSH is annoying as all of a sudden the deploy mechanism switches to zipping the artifacts.

I think the zipping should be independent of the number of files, or removed alltogether. When one wants a zip they should create it in the `script` section themselves.

The current implementation is very unintuitive and forces users to implement the `scp` call themselves.

fudanchii:
And support multiple commands to run on the target server.

Fixes #689 and #690 

As discussed at #689, use rsync instead of scp for deploying artifacts.
Drone will first create a temp file for listing the artifacts and then use rsync's `--files-from` option for the transfer, to enable deploying multiple artifacts (possibly under different directories) at once.

To support running multiple command on the target host, drone first create a script file on the target host (named `/tmp/drone_deploy`), and running it as `sh` script.

~~To specify multiple artifacts and/or multiple commands, write them as multi-line string literals (not as list) indicated by pipe character("|"):~~
``` yml
ssh:
  target: app@example.com:/srv/.app/myapp 2222
  artifacts:
    - ./
  cmd:
    - bundle intall --deployment
    - bundle exec rake assets:clobber
    - RAKE_ENV=production bundle exec rake assets:precompile
    - sv -w 240 restart puma
```

~~yml ref: http://yaml.org/spec/1.2/spec.html#id2760844~~

### DIFF
--- a/plugin/deploy/ssh.go
+++ b/plugin/deploy/ssh.go
@@ -28,31 +28,24 @@ type SSH struct {
 	// Artifacts is a list of files/dirs to be deployed
 	// to the target host and always relative to the project's root directory,
 	// like so:
-	//   artifacts: ./
+	//   artifacts:
+	//     - ./
 	// or
-	//   artifacts: build/
-	//
-	// To list multiple artifacts, please use multiple-line string
-	// directive instead
-	// eg.
-	//    artifacts: |
-	//      bin/
-	//      include/
-	//      lib/
-	Artifacts string `yaml:"artifacts,omitempty"`
+	//   artifacts:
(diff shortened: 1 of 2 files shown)


jeffutter:
Any news on this?

igoraj:
:+1:

kri5:
:+1:

bradrydzewski:
closing now that we have an offiicial rsync plugin

Issue #562, opened 2014-10-14
Changing pwd in `script` breaks `deploy`


nils-werner:
For example a `.drone.yml` file like

    image: base
    script: 
        - cd proj
        - make

    deploy:
      ssh:
        target: machine
        artifacts:
          - proj/a.out

will not successfully deploy `proj/a.out` as `deploy` is executed after `script` without changing back to the root dir and the `artifact` it would be looking for is `proj/proj/a.out`

    script: 
        - cd proj
        - make
        - cd ..

Fixes this.

I would expect all instructions to run with respect to the root directory: Changing dir in one does not affect any of the others.

bradrydzewski:
fixed in 0.4 branch, which pushes all deploys to external docker plugins. Each plugin container is started with the workingdir set to the repository root

PR #708, opened 2014-11-16
Wrap commands in sh -c ''

nils-werner:
This pull request wraps commands in sh -c '' so each of them runs in its own context.

It implements what [I have been proposing](https://github.com/drone/drone/issues/562#issuecomment-62550047) in issue #562.

This is mainly for discussion purposes. I have simply implemented the change and fixed all tests to pass. I have no idea if it will actually work of how big the concequences will be... :-)


bradrydzewski:
with the introduction of plugins in the latest release this is no longer required. Plugins always default to the root repository directory on start

PR #710, opened 2014-11-17
Slightly more useful homepage

nils-werner:
I know you want to redesign the homepage but this is just a quick fix for now.

 - A fix for the "finished 45 years ago" message
 - All active repos are shown (I have 36 repositories to watch over, I need an overview! :-D)
 - The most recently synced inactive repos are shown. Before this I could only see some very old repos that happened to come first in the alphabet.

### DIFF
--- a/server/app/views/home.html
+++ b/server/app/views/home.html
@@ -23,10 +23,12 @@
 
 	<section ng-if="feed.length > 0">
 		<div class="pure-g cards">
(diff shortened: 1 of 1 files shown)


solarfly73:
Tested this code, works perfectly. Makes the dashboard really useful for us (around 30 repos building).

mattbostock:
This would appear to fix issue #719.

PR #711, opened 2014-11-18
instructs Drone to ignore cert verification if self-signed

bradrydzewski:


deedubs:
@bradrydzewski Would be great to see this for gogs as well! Currently running into a problem running gogs with a self-signed cert and being unable to login :(

`2015/02/11 04:01:09 Get https://*/api/v1/users/deedubs/tokens: x509: certificate signed by unknown authority

Issue #677, opened 2014-11-07
Incremental build number


floatdrop:
I don't found any issues about `CI_BUILD_NUMBER` variable. Now [it contains](https://github.com/drone/drone/blob/master/shared/build/build.go#L504) commit hash, that triggered build.

This approach not working for building debian packages, which should have incremental verison `x.x.x-buildnumber` to be autoinstalled without warning (about downgrade).

Is there workarounds for incremental build number in drone or should it be implemented by hands?

floatdrop:
Resolved by https://github.com/drone/drone/pull/858, thanks @epipho :clap: !

PR #737, opened 2014-11-30
All repositories now have a monotonically increasing build number.

epipho:
This is for #677 

All repos now store a monotonically increasing build number. 
The build number assigned to a commit is stored so it is stable across rebuilds.
Changed CI_BUILD_NUMBER to expose the repo's build number
Added DRONE_BUILD_NUMBER for parity


mattbostock:
@bradrydzewski Can the build for this PR be restarted? I'd like to use this feature.

craigtracey:
:+1: we would also like to use this.

shared/model/repo.go
```diff
@@ -56,6 +58,7 @@ func NewRepo(remote, owner, name string) (*Repo, error) {
 	repo.PostCommit = true
 	repo.PullRequest = true
 	repo.Timeout = DefaultTimeout
+	repo.BuildNumber = 0
```
bradrydzewski: default value for a primitive integer is zero, so no need to specify here

server/worker/docker/docker.go
```diff
@@ -62,6 +62,14 @@ func (d *Docker) Do(c context.Context, r *worker.Work) {
 	// mark the build as Started and update the database
 	r.Commit.Status = model.StatusStarted
 	r.Commit.Started = time.Now().UTC().Unix()
+
+	// Update the sequential build number if it hasn't been set yet
+	if r.Commit.BuildNumber == 0 {
```
bradrydzewski: I think the ability to increment should live in the `database.PostCommit` function instead.

server/datastore/migrate/migrate.go
```diff
@@ -110,6 +127,13 @@ var repoTokenUpdate = `
 UPDATE repos SET repo_token = '';
 `
 
+var repoBuildNumberColumn = `
+ALTER TABLE repos ADD COLUMN repo_build_number INTEGER
```
bradrydzewski: I would prefer to calculate this value on-the-fly, instead of storing a reference in the repository table. Let's instead use `select max(build_number) from commits where repo_id = ?`. This is what we've done with the drone.io service (different codebase) which has worked really well

bradrydzewski:
out of curiosity, is the primary use case for this to tag artifacts (ie zip files, binaries, websites, etc) with a build version number? I'm wondering if we even need to store this number in the database. We could just `select count(*) from commits where repo_id=? and commit_id>?` at runtime and pass into the container and plugins

server/datastore/migrate/migrate.go
epipho: Is there any chance that two builds from the same repository could be running at the same time? If so, two builds could receive the same build number. Unfortunately I believe my current solution also has this issue.

epipho:
Other than the multiple simultaneous build issue noted above, I stored the version used along with the commit in order to facilitate rebuilds. Triggering a rebuild of the same commit should generate the same version number, even if more commits have been built since.

shared/model/repo.go
epipho: Old habit, fixed.

epipho:
I merged the current master into this change, but have not yet done any testing beyond automated tests.

craigtracey:
@bradrydzewski counting commits assumes that the trigger for a build is always a commit.  In the case of a rebuild (ie. an external dependency changes) you end up with the same version, but quite possibly differing bits in the resulting artifact. Every run, in my opinion, whether rebuild or not, should have a unique identifier.

epipho:
https://github.com/epipho/drone/pull/1 is an addition that ensures two commits can never have the same build number by serializing the increment and get in the database.  This also allows the code to be moved to PostCommit easily.

@craigtracey Currently it is set up specifically to PRESERVE the build number in the case of a rebuild. If a new number should be used for every build (even rebuilds) there is no reason to store the build number on the commit at all.

With the PR I linked it is easy to do either one, @bradrydzewski thoughts?

bradrydzewski:
@craigtracey the `select count(*) from commits where commit_id <= ? and repo_id = ?` would always return the same result, even for rebuilds. A rebuild updates the existing record in the database, and does not insert a new record. I believe this is the only implementation that is safe to concurrency.

I believe we could therefore get the build number on-demand (with a count), and inject into the container as an environment variable. I'm not sure we have a need (yet) to store this field in the database. I'd rather take incremental steps here.

epipho:
Unfortunately I am fully booked this weekend, but I will try to revisit this again in a few days w/ all feedback from above.

epipho:
Created a new version of this PR at #858 to preserve the work done here, it is ready for review.

bradrydzewski:
just merged #858 into master. The `$DRONE_BUILD_NUMBER` environment variable will now be available to your build environment.

PR #764, opened 2014-12-19
Add an option to disable recursive clone

msteinert:
This change adds the following configuration item to .drone.yml:

    git:
      recursive: false

If the item is left out then the default is `true`, i.e. perform a recursive clone.

I couldn't find where to add the documentation for the new configuration (granted I didn't do an exhaustive search). If you point it out I can update that.

I'm not even sure if you'll be interested in this change, however it was quite useful for me today!


vtolstov:
as of 0.4.0 version this is already implemented, may be close this ?

msteinert:
Sounds good to me.

PR #773, opened 2014-12-26
Allowing Binds Volumes to be Specified from .drone.yml

husobee:
On occasion, I have found the need to create a volume in the build container to the container host.  This is different from cache in that this use-case doesn't involve persisting between builds, but rather just having access to something on the host file system in the build container.  Primary use case i have found is to be able to build docker by docker using a file socket.  

I suppose if there was a security concern, we could add a configuration option to explicitly state which volumes were allowable volumes to bind.


bradrydzewski:
@husobee could we tweak this a bit to specify volumes in the docker section of the yaml file:
https://github.com/drone/drone/blob/master/shared/build/script/docker.go#L9

something like this:

```
image: go:1.3
docker:
  volumes:
    - /path/on/host:/path/in/container
```

also let's give flexibility to specify the host path and container path, in case someone wants them to be different:

```diff
+ for _, volume := range b.Build.Volumes {
-		host.Binds = append(host.Binds, fmt.Sprintf("%s:%s", volume, volume))
+		host.Binds = append(host.Binds, volume)
+		conf.Volumes[volume] = struct{}{}
+	}
```

Otherwise I think this is a great change and would love to get it merged.

husobee:
@bradrydzewski , made the suggested alterations, and added some coverage unit tests for this functionality, as well as merged master into this fork.  Let me know if there is anything else I can do to get this merged.  Thanks!

shared/build/script/docker.go
```diff
@@ -37,3 +44,31 @@ func DockerHostname(d *Docker) string {
 	}
 	return *d.Hostname
 }
+
+// DockerVolumes returns empty []string
+// when Docker.Volumes is empty.
+func DockerVolumes(d *Docker) []string {
```
bradrydzewski: we can remove this method and just access d.Volumes directly. This is because `d` will never be nil, and `d.Volumes` will never be nil, and if `len(d.Volumes) == 0` it is the same as returning an empty string slice :)

bradrydzewski:
we also need to make 1 more alteration. This technically creates a security hole that can be exploited, since it exposes the host machines file system. Someone could (for example) send me a pull request with an updated `.drone.yml` that mounts the host machine file system granting them access to sensitive data (like my drone sqlite database, or my drone config file with auth keys and such). I need to think through this a bit more before I merge, but I have some ideas.

shared/build/script/docker.go
husobee: +1 on nil slice returning 0 on len function.  Wasn't thinking.  In running the unit tests i ran into an issue where b.Docker was nil, which is what made me make that method, just like the d.HostName method.  Originally I was just going to access it directly until unit tests started failing.

husobee:
I completely agree on needing to address security ramifications.  I would suggest a white-list in the drone server configuration for allowable bind volumes.  Happy to hear your suggestions.  Let me know the direction you want to go and I will be happy to throw together a solution.

stefanv:
What is the current progress on this issue, and is there anything I can do to help push it forward?

bradrydzewski:
this is being actively worked on as part of #902 which is a re-design of the yaml format to expose more docker functionality, including volumes

husobee:
Awesome.  Very glad we are adding volume capabilities.  If there is anything I can do to assist, please let me know!

dennybaa:
@channel +1 to this feature.
It would be also nice to see not only host volumes but an ability to address -volumes-from option. I haven't seen that anyone's mentioned it.

The idea behind volumes-from is basically the same, but it allows to bind volumes from another container such as service container. It's useful when "service" container prepares intermediate data which is later used by the CI container. Here's reference https://docs.docker.com/compose/yml/#volumes-from

bradrydzewski:
this is now enabled in the 0.4 branch. check out the "build options" section in the docs http://readme.drone.io/docs/build/build.html

stefanv:
Thank you very much @bradrydzewski

dennybaa:
@bradrydzewski super, thank you!

PR #783, opened 2015-01-06
Enforce word breaks in terminal output

ap4y:
Fixes terminal output weirdness in `Firefox` with long lines in terminal output (`rspec` dot progress formatter for example).


edgurgel:
:+1:

ap4y:
I think css have to be recompiled from the `less`, I don't see it happen on ci machine (yet it install `lessc`). Checked with the latest `deb` package.

PR #785, opened 2015-01-06
Add support for Mercurial repositories (on Bitbucket at least)

faide:
I am open to comments and suggestions since I know the way I did it is suboptimal ie: if isGit returns False defaults to mercurial... which makes it impossible at the moment to add more VCSs


cedk:
I think Repo struct could have a Scm attribute which will be filled by remotes instead of guessing.

cedk:
I created #793 that adds and uses a scm on repo.

bradrydzewski:
I've also been kicking around the idea of letting the plugin define the shell commands (ie `git clone`). We could then add a plugin for almost any service, using any scm. We could even add a plugin for something like Dropbox (not that we would want to).

cedk:
I don't think it is a so good idea to have plugin defines the shell commands because we will loose some factoring of the code. There are many hosting services but not so many SCM's. Also as if you add a know plugin it should anyway modify the code, you can also modify the build of repo.

bradrydzewski:
@cedk one thing to consider is that the method for cloning pull requests differs depending on which hosting solution is used

cedk:
@bradrydzewski until now with the supported hosting, there is no differences. But I guess there could be some methods defined on Remote to retrieve the SCM commands or the some kind of flavor.
But could you take a look at #793 if it is a good start?

PR #788, opened 2015-01-07
Always pull newest base image when using docker publish plugin

gregory90:
Without this patch if base image was rebuilt with the same tag as before, drone wouldn't pick it up in docker build process in publish plugin.


azylman:
After this change, the docker publish plugin fails if the `docker_version` is set older than 1.4:

```
flag provided but not defined: --pull
```

Might be worth documenting this new requirement for the `docker_version`?

PR #794, opened 2015-01-11
Pin NPM packages using `npm shrinkwrap` for repeatable builds

mattbostock:
Pin NPM packages with `npm shrinkwrap --dev`.

Adds a minimal `package.json` created by running `npm init` and removing some lines; just enough to satistify `npm install` without it showing warnings. Also, run `npm install` each time `make deps` is invoked.

Note that there are changes to the generated `drone.css` when using Less version 2.2 (the latest) though in my testing I couldn't see any breakages.

I'm not very familiar with NPM or Less, so I'd appreciate any feedback.


Bugagazavr:
i think file `server/app/styles/drone.css` can be add to `.gitignore`

mattbostock:
@Bugagazavr I'm inclined to agree, since it's a generated file.

I think the `Makefile` would need changing, as we wouldn't want people to build drone without the CSS files. Perhaps the `deps` and `build` tasks could be merged into one? Or at least, `deps` would probably be part of the default talk.

@bradrydzewski, what are your thoughts?

bradrydzewski:
Since we only have lessc as a dependency I'd prefer to keep it as a global dependency requirement (autoprefixer is being removed in an upcoming PR, and we can use `lessc -x` instead of `lessc clean-css`). If we start to depend more heavily on node this is something I would consider.

@Bugagazavr since the majority of developers are contributing Go code I don't want them to have to install npm or node to test locally. My intention when checking in the CSS file was to minimize barrier to entry for Go developers that wanted to contribute to the project.

PR #796, opened 2015-01-11
Remove explicit `go get` for rice and goblin pkgs

mattbostock:
Both of these packages will be pulled in by the `go get` invocation on the following line as they are references by imports, so there's no need to list them separately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,6 @@ all: build
 deps:
 	# which npm && npm -g install uglify-js less autoprefixer
 	go get github.com/GeertJohan/go.rice/rice
-	go get github.com/franela/goblin
 	go get -t -v ./...
 
 test:


bradrydzewski:
Yes, we can definitely remove the reference to goblin. We will still need to `go get` or `go install` the `github.com/GeertJohan/go.rice/rice` package since it is a binary and needs to be installed to `$GOPATH/bin`

bradrydzewski:
we might be able to do something like this:

```
go get -t -v ./...
go install go get github.com/GeertJohan/go.rice/rice
```

the rice repository would get cloned during the first `go get` and then the binary would be installed with the subsequent command.

mattbostock:
Ah, my bad - I saw the import [here](https://github.com/drone/drone/blob/6ed8542f7d54027519f06aa794583f78244b0bb1/server/main.go#L17) but that wouldn't include the `rice` binary.

I think the following would do the trick:

```go
go get github.com/GeertJohan/go.rice/rice
go get -t -v ./...
```

mattbostock:
I've rebased to fix.

PR #797, opened 2015-01-11
Run gofmt and add test to prevent regressions

mattbostock:
Run `go fmt ./...` [1] and add a test to the `test` make target that checks if `go fmt` has been run.

[1]: http://blog.golang.org/go-fmt-your-code


mattbostock:
Looking into the build failure.

Bugagazavr:
because `go fmt -n ./...` returns list of packages with files, even if these files if normal.

bradrydzewski:
I really like the idea. What about something like `gofmt -l ./.. | read`?

Bugagazavr:
@bradrydzewski yes it works
`test -z "$(gofmt -l ./.. | grep "../drone/")" || echo "Need to run 'go fmt ./...'"; exit 1`

mattbostock:
@Bugagazavr Yep, that was the problem.

I've just pushed an amendment, let me know what you think.

bradrydzewski:
awesome, thanks!

PR #800, opened 2015-01-12
Add Gogs to the config file and Readme

dave-tucker:
While updating an Ansible module I noticed that while Gogs is supported,
it doesn't appear, commented out, in the TOML file or in the Readme.
This commit adds it

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>


bradrydzewski:
thanks!

PR #802, opened 2015-01-12
Bintray

Bugagazavr:
This is bintray plugin

```yaml
publish:
  bintray:
    username: $BINTRAY_USER
    api_key: $BINTRAY_TOKEN
    packages:
      - file: packaging/output/drone.deb
        owner: awesomeorg
        type: deb
        repository: deb
        package: drone
        version: 0.3.0-dev
        target: drone_0.3.0-dev.deb
        distr: main
        component: main
        arch: 
          - amd64
        publish: true
        override: true
      - file: packaging/output/drone.rpm
        owner: awesomeorg
        type: rpm
        repository: rpm
        package: drone
        version: 0.3.0-dev
        target: drone_0.3.0-dev.rpm
        publish: true
        override: true
```

**type** can be a `deb`, `rpm`, `mvn`, if not specifed used simple upload like `rpm` and `mvn`
debian packages requires **distr**, **component**, **arch**

For upload used this syntax: 
```
https://api.bintray.com/content/:owner/:repository/:package/:version/:target
```
To this
```
https://api.bintray.com/content/awesomeorg/deb/drone/0.3.0-dev/drone_0.3.0-dev.deb
```

About `publish`, `override`:

- **publish** - by default uploaded package doesn't published, if you need publish on upload you can set `publish: true`
- **override** - if we already publish `drone_0.3.0-dev.deb`, we can't upload `drone_0.3.0-dev.deb` second time, we need to enable **override**, to allow overwrite `drone_0.3.0-dev.deb`


Bugagazavr:
Also you can test my test repo
```bash
curl -s https://bintray.com/user/downloadSubjectPublicKey?username=awesomeorg | apt-key add -
echo "deb http://dl.bintray.com/awesomeorg/deb main main" | sudo tee -a /etc/apt/sources.list.d/drone.list
sudo apt-get update
sudo apt-get install drone
```

Bugagazavr:
@bradrydzewski what you think about delete `dev` from version, drone is not stable and not rc.

bradrydzewski:
@Bugagazavr I would consider Drone alpha, so perhaps we could do something like this: `0.3.0-alpha`

Bugagazavr:
@bradrydzewski I'm just worried that because of the letters will be a problem with certain versions of package managers. As variant i suggest use semantic versions, where major release < 1 is unstable/alpha.

Bugagazavr:
@bradrydzewski ok

PR #803, opened 2015-01-12
Fixing an issue with branch grouping due to substring matching

andy-g:
I noticed that if one branches name exists within another branch, its commits are included in both branches, this is because the AngularJS "filter" filter defaults to a substring match rather than a strict equality comparison. Simply specifying the comparator explicitly resolves the issue. 

https://docs.angularjs.org/api/ng/filter/filter


mattbostock:
:+1:

bradrydzewski:
awesome! that was bothering me as well

PR #814, opened 2015-01-14
[Fixes #660] Display login errors to clients

### LINKED ISSUE #660: Display login failure messages
When using Drone with GitLab and using invalid credentials, the page shows a white screen.
It would be helpful for the end user if there was an error message shown.

dcarley:
So that it's clear to an end user (particularly a human, rather than an API
client) why a login failed. I've applied this all places where we return a
non-200 response code from `/login`.

---

I don't think this would expose anything sensitive that the client doesn't already know?

### DIFF
--- a/server/handler/login.go
+++ b/server/handler/login.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"encoding/json"
+	"fmt"
 	"log"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


dcarley:
Re-ran build which has failed on `go get`.

bradrydzewski:
@dcarley I appreciate you looking into this particular issue. We have a release coming out on the 27th and were planning to more closely mirror Google's client-side OAuth flow [1]. We would redirect to the login screen with "/login#error={error}" so that we can render a proper error message in the UI

Let me know if this is something you're interested in.

[1] https://developers.google.com/accounts/docs/OAuth2UserAgent

dcarley:
No problem. That's probably beyond my comprehension at the moment, so I'll leave you to it.

Issue #749, opened 2014-12-09
RFC - Simplify build process


bradrydzewski:
EDIT: here is the status of the 8 PRs described below:

- [x] PR1
- [x] PR2
- [ ] PR3
- [ ] PR4
- [ ] PR5
- [ ] PR6
- [ ] PR7
- [ ] PR8

---

I want to change how we do builds. Today we do something like this:

* generate build script, save to disk
* generate ssh keys, save to disk
* generate dockerfile with env vars, save to disk
* create image
* start container that runs build script in container
* delete container
* delete build script

I would like to simplify:

* start container, injecting the build script as a command line arg to the container
* delete container

To do this we can pass the build script into the Docker like this:

```
docker run {image} /bin/bash -c "echo {script} | base64 --decode | tee run.sh | sh run.sh && rm run.sh"`
```

The build script is base64 encoded in the command line argument. The build script is then removed once the build is complete.

This is a big change, and since there have already been a lot of big changes, I would like to split this into a series of PRs to minimize impact to the stability of the project. Having multiple PRs will let us test each change in isolation and make it easier to troubleshoot issues. It will also make code reviews easier :)

### PR 1
the build script should be responsible for writing the ssh key to the filesystem instead of ADD in the Dockerfile. See the following code:
https://github.com/drone/drone/blob/master/shared/build/build.go#L561

Below is an example of what the bash function might look like.  Note that Docker may set $HOME=/ by default. We may need to query the user home directory. I'm not entirely sure though ... needs testing

```sh
[ -d $HOME/.ssh ] && mkdir -p $HOME/.ssh
chmod 700 $HOME/.ssh
cat %q > $HOME/.ssh/id_rsa
chmod 600 $HOME/.ssh/id_rsa
```

### PR 2
the build script should export all environment variables, instead of using the Dockerfile. See:
https://github.com/drone/drone/blob/master/shared/build/build.go#L489

### PR 3
the build script should no longer user `/var/cache/drone`. It should instead use the default `WORKINGDIR`. We will also need to account for `$GOPATH`. We also need to account for the fact that the git clone path can be specified in the `.drone.yml`

here is what I'm thinking:

```sh
export DRONEPATH = $HOME # or git clone directory in yaml, if set
[ -v $GOPATH ] && DRONEPATH = $GOPATH/src
cd $DRONEPATH

# ... rest of build script. git clone, etc
```

### PR 4
the build script should use the default USER. It should never attempt to set or override the USER

### PR 5
[...]

Issue #483, opened 2014-09-25
Support for docker "official" unprefixed images


txase:
Drone overrides Docker unprefixed "official" images used for services and the base image (ubuntu, node, etc.). In some cases, I would rather use the unprefixed Docker images, but I can't find a way to do it in drone.

bradrydzewski:
fixed in 0.4 branch with deprecation / removal of official drone images

PR #816, opened 2015-01-14
Configurable repository path

stevenpall:
Having the repository being built checked out to a specific directory would be useful in various cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ test_postgres:
 build:
 	mkdir -p packaging/output
 	mkdir -p packaging/root/usr/local/bin
-	go build -o packaging/root/usr/local/bin/drone  -ldflags "-X main.revision $(SHA)" github.com/drone/drone/cli
-	go build -o packaging/root/usr/local/bin/droned -ldflags "-X main.revision $(SHA)" github.com/drone/drone/server
+	go build -o packaging/root/usr/local/bin/drone  -ldflags "-X main.revision $(SHA)" github.com/stevenpall/drone/cli
+	go build -o packaging/root/usr/local/bin/droned -ldflags "-X main.revision $(SHA)" github.com/stevenpall/drone/server
 
 install:
 	install -t /usr/local/bin packaging/root/usr/local/bin/drone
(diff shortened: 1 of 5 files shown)


bradrydzewski:
@stevenpall there is partial support for this feature:

```
git:
  path: path/to/clone
```

This clones the repository into `/var/cache/drone/src/{ git.path }`. This could be adjusted to work with absolute paths (that start with `/`). If is probably just a one or two line change here:
https://github.com/drone/drone/blob/0a6227930dce3e356667506bc950530a9d847c61/server/worker/docker/docker.go#L119

bradrydzewski:
@stevenpall I think the path forward here is to alter the following configuration to respect absolute paths in the `.drone.yml` file:

```
git:
  path: path/to/clone
```

It could look something like this:

```diff
path := r.Repo.Host + "/" + r.Repo.Owner + "/" + r.Repo.Name
	repo := &repo.Repo{
		Name:   path,
		Path:   r.Repo.CloneURL,
		Branch: r.Commit.Branch,
		Commit: r.Commit.Sha,
		PR:     r.Commit.PullRequest,
-		Dir:    filepath.Join("/var/cache/drone/src", git.GitPath(script.Git, path)),
		Depth:  git.GitDepth(script.Git),
	}

+ if len(script.Git) != 0 && filepath.IsAbs(script.Git) {
+ 	repo.Dir = git.GitPath(script.Git, path)
+ } else {
+ 	repo.Dir = filepath.Join("/var/cache/drone/src", git.GitPath(script.Git, path))
+ }
```

I do agree the current default directory (`/var/cache/drone`) is not ideal. This is one of the reasons we'll switch to use the default `WORKDIR` as defined in the Dockerfile (detailed in #749). This is important for compatibility with the official Docker language stacks (see #483).

stevenpall:
@bradrydzewski That makes sense, and I like the idea of using `WORKDIR` to define the clone path. What do you foresee the timeline being on PR 3 of #749 getting merged in?

As an aside, I was having a pretty hard time testing the changes made in my fork since a lot of the Makefile has [github.com/drone/drone](github.com/drone/drone) hardcoded. What is the preferred method of building and running Drone from a fork like this for testing?

PR #820, opened 2015-01-15
Support org whitelists for GitHub+GHE remotes

dcarley:
Allow the GitHub and GitHub Enterprise remotes to restrict who can login
based on a user's organisation membership. This can be used as a safe
addition to open registration and also ensures that access is revoked when a
user is subsequently removed from the org. The default is not to restrict at
all.

---

Plus some supporting tests..

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,12 +100,14 @@ open=true
 [github]
 client=""
 secret=""
+orgs=[]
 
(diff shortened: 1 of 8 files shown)


bradrydzewski:
looks great, thanks for the contribution

Issue #761, opened 2014-12-16
Bitbucket integration should use the slug to build URLs, not the name of the repo


orjanbruland:
Hi!

Currently the bitbucket integration uses the name of the repo to build up the clone, git and SSH urls. This breaks if the name of the repo includes whitespace or special characters since Bitbucket generates a slug from the name of the repo to build the URLs. Using the slug returned from the Bitbucket API instead would solve this.

bradrydzewski:
this was fixed by @gtaylor

PR #824, opened 2015-01-16
Use Slug instead of Name for Bitbucket repos.

orjanbruland:
From #761 

This seems to solve the issue.


bradrydzewski:
great, thanks!

Issue #805, opened 2015-01-12
Ability to enable open registration on a per-plugin basis


mattbostock:
We use both GitHub.com and a private instance of GitHub Enterprise.

For our GHE users, we would be happy to allow open registration, i.e. anyone with a GHE account can log in. For GitHub.com however, we would need to restrict which users can use Drone.

Currently open registration is a global option affecting all remote plugins. I'd like to be able to enable/disable open registration for specific remotes.

bradrydzewski:
Agreed. I'd like to move this setting to the plugin level and remove as a global plugin.

mattbostock:
Fixed in https://github.com/drone/drone/pull/826

PR #826, opened 2015-01-16
Per-remote open registration

mattbostock:
...so that it's possible to enable or disable open registration on a
per-remote basis.

For example, the `DRONE_REGISTRATION_OPEN` environment variable now
becomes `DRONE_GITHUB_OPEN` when using GitHub as a remote.

The default for open registration in this commit is `false` (disabled),
which matches the existing behaviour.

This is useful if you need to support both public and private remotes,
e.g. GitHub.com and GitHub Enterprise, where you trust all of the
private users and want to allow open registration for those but would
not want all GitHub.com users to run builds on your server.

Tested with GitHub and GitLab.

* * *

Also, remove the capability package as it's no longer required. This commit can always be reverted later if it's needed again in future.


mattbostock:
Fixes #805.

bradrydzewski:
this is really great. much appreciated.

dcarley:
:+1:

PR #793, opened 2015-01-11
Add support of mercurial for bitbucket

cedk:
Following my comment on #785, this version adds a scm field on repo filled by the remote plugins.

PR #829, opened 2015-01-18
SCM plugins

Bugagazavr:
This is rework pull requests https://github.com/drone/drone/pull/793 and https://github.com/drone/drone/pull/785

This PR allow move scm logic outside build package, to create a simple scm implementations inside drone.

Also has been added scm "local", because when docker add files from root user, non-root users doesn't have write access to this files. For now, files with source adds to `/tmp/drone_cache` and then we use rsync to copy files inside build directory `rsync -rt /tmp/drone_cache/ /var/cache/drone/src/github.com/drone/drone`. Local scm needed for https://github.com/drone/drone/issues/749#issuecomment-67302820 if we want to set `$DRONEPATH` when we use drone cli

### DIFF
--- a/cli/build.go
+++ b/cli/build.go
@@ -13,6 +13,8 @@ import (
 	"github.com/drone/drone/shared/build/repo"
 	"github.com/drone/drone/shared/build/script"
 
+	"github.com/drone/drone/plugin/scm/local"
+
 	"github.com/codegangsta/cli"
 )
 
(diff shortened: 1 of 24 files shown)


bradrydzewski:
@Bugagazavr have a look at https://github.com/drone/drone/pull/785#issuecomment-69507212

I've been thinking that the remote plugin should be responsible for determining the clone commands. This allows us to easily handle, for example, different pull request logic per remote. The allows the remote to query for extra data required to clone the PR instead of storing extra fields in the database (like source repo and source branch required for GitLab PR testing).

We probably wouldn't need to store scm in the database. We can look at the Bitbucket SSH URL to determine Git vs Mercurial. If we added a Google Code remote it could lookup the VCS using the same method as Go's `go get` command.

It is definitely something I'd like to explore.

cedk:
I don't think all hosting services can be managed by one set of command per SCM especially for Pull Request. For me, each remote plugin must define its own commands like I did in #871

PR #832, opened 2015-01-20
Make live updates to the commits (repo controller) without needing to reload

andy-g:
On Build completions (Failure or success), and on new Build starts, this updates the build list without requiring a reload, rather than just displaying the build notification message.

Successes, failures and rebuilds will update existing records, and new builds will add a new build item.


bradrydzewski:
@andy-g I definitely agree that we should update the status of existing items on the page (and would merge that piece of code).

We did make the explicit decision not to automatically add new items to the page. The reason is that it can interrupt someone who is reading the page or attempting to click an item. This is especially true for larger teams that have high commit volume. We therefore chose to ask users to manually refresh, similar to Google Plus or Twitter.

andy-g:
Perfect, I'll tweak the PR to remove the adding of new items. Should I also only display the existing message on new items and hide it if updating an existing item?

bradrydzewski:
> Should I also only display the existing message on new items and hide it if updating an existing item?

great question. I'm 50/50 so go with your gut

andy-g:
I can't see how much commit could have broken the build. Are you able to retry the build, or check if it's environmental?

bradrydzewski:
I recently updated the `.drone.yml` file. I included `apt-get update` because ubuntu updated its package index and the local cache (in the container image) is out of date and was failing the build. Can you merge with upstream to get the new yaml?

andy-g:
I've merged with upstream and the build seems happier.

bradrydzewski:
awesome, didn't even know that you could use the `some` function on an angular array. THanks for the update

PR #834, opened 2015-01-20
npm publish only when version not exist

### LINKED ISSUE #799: npm publish plugin, only publish if version not exists
The proposed change is to check if the npm package/version have already been published. We can add something like this to check:

```sh
test -z "$(npm info bower@1.3.12)"
if [ $? -eq 0 ]
then
       npm publish bower
else
        echo "skipping publish, package bower@1.3.12 already published"
fi
```

Note this is just an example. We would need to substitute `bower` and `1.3.2` for injected values. We would also need a way to determine the current package version. I'm guessing we can get this with a command somehow.

/cc @donny-dont

Anything else I'm missing?

shawnzhu:
fix for #799

### DIFF
--- a/plugin/publish/npm/npm.go
+++ b/plugin/publish/npm/npm.go
@@ -18,8 +18,22 @@ email = %s
 EOF
 `
 
+// command to publish npm package if not published
+const CmdPublish = `
+_NPM_PACKAGE_NAME=$(cd %s && npm list | head -n 1 | cut -d ' ' -f1)
+_NPM_PACKAGE_TAG="%s"
+if [ -z "$(npm info ${_NPM_PACKAGE_NAME})" ]
+then
+	npm publish %s
+	[ -n ${_NPM_PACKAGE_TAG} ] && npm tag ${_NPM_PACKAGE_NAME} ${_NPM_PACKAGE_TAG}
+else
+	echo "skipping publish, package ${_NPM_PACKAGE_NAME} already published"
+fi
+unset _NPM_PACKAGE_NAME
+unset _NPM_PACKAGE_TAG
+`
+
 const (
-	CmdPublish     = "npm publish %s"
 	CmdAlwaysAuth  = "npm set always-auth true"
 	CmdSetRegistry = "npm config set registry %s"
 )
@@ -42,10 +56,6 @@ type NPM struct {
 	// The Password used by NPM to connect
 	// and publish to a repository
 	Password string `yaml:"password,omitempty"`
-
-	// Fails if the package name and version combination already
-	// exists in the registry. Overwrites when the "--force" flag is set.
-	Force bool `yaml:"force"`
 
 	// The registry URL of custom npm repository
 	Registry string `yaml:"registry,omitempty"`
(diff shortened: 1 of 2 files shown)


shawnzhu:
fix for #799

donny-dont:
@shawnzhu awesome! This is really appreciated.

I was wondering if you were trying to publish directly to the global npm server or if you had something like https://github.com/rlidwka/sinopia for a private npm server. At work we're using Sinopia so I just wanted to make sure it would work with it.

Thanks again! Great work!

shawnzhu:
It supports specifying registry URL in .drone.yml so it works when
publishing to either nodejitisu or any other npm repo supports npm login.

However, it doesn't support publishing scoped packages. So this maybe
another PR

bradrydzewski:
@shawnzhu much thanks!

@donny-dont would you mind testing this out? Reply to this thread and let us know how it goes

donny-dont:
So one bug is if the package was never in the registry before.

```
npm ERR! 404 no such package available : XXXXXX
npm ERR! 404 
npm ERR! 404 'XXXXXX' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

donny-dont:
Hmmm it doesn't seem like the username/password is in there. Not sure if somethings up with our network but this is what I'm seeing.

```
$ npm config set registry http://xxxx:4873/
$ npm set always-auth true
npm ERR! publish Failed PUT 403
npm ERR! Linux 3.13.0-43-generic
npm ERR! argv "node" "/usr/local/bin/npm" "publish" "."
npm ERR! node v0.10.35
npm ERR! npm  v2.1.18
npm ERR! code E403

npm ERR! can't publish restricted package, bad username/password, access denied : XXXXXX
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <http://github.com/npm/npm/issues>
```

@shawnzhu @bradrydzewski any thoughts on what might be happening?

shawnzhu:
@donny-dont it should allow publishing when the npm package doesn't exists in registry. will fix it soon.

donny-dont:
Sounds good @shawnzhu 

Any tips on how to debug if the username/password is being set by the plugin? Might be a good idea to have it echo when it is doing that.

I added this to the .drone.yml
```
publish:
  npm:
    username: 'drone'
    password: 'xxxxx'
    email: 'drone-do-not-reply@xxxx.com'
    registry: 'http://xxxxx:4873/'
    always_auth: true
```

bradrydzewski:
merged the fix from @shawnzhu when package does not yet exist

shawnzhu:
@donny-dont I've tested this plugin to publish a package to sinopia, and I can confirm it sent the username/password via basic authentication in the PUT request (actually it works against npmjs.org). you can debug by dumping HTTP traffic to sinopia from drone host via command `sudo tcpdump -vvvs 1024 -l -A dst port 4873` and you should see the `authorization` header in the HTTP request.

However, I keep getting status 403 with error message:
```shell
npm http PUT http://192.168.59.103:4873/characteristic1
npm http 403 http://192.168.59.103:4873/characteristic1
npm ERR! publish Failed PUT response 403
npm ERR! Error: user shawnzhu not allowed to publish it : characteristic1
```

But this error message indicates that sinopia can parse my username from the `authorization` header which means this plugin sent it out as designed.

donny-dont:
Thanks for all the help @shawnzhu I'll work with our IT and see if something is going wrong on the network side.

donny-dont:
We had a configuration issue on the network side. Its been fixed and I can confirm the patch from @shawnzhu works perfectly as advertised. Thanks again for this its been a huge help!

PR #841, opened 2015-01-21
Fix services behaviour when docker net is host

### LINKED ISSUE #839: Services do not operate correctly when docker net host option is specified
The addition of `--net` in #617 is awesome, I need it, but I may have found a bug - I don't have bandwidth to test this one out at the moment though...

The implementation of `services` looks to rely on Docker Links 
https://github.com/drone/drone/blob/0a6227930dce3e356667506bc950530a9d847c61/shared/build/build.go#L346

Links and `--net=host` are not supported https://github.com/docker/docker/pull/7066

A possible workaround might be to run service containers in `--net=host` but you might run in to port conflict issues if you are, for example, using postgres as the backend for your Drone install and attempting to run a postgres service.

dave-tucker:
When the image runs with `--net host`, docker links will not work.
This commit checks to see if `--net host` has been set, and if so
creates the services in `--net host` also.

Fixes #839

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>

### DIFF
--- a/shared/build/docker/container.go
+++ b/shared/build/docker/container.go
@@ -144,3 +144,11 @@ func (c *ContainerService) RunDaemonPorts(image string, ports map[Port]struct{})
 	//map[3306/tcp:{}] map[3306/tcp:[{127.0.0.1 }]]
 	return c.RunDaemon(&config, &host)
 }
+
+func (c *ContainerService) RunDaemonHost(image string) (*Run, error) {
+	// setup configuration
+	config := Config{Image: image}
+	host := HostConfig{}
+	host.NetworkMode = "host"
+	return c.RunDaemon(&config, &host)
+}
--- a/shared/build/build.go
+++ b/shared/build/build.go
@@ -198,7 +198,15 @@ func (b *Builder) setup() error {
 		log.Infof("starting service container %s", cname)
 
 		// Run the contianer
-		run, err := b.dockerClient.Containers.RunDaemonPorts(cname, img.Config.ExposedPorts)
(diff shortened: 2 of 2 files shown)


dave-tucker:
@bradrydzewski looks like CI may be having some issues....

bradrydzewski:
@dave-tucker I just pushed a new yaml that runs `apt-get update` before installing packages, which fixed the issue

shared/build/docker/container.go
```diff
@@ -144,3 +144,11 @@ func (c *ContainerService) RunDaemonPorts(image string, ports map[Port]struct{})
 	//map[3306/tcp:{}] map[3306/tcp:[{127.0.0.1 }]]
 	return c.RunDaemon(&config, &host)
 }
+
+func (c *ContainerService) RunDaemonHost(image string) (*Run, error) {
+	// setup configuration
+	config := Config{Image: image}
+	host := HostConfig{}
+	host.NetworkMode = "host"
```
bradrydzewski: I'm definitely not an expert in Docker networking, but their documentation indicates this should be used with caution. What security implications are there? Could someone issue a pull request that alters my yaml and do bad things to my test.drone.io server?

dave-tucker: When using `--net=host` all ports are made available to the host system and an application can see/use then hosts network adapters.

Therefore we only launch the services this way when:

- This is not a PR
- The user has this in their YAML:

```
docker:
  net: host
```

dave-tucker:
Closing this one due to lack of activity

Issue #819, opened 2015-01-15
Handling PR and commit webhooks for the same Commit


andy-g:
I work in a team where we all have access to the private github repo, and develop directly against it (as opposed to forking it to our personal GitHub accounts). All our changes are submitted as Pull Requests though (we push the feature branch to the shared private repo on GitHub, and then create a PR from there). 

The issue is that because the branch and the PR are part of the repository linked to drone, github will send a webhook for both the commit on the branch and the PR. This then continues for each subsequent change to that branch after a PR has been created. Drone seems to just respond to 1 and ignore the other (I'm guessing it just responds to the first one it receives 'cause sometimes it is the PR one and sometimes it's the commit one). This results in the following issues:

* If it handles the PR one and ignores the commit one, the commit message is the PR title and not the actual commit message.
* If it handles the PR one after the initial PR was created, it then results in a duplicate block for the same PR under the Pull Requests section
* If it handles the commit one, it doesn't update the PR block at the bottom. For example, if I submit a PR and the build fails, and I then push a fix which builds successfully, if drone runs the commit webhook (rather than the PR one), the PR block will still reflect a failure even though the commit from that PR has built successfully.

I realise that this may be slightly unusual workflow (especially compared to open source repositories), but I'm assuming it should still handle it. Possibly, if a PR webhook is received after the commit webhook for the same SHA, it should update the record in the commits table (updating the commit_pr column, and probably replacing the value of the commit_message column to rather be the PR title too).

bradrydzewski:
is this similar to #740 ?

andy-g:
I think it is related, yes. I think it's all related to 2 webhooks being sent from GitHub when a PR is updated (one for the branch change and the other for the PR change). The inconsistancy of which one drone picks up first and runs results in the issues I specified along with the issues in #740.

I think the isssue that @dangra raised:
> UI shows a branch and PR build but both links to PR, and none to the branch

Is as a result of  
```html
<section ng-if="layout == 'grid'" ng-repeat="group in commits | filter: { pull_request: '' } | unique: 'branch'">
```
which should actually force a strict equality comparison rather than a substring match:
```html
<section ng-if="layout == 'grid'" ng-repeat="group in commits | filter:{ pull_request: '' }:true | unique: 'branch'">
```
So PR's aren't actually excluded from the branch list (similar to what I fixed in #803). I'll submit a new PR for this issue, but we'll still need to work out how to handle the double webhook call.

x-cray:
[...]

PR #844, opened 2015-01-22
Exclude Pull Requests from the branches.

andy-g:
Currently they are included because the AngularJS "filter" filter defaults to a substring match rather than a strict equality comparison, so filtering by '' includes everything.
https://docs.angularjs.org/api/ng/filter/filter

This issue was also discussed in #819.


andy-g:
Build fail seems environmental

bradrydzewski:
yes, I re-ran and it is passing now

PR #846, opened 2015-01-23
Gitlab oauth login

Bugagazavr:
Since 7.7 version GitLab can provide ouath auth.
Now GitLab versions less than 7.7 not supported

Looks like since GitLab 8, session api can be removed.


Bugagazavr:
https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/oauth2.md

bradrydzewski:
awesome!!!

PR #847, opened 2015-01-26
Getting github orgs more than 30

udzura:
Hi, dear team

github.com (or GH Enterprise)  API have a default per_page limit to `30`. So `GetOrgs` function could not get organizations after the 31st one.

I have added the pagination loop and found this fixed.


udzura:
I have also found https://github.com/stvp/flowdock is missing... Is there any fork?

plugin/remote/github/helper.go
```diff
@@ -140,8 +140,21 @@ func GetOrgRepos(client *github.Client, org string) ([]github.Repository, error)
...
+	var allOrgs []github.Organization
+	var err error
+	page := 1
+	for {
+		options := &github.ListOptions{Page: page}
+		orgs, _, err := client.Organizations.List("", options)
+		if len(orgs) == 0 || err != nil {
+			break
+		}
+		for _, v := range orgs {
```
bradrydzewski: instead of looping, you can do something like this:

```
allorgs = append(allorgs, orgs...)
```

plugin/remote/github/helper.go
```diff
@@ -140,8 +140,21 @@ func GetOrgRepos(client *github.Client, org string) ([]github.Repository, error)
 // GetOrgs is a helper function that returns a list of
 // all orgs that a user belongs to.
 func GetOrgs(client *github.Client) ([]github.Organization, error) {
-	orgs, _, err := client.Organizations.List("", nil)
-	return orgs, err
+	var allOrgs []github.Organization
+	var err error
+	page := 1
```
bradrydzewski: There are a few [places in the code](https://github.com/udzura/drone/blob/master/plugin/remote/github/helper.go#L95) where we loop through the paginated lists. Could we slightly adjust the function so that it is consistent with the other loops? Thanks!

```
	var orgs []github.Organization
	var opts = github.ListOptions{}
	opts.PerPage = 100
	opts.Page = 1

	// loop through user org list
	for opts.Page > 0 {
		list, resp, err := client.Organizations.List("", options)
		if err != nil {
			return nil, err
		}
		orgs = append(orgs, list...)

		// increment the next page to retrieve
		opts.Page = resp.NextPage
	}
```

bradrydzewski:
> I have also found https://github.com/stvp/flowdock is missing... Is there any fork?

@udzura I have a local copy. I'll vendor the dependency and push and update shortly

plugin/remote/github/helper.go
udzura: Yes, I understand!

bradrydzewski:
ok, just merged #848 to fix the flowdock issue!

udzura:
@bradrydzewski Hi, test passed again. Thank you for review!

bradrydzewski:
cool, much thanks for the patch!

PR #856, opened 2015-01-27
Refresh only expired tokens

Bugagazavr:


shared/model/user.go
```diff
@@ -5,21 +5,22 @@ import (
...
+	Email       string `meddler:"user_email"         json:"email,omitempty"`
+	Gravatar    string `meddler:"user_gravatar"      json:"gravatar"`
+	Token       string `meddler:"user_token"         json:"-"`
+	Admin       bool   `meddler:"user_admin"         json:"admin"`
+	Active      bool   `meddler:"user_active"        json:"active"`
+	Syncing     bool   `meddler:"user_syncing"       json:"syncing"`
+	Created     int64  `meddler:"user_created"       json:"created_at"`
+	Updated     int64  `meddler:"user_updated"       json:"updated_at"`
+	Synced      int64  `meddler:"user_synced"        json:"synced_at"`
+	TokenExpiry int64  `meddler:"user_token_expires" json:"_"`
```
bradrydzewski: could we rename the database value to `user_access_expires`? The reason is that the `user_token` refers to the Drone API key (bad naming I guess). But I don't want any confusion around which value is actually expiring.

also should be a `-` instead of an `_` for the json tag :)

bradrydzewski:
looks great. 1 minor comment regarding the database column name.

Bugagazavr:
@bradrydzewski done

bradrydzewski:
thanks!

PR #857, opened 2015-01-28
Adding a list-view layout

andy-g:
I haven't yet created a user preference for this, so as to persist the setting, but if it's something you're happy to merge in I'll look into doing that as an enhancement to it.

Here's what it looks like, with an icon on the top right to toggle between layout options.

image

It appears that the less and css files are a little out of sync. Running `make lessc` before any of my changes compiled a css file quite different to the current drone.css. In light of that I compiled using less and manually applied the difference to the css file so I'm not reverting any changes that were made directly to it.


bradrydzewski:
looks good. I'm looking forward to trying this out. Perhaps some of the ideas can get included in our upcoming design.

bradrydzewski:
on minor thing to note is that `commit_id` is the table's primary key. The screenshot shows a nice sequential set of numbers (1,2,3,4) but in a real world installation with multiple repositories you will see gaps between the numbers (i.e. 1,3,8,9,14, ...)

andy-g:
Yeah, that's a good point, we have a single repo so it's incremental for us, but I don't think it's too much of a problem if you know that the build number is incremented for all repos. It's quite nice to see how many builds you CI server is pushing out :). Let me know if you have any other suggestions after you start using it, I'm happy to make some changes.

andy-g:
And is it just me, or are the less and css files out of sync?

PR #858, opened 2015-01-29
All builds now have a monotonically increasing build number (v2)

epipho:
Replacement for #737 based on all feedback.

### DIFF
--- a/server/datastore/commit.go
+++ b/server/datastore/commit.go
@@ -47,6 +47,10 @@ type Commitstore interface {
 	// KillCommits updates all pending or started commits
 	// in the datastore settings the status to killed.
 	KillCommits() error
+
+	// GetCommitBuildNumber retrieves the monotonically increaing build number
+	// from the commit's repo
+	GetBuildNumber(commit *model.Commit) (int64, error)
 }
 
 // GetCommit retrieves a commit from the
@@ -111,4 +115,10 @@ func DelCommit(c context.Context, commit *model.Commit) error {
 // in the datastore settings the status to killed.
 func KillCommits(c context.Context) error {
 	return FromContext(c).KillCommits()
+}
+
+// GetBuildNumber retrieves the monotonically increaing build number
+// from the commit's repo
+func GetBuildNumber(c context.Context, commit *model.Commit) (int64, error) {
+	return FromContext(c).GetBuildNumber(commit)
 }
--- a/server/datastore/database/commit.go
+++ b/server/datastore/database/commit.go
@@ -1,6 +1,7 @@
 package database
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/drone/drone/shared/model"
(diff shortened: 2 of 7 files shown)


bradrydzewski:
Thanks so much for this! Just a couple minor changes:

* can we add a test for the database change?
* should GetBuildNumber return the count + 1

Lastly, I would like to propose a minor change to how we inject the build number. It would be great if the build number were also injected into the yaml similar to private variables.

So in this code we could do something like this:
https://github.com/drone/drone/blob/master/server/worker/docker/docker.go#L94

```diff
	params, err := r.Repo.ParamMap()
	if err != nil {
		...
	}
+	params["DRONE_BUILD_NUMBER"], err := datastore.GetBuildNumber(c, r.Commit)
+	if err != nil {
+		log.Printf("Unable to fetch build number, Err: %s", err.Error())
+	}
	script, err := script.ParseBuild(script.Inject(r.Commit.Config, params))
```

This means we wouldn't require the changes made to `repo.go` and `build.go`

---

As a side note (in a separate PR) I may start using this method to inject all of the standard Drone environment variables (like `DRONE_COMMIT`, etc)

bradrydzewski:
note: I edited my prior comment

epipho:
Working on first unit test now. As for the 0 commits case, shouldn't this be an error as it means that no commit was added to the db?

bradrydzewski:
> shouldn't this be an error as it means that no commit was added to the db?

Yes, sorry, ignore me :)

craigtracey:
I looked at this PR again, and I don't see any issues with the implementation. Maybe this is a nit, but I still think that a monotonically incrementing $BUILD_NUM would be a highly desired feature.  Perhaps this should be named $COMMIT_NUM?

I know that "other $THING does $OTHERTHING" is not always a relevant argument, but I do think it says something that most other CI platforms implement $BUILD_NUM of some form.  It would be nice to support that case, and to avoid confusion, I think it would be ideal to use similar in syntax.

epipho:
@bradrydzewski  Cleanup completed, with db test.

@craigtracey I am not sure what you are asking for. This number is exposed as CI_BUILD_NUMBER and DRONE_BUILD_NUMBER, are you simply asking for another name with the same value?

bradrydzewski:
The `CI_*` prefixes were added for a more seamless integration with [coveralls](https://coveralls.io).

I think we can definitely add other environment variables names to promote compatibility across platforms. I would probably want to model these after Jenkins (where it makes sense):
https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-below

Let's address this as a separate pull request

craigtracey:
@epipho indeed that was what I was asking for.  I had forgotten that $CI_BUILD_NUM was already a thing. But it seems to be a misnomer. This is really the number of commits from a point in time.

Agree that it could come by way of another PR.  Was trying to avoid the "syntax has changed" dance should a true build number ever get implemented.

bradrydzewski:
@epipho  thanks so much for seeing this through!

craigtracey:
Yes, thanks @epipho!

PR #862, opened 2015-01-30
Possibility to retag docker images

ilbambino:
From docker 1.4 requires to pass a param to be able to retag images. This https://github.com/drone/drone/pull/782 doesn't really solve fully the problem.

We always push to the docker hub the latest image with the branch tag, so we can always pull the latest successful image of a given branch from the repo easily, `docker pull repo/image:master` instead of using the git hash

We have something like this in the `.drone.yml`

```
publish:
    docker:
  ...
        tags: [$(echo $DRONE_BRANCH), $(git rev-parse --short HEAD)]
        force_tags: true
```

### DIFF
--- a/plugin/publish/docker.go
+++ b/plugin/publish/docker.go
@@ -33,6 +33,7 @@ type Docker struct {
 	KeepBuild bool     `yaml:"keep_build"`
 	Tag       string   `yaml:"tag"`
 	Tags      []string `yaml:"tags"`
+	ForceTags bool     `yaml:"force_tags"`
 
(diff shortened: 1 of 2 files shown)


bradrydzewski:
thanks for this! looks like the unit test needs to be updated, and then we'll be able to merge.

ilbambino:
yeah, I just saw that, I will try to update them

davidwindell:
@bradrydzewski could you merge this please?

davidwindell:
:+1:

davidwindell:
@ilbambino Could you add this option flag to the docs? drone/docs

PR #864, opened 2015-02-01
Response as JSON instead plain text

Bugagazavr:


server/middleware/response.go
```diff
@@ -0,0 +1,16 @@
+package middleware
+
+import (
+	"net/http"
+
+	"github.com/zenazn/goji/web"
+)
+
+// Set response as JSON
+func SetJSONResponse(c *web.C, h http.Handler) http.Handler {
```
bradrydzewski: for consistency, perhaps we could put it in `SetHeaders` with all the other headers?
https://github.com/drone/drone/blob/master/server/middleware/header.go

bradrydzewski:
1 minor comment above but otherwise looks good to merge! thanks!

Issue #776, opened 2014-12-30
RFC - Repository Syncing?


bradrydzewski:
I'd like to have a dialog about syncing repositories in Drone. I'd like to understand if it is the right approach going forward, or if it is something we should alter prior to moving from alpha to beta status. The purpose of this thread is to discuss the existing approach, the benefits and pitfalls, and possible alternatives. If we decide to stick with the current approach we can figure out how to fix some of the existing issues.

The more input the better, so please feel free to add your 2 cents.

Here is how we sync repositories today upon login:

* Get all my repositories from remote system (GitHub, GitLab, etc)
* Insert into the database if not exists
* Insert my role into the database if not exists
* If a repository or role already exists, it is skipped

This leads to a few issues:

* If my access is revoked from a repository it is not currently reflected in Drone
* If my access is modified (changed from write -> read) it is not updated in Drone
* Bitbucket does not provide permission data when retrieving the list of repositories, so we grant admin access by default (which really bothers me)
* I may see repositories on my Dashboard that aren't relevant to my daily activities, just because I have access to the repository
* It doesn't allow for role customization. For example, perhaps I should have write access to the repo in GitHub but read access in Drone

We could definitely improve the sync process and fix the issues mentioned above. That being said, syncing is kind of a pain and complicates the underlying codebase. This approach, however, does have two major benefits worth mentioning:

* I only need to manage permissions in one place (ie GitHub)
* As a new user, I immediately see any projects I have access to on my dashboard

---

#### Repository Setup:

1. User authenticates (no syncing)
2. User clicks button to add repository
3. User types repository name (there is no list here, since we didn't sync)
4. Query remote system (ie GitHub) to see if user has access to repository
  1. If Yes, continue
  2. If No, show Not Found error message
5. Does repository already exist in Drone database (ie already setup by another user)?
  1. If Yes, grant user Read / Write access based on permissions in remote system
  2. If No, but user does not have Admin rights to the repository, show Error message
  2. If No, but user has Admin rights, add repository to Drone and grant user Admin access in Drone

#### Access Management

This is kind of a hybrid approach to access management. The user can automatically add themselves to a project without having to involve the repository admin in Drone, using native GitHub permissions. The key difference is that a user is granted access to a repository in Drone on-demand, as opposed to on-sync when they login.

[...]

PR #866, opened 2015-02-04
Delete Repository in Drone DB 

ulrichSchreiner:
Implement the deletion of a Repository in Drone. Also remove the Webhook and the deploy key from the remote vcs. Gitlab is implemented/tested, github untested. Bitbucket and gogs are empty.

### DIFF
--- a/client/repos.go
+++ b/client/repos.go
@@ -35,12 +35,12 @@ func (s *RepoService) Enable(host, owner, name string) error {
 // DELETE /api/repos/{host}/{owner}/{name}
 func (s *RepoService) Disable(host, owner, name string) error {
 	var path = fmt.Sprintf("/api/repos/%s/%s/%s", host, owner, name)
-	return s.run("DELETE", path, nil, nil)
+	return s.run("PATCH", path, nil, nil)
 }
 
 // DELETE /api/repos/{host}/{owner}/{name}?remove=true
 func (s *RepoService) Delete(host, owner, name string) error {
-	var path = fmt.Sprintf("/api/repos/%s/%s/%s?remove=true", host, owner, name)
+	var path = fmt.Sprintf("/api/repos/%s/%s/%s", host, owner, name)
 	return s.run("DELETE", path, nil, nil)
 }
 
--- a/client/repos.go
+++ b/client/repos.go
@@ -32,10 +32,10 @@ func (s *RepoService) Enable(host, owner, name string) error {
 	return s.run("POST", path, nil, nil)
 }
 
-// DELETE /api/repos/{host}/{owner}/{name}
+// POST /api/repos/{host}/{owner}/{name}/deactivate
 func (s *RepoService) Disable(host, owner, name string) error {
-	var path = fmt.Sprintf("/api/repos/%s/%s/%s", host, owner, name)
-	return s.run("PATCH", path, nil, nil)
+	var path = fmt.Sprintf("/api/repos/%s/%s/%s/deactivate", host, owner, name)
+	return s.run("POST", path, nil, nil)
 }
 
 // DELETE /api/repos/{host}/{owner}/{name}?remove=true
--- a/client/repos.go
+++ b/client/repos.go
@@ -38,6 +38,12 @@ func (s *RepoService) Disable(host, owner, name string) error {
 	return s.run("DELETE", path, nil, nil)
 }
 
+// DELETE /api/repos/{host}/{owner}/{name}?remove=true
+func (s *RepoService) Delete(host, owner, name string) error {
+	var path = fmt.Sprintf("/api/repos/%s/%s/%s?remove=true", host, owner, name)
+	return s.run("DELETE", path, nil, nil)
+}
+
 // PUT /api/repos/{host}/{owner}/{name}
 func (s *RepoService) SetKey(host, owner, name, pub, priv string) error {
 	var path = fmt.Sprintf("/api/repos/%s/%s/%s", host, owner, name)
--- a/plugin/remote/gogs/gogs.go
+++ b/plugin/remote/gogs/gogs.go
@@ -153,6 +153,12 @@ func (r *Gogs) Activate(user *model.User, repo *model.Repo, link string) error {
 	return err
 }
 
+// Deactivate removes a repository by removing all the post-commit hooks
+// which are equal to link and removing the SSH deploy key.
+func (r *Gogs) Deactivate(user *model.User, repo *model.Repo, link string) error {
+	return fmt.Errorf("Remove %#v in gots not implemented", *repo)
+}
+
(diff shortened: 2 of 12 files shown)


plugin/remote/gogs/gogs.go
```diff
@@ -153,6 +153,12 @@ func (r *Gogs) Activate(user *model.User, repo *model.Repo, link string) error {
 	return err
 }
 
+// Deactivate removes a repository by removing all the post-commit hooks
+// which are equal to link and removing the SSH deploy key.
+func (r *Gogs) Deactivate(user *model.User, repo *model.Repo, link string) error {
+	return fmt.Errorf("Remove %#v in gots not implemented", *repo)
```
bradrydzewski: minor spelling. Gogs

server/handler/repo.go
```diff
@@ -52,6 +52,42 @@ func DelRepo(c web.C, w http.ResponseWriter, r *http.Request) {
...
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		// setup the post-commit hook with the remote system and
+		// if necessary, register the public key
+		var hook = fmt.Sprintf("%s/api/hook/%s/%s", httputil.GetURL(r), repo.Remote, repo.Token)
+		if err := remote.Deactivate(user, repo, hook); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
```
bradrydzewski: I think we should proceed with deleting the repository from Drone even if it fails to delete the key or hook. We could log the error for debugging purposes, but it would proceed to the delete step.

The reason is that we cannot always control what is happening in GitHub, etc. There are edge cases I can think of where I want to delete the repository from Drone (perhaps as the Drone admin) but I don't have access to the repository in GitHub to remove the hooks.

bradrydzewski:
Added my notes. @Bugagazavr would you be able to share your thoughts on improving this REST endpoint to better conform to principles?

Bugagazavr:
i think wee need to separate deactivation ( `PATCH` )  and removing method ( `DELETE` )

server/handler/repo.go
Bugagazavr: Yep need to compare errors, and return multiple errors http://godoc.org/github.com/augustoroman/multierror

ulrichSchreiner: as far as i understand brad's note it would be better to log the error's from the remote (github, gitlab, ...), but not return the error to the caller. so i don't think his suggestion is to accumulate them but to log some of them and don't return with an error.

ulrichSchreiner:
Please note that the semantic of the REST-API now changed. In the current master the `DELETE /api/repos/:host/:owner/:name` only deactivated it, now it really deletes the repository. I don't know if any systems are currently calling this endpoint.

bradrydzewski:
@Bugagazavr I think we need to be a bit more explicit with the API endpoint. The PATCH endpoint, doesn't really convey the purpose of the endpoint. It is not just updating fields, it is suspending all further builds for your repository. This feels like it should be a bit more explicit.

I researched this a bit and found many REST APIs (Google, Digital Ocean, Stripe, etc) put actions in the URL. I wasn't really a fan of this approach, but after further consideration, I think the fact that it is self-documenting is pretty important. For example:

```
POST https://api.stripe.com/v1/charges/{CHARGE_ID}/dispute/close
```

So perhaps we do this instead:

```
POST /api/repos/{host}/{owner}/{name}/deactivate
```

ulrichSchreiner:
I would vote for @bradrydzewski proposal with the `POST ..../deactivate` call. It is more intuitiv than a `PATCH`.

Bugagazavr:
maybe?

`POST /api/repos/{host}/{owner}/{name}/activate`
`DELETE /api/repos/{host}/{owner}/{name}/deactivate`

ulrichSchreiner:
hm,

in this case we would have 
```
POST /api/repos/{host}/{owner}/{name}/activate
DELETE /api/repos/{host}/{owner}/{name}/deactivate
DELETE /api/repos/{host}/{owner}/{name}
```
but the
```
DELETE /api/repos/{host}/{owner}/{name}/deactivate
```
does not deletey anything, it only deactivates. so the `DELETE` is a little misleading.

daMupfel:
and GET /api/repos/{host}/{owner}/{name} is getting new field to show if activated or not? Or how do you want to query the active status?
I would like to have a common verb, not to. For example enabled and denn POST and DELETE or something like that.

ulrichSchreiner:
I think the current `GET` returns a JSON struct with the `Active` field.

bradrydzewski:
Iet's modify this PR to the following:

```
POST /api/repos/{host}/{owner}/{name}/deactivate
DELETE /api/repos/{host}/{owner}/{name}
```

we could also modify the activation endpoint (below) however, depending on #776 it may not be necessary.

```
POST /api/repos/{host}/{owner}/{name}/activate
```

ulrichSchreiner:
ok, i changed the `deactivate` endpoint to `POST` and i changed it to return the repo-data as json, because it is some sort of "update" so i think it would be great to return the new data to the client. the activation endpoint is unchanged.

i will do a PR on the docs-repo to reflect this changes.

server/handler/repo.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/drone/drone/plugin/remote"
 	"github.com/drone/drone/server/datastore"
+	"github.com/drone/drone/shared/build/log"
```
bradrydzewski: I realize this is confusing (sorry about that) but the build log only exists for the CLI and is going to be deprecated for something smarter. Let's use the standard `log` package instead

client/repos.go
```diff
@@ -35,6 +35,12 @@ func (s *RepoService) Enable(host, owner, name string) error {
 // DELETE /api/repos/{host}/{owner}/{name}
 func (s *RepoService) Disable(host, owner, name string) error {
 	var path = fmt.Sprintf("/api/repos/%s/%s/%s", host, owner, name)
```
bradrydzewski: update method and path?

ulrichSchreiner:
ok, no using standard log package. there was an error too when formatting the error message. i use the prfix `[ERROR]` so if you like you can use [hashicorp logutils](https://github.com/hashicorp/logutils) which is a simple but great tool around the default log package.

bradrydzewski:
Thanks! I definitely like the idea of outputting log levels and using a more sophisticated logging package, and will play around with the hashicorp implementation. I've also been looking at [logrus](github.com/Sirupsen/logrus) due to its support for custom log formatters.

Note: one minor change required to `client/repos.go` to change from `PATCH` to `POST` and update path to `/deactive` as well :)

bradrydzewski:
merged. thanks for your patience! :)

PR #871, opened 2015-02-08
Move build commands to remote plugin

cedk:
As discussed some weeks ago, it will be good to have the remotes generate the build commands.


cedk:
@bradrydzewski Is it possible to have some feedback on this change?
Thanks.

bradrydzewski:
@cedk sorry for the delay. The change looks good but we are also experimenting with a radical new approach to plugins. Every build step will be delegated to a Docker-based plugin, and will be composed (like a mixin) with the running build container.

We have an example git clone plugin here:
https://github.com/drone-plugins/drone-git

We can probably take the code you've written and port directly to the new plugin model, so let's keep this open for now. Once I have a working version of Drone using the new plugin model I'll let you know and we can figure out how to port.

cedk:
@bradrydzewski So this mean, I will have to create a drone-hg plugin. Using specific docker image per kind of repository sounds pretty good to reduce the size of the images.

bradrydzewski:
now that 0.4 is merged to master, the path forward is to create a drone-hg plugin

PR #875, opened 2015-02-12
Implementing angular-moment to automatically update dates / durations

andy-g:
Allowing durations and dates to auto update over time, similar to how GitHub and twitter handle time periods. It still uses Momemt.js, but implements it as a directive and allows the time periods to be updated automatically.


bradrydzewski:
very cool. thanks!

PR #876, opened 2015-02-13
Change repo to be private when GHE is private mode.

masarakki:
When the case of sending pull-request to repository on private-mode Github:enterprise,
build of `origin` fails because it can't checkount repository.  
(build of `forked` repository will success)

It is because of [builder writes private-key when repository is private](https://github.com/drone/drone/blob/master/server/worker/docker/docker.go#L155),
however, private-mode ghe's repository is not flagged as private,
so builder doesn't write private-key.


bradrydzewski:
yes, great catch. thanks!

PR #880, opened 2015-02-18
Remove LANG and LANGUAGE from environment

floatdrop:
LANG and LANGUAGE should be set by container image, not by drone.

### DIFF
--- a/shared/build/build.go
+++ b/shared/build/build.go
@@ -448,6 +448,8 @@ func (b *Builder) writeDockerfile(dir string) error {
 		// is the "ubuntu" user, since all build images
 		// inherit from the ubuntu cloud ISO
 		dockerfile.WriteUser("ubuntu")
+		dockerfile.WriteEnv("LANG", "en_US.UTF-8")
+		dockerfile.WriteEnv("LANGUAGE", "en_US:en")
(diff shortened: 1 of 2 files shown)


bradrydzewski:
Thanks for this. One minor request. Can we keep the `lang` and `language` for the default `drone/images`? https://github.com/floatdrop/drone/blob/master/shared/build/build.go#L450

```
	switch {
	case strings.HasPrefix(b.Build.Image, "bradrydzewski/"),
		strings.HasPrefix(b.Build.Image, "drone/"):
+		f.WriteEnv("LANG", "en_US.UTF-8")		
+		f.WriteEnv("LANGUAGE", "en_US:en")
		dockerfile.WriteUser("ubuntu")
		dockerfile.WriteEnv("LOGNAME", "ubuntu")
		dockerfile.WriteEnv("HOME", "/home/ubuntu")
		dockerfile.WriteRun("sudo chown -R ubuntu:ubuntu /var/cache/drone")
		dockerfile.WriteRun("sudo chown -R ubuntu:ubuntu /usr/local/bin/drone")
	default:
```

This will keep things backward compatible for now when using the official Drone images, but will prevent them from being set in custom images.

Issue #882, opened 2015-02-19
WS: Error getting permissions for repository


floatdrop:
In some repositories I get following error in logs:

```WS: Error getting permissions for repository grid-client-js. Error: sql: no rows in result set```

And no output in UI is printed. However after build is finished, all output is there. Where should I start digging?

**Comment by floatdrop** [comment]
May be this is related: I have `selenium-binaries` in node dependencies. When it try's to build itself it gives tons of output: 

```
[selenium-binaries] Downloading selenium
[selenium-binaries] http://selenium-release.storage.googleapis.com/2.44/selenium-server-standalone-2.44.0.jar
[selenium-binaries]  -> /home/ubuntu/.selenium-binaries/selenium/2.44.0
[...]

PR #884, opened 2015-02-20
Allow to read logs for public repo in ws

floatdrop:
Fix for #882. If repository is public, but no rows in `perms` table is met - logs should still be accessable.

### DIFF
--- a/server/handler/ws.go
+++ b/server/handler/ws.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"database/sql"
 	"log"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


bradrydzewski:
We should only log when `err != nil && err != sql.ErrNotFound` otherwise it will spam the logs with false positive errors

floatdrop:
Done.

server/handler/ws.go
```diff
@@ -65,14 +66,16 @@ func WsUser(c web.C, w http.ResponseWriter, r *http.Request) {
...
+				// user must have read access to private the repository
 				// in order to pass this message along
-				if role, err := datastore.GetPerm(ctx, user, work.Repo); err != nil || role.Read == false {
-					if err != nil {
-						log.Printf("WS: Error getting permissions for repository %s. Error: %s\n", work.Repo.Name, err)
-					} else {
-						log.Printf("WS: No read access for repository %s\n", work.Repo.Name)
-					}
+				if work.Repo.Private == true && role.Read == false {
+					log.Printf("WS: No read access for repository %s\n", work.Repo.Name)
```
bradrydzewski: actually this loggin should be removed also, since it is expected behavior. This is a global event feed. This means if 1000 users are logged-in to Drone, 1000 users will get this message. If only 1 user has access, we would see 999 log entries.

floatdrop: @bradrydzewski make sens, removed logging, but build is failing on fetching deps.

PR #897, opened 2015-02-24
quick patch that lets someone with write access restart a build

bradrydzewski:


andy-g:
Thanks for this @bradrydzewski, this has been niggling me for some time.

It seems like I'm still seeing an issue where the rebuild button is not visible for users that have write access to the repo but are not repo admins. Looking at the angular conditions, it seems like the button needs `repo.role` to be populated because it checks `repo.role.admin` and `repo.role.write`. Looking at the json response from `handler.GetRepo` ("/api/repos/:host/:owner/:name") for one of the users that isn't seeing the button, it turns out there is no `repo.role` object. Seem like it just totally skips it at the `if role.Admin == false {` line in handler/repo.go (around line 32).

Is the solution to simply include the role object, but exclude the public_key & params? Something like:

```GO
if role.Admin == false {
	json.NewEncoder(w).Encode(struct {
		*model.Repo
		Perm      *model.Perm `json:"role"`
	}{repo, role})
	return
}
```

I'm happy to make the change and submit a PR but just wanted to make sure this this would work and wasn't going to cause other issues.

PR #899, opened 2015-02-25
Displaying rebuild button for users with write access to the repo (not necessarily admins)

andy-g:
Including repo.role for non admins to allow rebuild button to display when the user has write access to the repository.

Initial conversation started in #897

### DIFF
--- a/server/handler/repo.go
+++ b/server/handler/repo.go
@@ -30,7 +30,10 @@ func GetRepo(c web.C, w http.ResponseWriter, r *http.Request) {
 	// if the user is not requesting (or cannot access)
 	// admin data then we just return the repo as-is
 	if role.Admin == false {
-		json.NewEncoder(w).Encode(repo)
+		json.NewEncoder(w).Encode(struct {
+			*model.Repo
+			Perm      *model.Perm `json:"role"`
+		}{repo, role})
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


andy-g:
Couldn't glean too much info from the CI failure, not sure if it's environmental or not?

bradrydzewski:
Thanks for the PR.  Looks like the code change needs to be run through `gofmt` which will adjust the alignment on the below block:

```
		json.NewEncoder(w).Encode(struct {
			*model.Repo
-			Perm      *model.Perm `json:"role"`
+			Perm *model.Perm `json:"role"`
		}{repo, role})
		return
```

If you use Sublime there is an awesome plugin for Go that will automatically format Go files on save. Very convenient and hightly recommended.

daMupfel:
You might also want to check out https://github.com/bradfitz/goimports from one of the maintainers of golang. Its basicly go fmt with autodetection of missing imports (it also removes imports). You can use te goimports command instead of go fmt.
Not part of this discussion. But interesting :)

andy-g:
Thanks @bradrydzewski  and @daMupfel for the tips, I'm really new to Go, so I appreciate the suggestions. Build seems happier now.

bradrydzewski:
thanks!

Issue #363, opened 2014-06-26
Run Docker with Custom Options


907th:
I broke my brain trying to understand why my custom image is running under `root` instead of `ubuntu` user. [The issue](https://github.com/drone/images/issues/32). Finally I found that Drone uses `root` [for all images](https://github.com/drone/drone/blob/678e4ddea76a47004cc30139f3c2467c049198b3/pkg/build/build.go#L397) except for`bradrydzewski`'s ones. :astonished:

I think Drone should allow to specify a custom user for build in `.drone.yml`

bradrydzewski:
we've added the `docker` section to the yaml and many customization options. We are altering the Drone build process to use the default `USER` assigned to the image. We are tracking this change in #749 so I'm going to close this issue.

907th:
@bradrydzewski Thanks! Is new `docker` section documented anywhere? Can not find docs for it :(

nardeas:
Still cannot find this anywhere in documentation. Any help here?

PR #900, opened 2015-02-25
enable to set user name who builds in .drone.yml

masarakki:
I want to use my own docker image but run builds by non-root user.

### DIFF
--- a/shared/build/build.go
+++ b/shared/build/build.go
@@ -440,19 +440,22 @@ func (b *Builder) writeDockerfile(dir string) error {
 		dockerfile.WriteAdd("src", filepath.Join(b.Repo.Dir))
 	}
 
+	// the default user for all official Drone image
+	// is the "ubuntu" user, since all build images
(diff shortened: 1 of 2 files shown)


masarakki:
related: https://github.com/drone/images/issues/32

tsuchikazu:
:+1:

purehyperbole:
+1

gutenye:
:+1: I really need this to run my own archlinux image to publish AUR package. And `$ makepkg from ArchLinux` must run as non-root. When will this be merged?

shawnzhu:
see [issue #363(comment)](a quick workaround could be adduser and then su in your build script.)
> a quick workaround could be `adduser` and then `su` in your build script.

bradrydzewski:
merged 0.4 into master. Because 0.4 plugins rely on a shared `/drone` volume across all containers, and the plugins (including git plugin) run as root we need the build container to run as root as well. This is a technical limitation of the 0.4 design.

PR #901, opened 2015-02-26
Make bundled assets overridable

MerlinDMC:
This adds a `server-assets-folder` configuration option which allows overriding the assets bundled to the binary.

This makes iterations on the UI easier and files can be updated without rebuilding and restarting droned.

### DIFF
--- a/packaging/root/etc/drone/drone.toml
+++ b/packaging/root/etc/drone/drone.toml
@@ -9,6 +9,12 @@ port=":80"
 # key=""
 # cert=""
 
+#####################################################################
+# Assets configuration
+#
+# [server.assets]
+# folder=""
+
 # [session]
 # secret=""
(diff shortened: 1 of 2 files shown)


bradrydzewski:
So as long as you don't embed the files in the binary (ie don't run the `make embed` command) the static assets are served directly from disk. It is actually one of the selling points of the library we are using:

> During development go.rice will load required files directly from disk. Upon deployment it is easy to add all resource files to a executable using the rice tool, without changing the source code for your package. go.rice provides several methods to add resources to a binary.

That being said, this change is important because I want to support themes. People should be able to completely swap out the user interface with their own custom theme.

So I'll definitely merge this change. I don't think it is necessary for local development, but it does allow us to support themes which is really awesome. So thanks!

Issue #902, opened 2015-02-28
More flexible Yaml with Matrix


bradrydzewski:
I would like to solve issues #865 #874 and #696 by proposing a more flexible yaml format. The initial yaml was inspired by Travis with some minor modifications to better align with Docker.

Below is my proposal. It is an evolution of the existing yaml with inspiration from `docker compose`. It also takes the new (proposed) plugin system into account.

Example:

```yaml
build:
  image: python
  envs:
    - foo=bar
    - baz=boo
  cmds:
    - go build
    - go test

compose:
  redis:
    image: redis

  postgres:
    image: postgres

publish:
  azure:
    image: azure
    args:
      token: 123123121
      region: east-1
```

This should work well for #696 #874 since it would allow multiple named notifications or deployments for the same plugin:

```yaml
notify:
  proj_managers:
    image: slack
    when:
      branch: prod
    args:
      channel: project_managers
      username: drone

  devs:
    image: slack
    args:
      channel: devs
      username: drone
```

The above example may be contrived, however, it would allow us to notify two different groups, to two different slack channels, for two different events. Notice the `slack` is the Docker image (plugin) and `devs` and `proj_managers` are just labels that we use to differentiate / label the notification.

I believe this could also satisfy #865 since service declarations could be more verbose, similar to `docker compose`:

```yaml
compose:
  redis:
    image: redis
    volume: {{awesome_volume}}:/var/lib/pgsql/data
```

The downside here is that things are a bit more verbose than before. So I wanted to get everyone's thoughts and see if it makes sense to sacrifice brevity in favor of flexibility.

cc @donny-dont / @dave-tucker / @dvigueras / @vaijab / @epipho / @gdborton

vaijab:
Looks much more powerful and flexible. However I see one potential issue - yaml hashes are not sorted, which could cause undesired behaviour.

bradrydzewski:
maybe we could add an optional "weight" or "order" attribute and sort the values prior to execution?

epipho:
I like the order attribute. Long term we could allow sections with identical order values to execute concurrently.

gdborton:
I think this is great, the added verbosity shouldn't be an issue as long as it is well documented.

dave-tucker:
:+1: 
Providing we have access to specify the parameters to be used by "docker run" for both services and the main test container, I think this will meet the use case described in #865.

[...]

Issue #697, opened 2014-11-12
Support for Docker --pull Flag


vaijab:
Currently, at least I couldn't find in the code how to tell drone to pull base build image before running `script` commands.

We use our own build images and once we update them, drone has no way of knowing if we use the same build image tag, like `image: foo/bar:latest`.

This should default to `false`, but there should be an option in `.drone.yml` to change that.

```yaml
image: foo/bar:latest
image_pull: true
script:
  - exit 0
```

bradrydzewski:
sounds like the `docker --pull` flag. We could do something like this in the yaml:

```
image: go
docker:
  pull: true
```

vaijab:
I can look into implementing that, I guess, but cannot promise any ETA.

vaijab:
Closing as #909 solves this problem.

PR #909, opened 2015-03-05
always update docker image

masarakki:


bradrydzewski:
What about making this behavior a bit more explicit. For example, if I specify `:latest` in the image name it will always check the registry for an updated version? Instead of checking every single time?

```diff
-	if _, err := b.dockerClient.Images.Inspect(b.Build.Image); err == docker.ErrNotFound {
+	if _, err := b.dockerClient.Images.Inspect(b.Build.Image); err == docker.ErrNotFound || strings.HasSuffix(b.Build.Image, ":latest") {
		// download the image if it doesn't exist
		if err := b.dockerClient.Images.Pull(b.Build.Image); err != nil {
			return fmt.Errorf("Error: Unable to pull image %s. %s", b.Build.Image, err)
		}
	}
```

jvliwanag:
Hm, or might it be better as an explicit option in the yaml? And by default, this is `false`.

Setting this to be true when the tag is set as `latest` is odd.

bradrydzewski:
@jvliwanag `latest` is the official Docker tag for the latest revision:
https://github.com/docker/docker/blob/8f3ecf8779fd450d63049f53f65899d41cf759d6/graph/tags.go#L21

I think it is pretty appropriate

jvliwanag:
Yes, but in the same manner as `docker run myimage:latest` doesn't pull an image every time I run it, I'm thinking nor should drone do the same per build.

I can imagine pulling the image of a certain tag to be useful in some cases. Tags which are not necessarily `latest`. (e.g. if i have a `dev` tag). So having it as an explicit yaml option would be my vote, vs being implicitly defined by using the `:latest` tag. Moreover, it's confusing that using the bare image name without any tags specified would pull the `latest` on first use but not pull it every build.

bradrydzewski:
The services section of the yaml file is a string array of image names. Using `latest` would allow us to pull individual service images without having to change the yaml file. But maybe service images aren't as big of a deal.

Either way I vote for the simplest solution, whatever that may be, since it will get deprecated in a few weeks when #902 is introduced.

masarakki:
I think in drone, we use tags like `go:1.2` with meanings that 'this image has go version 1.2',
so I thinks `go:1.2` is NOT immutable.   
for example, current `ruby:2.0.0` has `ruby-2.0.0-p195`, and upgrade it to `ruby-2.0.0-p598`,
I expected `ruby:2.0.0` will be updated.
Is it wrong way of thinking?

bradrydzewski:
The official language stacks are immutable. They share the same underlying base image to save space, but do not build on eachother. Check out the official [go 1.3](https://github.com/docker-library/golang/blob/1a422afd7db928a821e97906ed27ed606e2f072a/1.3/Dockerfile) and [go 1.4](https://github.com/docker-library/golang/blob/1a422afd7db928a821e97906ed27ed606e2f072a/1.4/Dockerfile) dockerfiles for example.

We have already solved this problem in the upcoming 0.4 release as part of the yaml file re-design, and in response to #697. Anything we do here will be a temporary fix until 0.4 is available.

masarakki:
change to update only when `:latest` tag.

bradrydzewski:
thanks for being open to the temp fix. The permanent fix will master in a couple weeks :)

Issue #869, opened 2015-02-07
Remove Socat for local access to service containers


bradrydzewski:
Ideally we could have a fallback when `socat` is not installed. My idea is to inject a perl script in the container when `socat` isn't available that is capable of proxying connections. Why choose Perl? It is the only scripting language included in the default Ubuntu and Debian containers.

I found the an proxy Perl script here:
https://github.com/pkrumins/perl-tcp-proxy/blob/master/tcp-proxy.pl

I adjusted the below example to work with Redis:

```perl
#!/usr/bin/perl
use warnings;
use strict;

use IO::Socket;
use IO::Select;

my $ioset = IO::Select->new;
my %socket_map;

my $debug = 1;

sub new_conn {
    my ($host, $port) = @_;
    return IO::Socket::INET->new(
        PeerAddr => $host,
        PeerPort => $port
    ) || die "Unable to connect to $host:$port: $!";
}

sub new_server {
    my ($host, $port) = @_;
    my $server = IO::Socket::INET->new(
        LocalAddr => $host,
        LocalPort => $port,
        ReuseAddr => 1,
        Listen    => 100
    ) || die "Unable to listen on $host:$port: $!";
}

sub new_connection {
    my $server = shift;
    my $client = $server->accept;

    my $remote = new_conn($ENV{"REDIS_PORT_6379_TCP_ADDR"}, $ENV{REDIS_PORT_6379_TCP_PORT});
    $ioset->add($client);
    $ioset->add($remote);

    $socket_map{$client} = $remote;
    $socket_map{$remote} = $client;
}

sub close_connection {
    my $client = shift;
    my $remote = $socket_map{$client};
    
    $ioset->remove($client);
    $ioset->remove($remote);

    delete $socket_map{$client};
    delete $socket_map{$remote};

    $client->close;
    $remote->close;
}

my $server = new_server('0.0.0.0', $ENV{REDIS_PORT_6379_TCP_PORT});
$ioset->add($server);

while (1) {
    for my $socket ($ioset->can_read) {
        if ($socket == $server) {
            new_connection($server);
        }
        else {
            next unless exists $socket_map{$socket};
            my $remote = $socket_map{$socket};
            my $buffer;
            my $read = $socket->sysread($buffer, 4096);
            if ($read) {
                $remote->syswrite($buffer);
            }
            else {
                close_connection($socket);
            }
        }
    }
}
```

So how would this work? We would embed the perl script in our `build.sh` file and write that file from inside the container. It could look something like this:

[...]

PR #911, opened 2015-03-11
Use 'which' (if present) to check socat anywhere in path #869.

jvortmann:
When using a base image that is not the provided ones (some custom container), the socat utility
could have been installed on a diferent place than /usr/bin. For instance, yum on CentOS installs
socat on /usr/local/bin.

This commit just rely ou 'which' (when present) to check if socat exists anywhere on the path.
Also, if socat is installed anywhere by /usr/bin before this, the access to services failed
silently.

### DIFF
--- a/shared/build/proxy/proxy.go
+++ b/shared/build/proxy/proxy.go
@@ -17,7 +17,7 @@ set +e
 // this command string will check if the socat utility
 // exists, and if it does, will proxy connections to
 // the external IP address.
-const command = "[ -x /usr/bin/socat ] && socat TCP-LISTEN:%s,fork TCP:%s:%s &\n"
+const command = "[ -x /usr/bin/which ] && which socat >/dev/null && socat TCP-LISTEN:%s,fork TCP:%s:%s &\n"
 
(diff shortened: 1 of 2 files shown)


JeanMertz:
While in favour of this change, I'd suggest changing `which` to something like `command -v`.

See: http://stackoverflow.com/a/677212/747032

jvortmann:
@JeanMertz makes sense. Already push the changes. Using command is even better because you don't even need to check for which presence, as it is builtin.

jvortmann:
@JeanMertz I've opened a new pull-request, as this one fails the tests because there is more than 1 commit. Check out #932.

bradrydzewski:
> I've opened a new pull-request, as this one fails the tests because there is more than 1 commit. Check out #932.

fyi, has nothing to do with > 1 commit. There is an issue with one of our postgres unit test that occasionally fails due to incorrect ordering. I re-ran the build and it passed

jvortmann:
@bradrydzewski thanks for the heads up :+1: The useful commit is in the new PR anyway :)

Issue #147, opened 2014-03-02
Implement proper Caching


skorfmann:
I was playing around with Drone.io over the weekend and I'm really impressed. 

However, there is one big issue for our Rails project: Bundling all the Gems (~ 250 Gems, where about 10 of these are git checkouts) takes about 10 minutes for each build, as I've found no way to provide a cacheable directory to Bundler.

I've seen the issues #43 and #143, but as far as I understood the solution proposed in #43, the cache would only be invalidated when the actual setup commands have changed. In my case, it would need to re-run the commands when the content of the project's ```Gemfile.lock``` has changed.

Furthermore, it would be neat to be able to share a cache directory between different projects. In our current Jenkins setup, we're sharing a global Bundler directory, which speeds up new project builds enormously. 

Here is an excerpt of our build file:

```bash
BUNDLE_PATH="$JENKINS_HOME/shared/bundle"
bundle install --path="$BUNDLE_PATH"
```

kylewelsby:
I would really like for my build to run faster, all the `npm install` and `bundle install` slows the builds down considerably. 
Could drone have a local HTTP cache/mirror or something for regular requests to known services.

glaszig:
i am doing a `ls -la` before `bundle install` in the `script` phase and the cache directory is _always_ empty.

nathwill:
it looks like the cache is currently branch-specific, which makes it awkward for feature-branch based PRs, as the cache gets duped for every feature branch, but doesn't have the advantage of faster builds

bradrydzewski:
Per branch caching is removed per #912 (thanks @nathwill)

Note that #902 will expose much more of the underlying Docker implementation and will allow mounting volumes from the yaml file. So #902 will end up replace the `cache` section with a `volumes` section.

We're also working on modularity and plugins. The `git clone` functionality is being moved to a plugin:
https://github.com/drone-plugins/drone-git

This will give us much more flexibility and should allow us to perform a `git checkout` instead of a `git clone` if a `.git` directory already exists. This would allow us to start caching the repository root. Stay tuned.

davidak:
consider this:

you have cached installed dependencies and run a build again. since the last build the dependencies have changed.

so we have to run the commands from setup section again to update the dependencies. that would be still faster than installing all again.

glaszig:
@nathwill #912 does not seem to change anything for me. i'm a little confused now -- is this still a work in progress or should the simple case of sharing one folder (for rubygems) basically work?

[...]

PR #912, opened 2015-03-11
use same cached volumes between branches

nathwill:
my first stab at messing with golang ever, so merge with caution, but I *think* this addresses https://github.com/drone/drone/issues/147#issuecomment-78348874


bradrydzewski:
Yes, in retrospect per-branch caching wasn't a great idea. Thanks for the fix.

PR #916, opened 2015-03-15
Add new Python images

cedk:


meshy:
This, please!

donny-dont:
@meshy the drone specific images are deprecated. Instead you can use the language stacks from docker. There's a Python one at https://hub.docker.com/_/python/ that you can use. These will end up being more up to date than anything drone provides.

meshy:
Thanks for the quick reply @donny-dont.

I can't find any mention of this deprecation in the docs -- am I looking in the wrong place?

I'm just trying to get this set up for the first time, and finding it very confusing, sorry.

I've followed this setup:
```yml
image: python:3.4
script:
  - psql -c 'create database caas;' -U postgres -h localhost
  - pip install -r requirements.txt
  - make test
services:
  - postgres:9.4
```

But I get

```
/usr/local/bin/drone: line 81: psql: command not found
```
I tried installing `postgresql-client` and I get:
```
psql: could not connect to server: Connection refused
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
```

I tried  `socat`, but that has no effect.

Am I missing something? The drone images seem like a better idea than these other ones to me. `:/`

meshy:
I've also tried with the `library/` images for python and postgres.

donny-dont:
You should ask on the gitter channel. Its the support forum for Drone.

meshy:
Thanks

bradrydzewski:
as @donny-dont mentioned in 0.4 the `github.com/drone/images` are deprecated and we fully support any image in the index, including `library` images without the suffix specified.

maintaining the internal map of images in `shared/build/images.go` is no longer required.

cedk:
But 0.4 is not yet released :-(

PR #926, opened 2015-03-18
Provide a warning in the CLI when server scheme isn't supplied on cli

BlakeMesdag:


daMupfel:
You should create a pull request here: https://github.com/drone/drone-cli

the cli was split out of the main repo. The cli package should probably removed from drone/drone since it may be confusing for people.

BlakeMesdag:
Ah, all the imports still point to this repo. I'll move this over

PR #932, opened 2015-03-22
Use 'comand -v' to check presence of socat anywhere in path #869.

jvortmann:
When using a base image that is not the provided ones (some custom container), the socat utility could have been installed on a diferent place than /usr/bin. For instance, yum on CentOS installs socat on /usr/local/bin.

This commit just rely ou 'command -v' to check if socat exists anywhere on the path.
Also, if socat is installed anywhere by /usr/bin before this, the access to services failed
silently.


bradrydzewski:
thanks!

PR #935, opened 2015-03-23
Vagrant kick over to Dockerfile

jb11211:
Hey Guys,

I saw the discussion around issue 915 and although removing
Vagrant makes sense if no-one is looking after it I thought
I would suggest an alternative.

The associated change will allow vagrant based toolsets to 
continue to run drone, only with some extra dependencies:
* Docker
* vagrant docker provider (std since v1.6 - Apr 2014)

The change itself is boiler plate. grabbed directly from the 
Vagrant docker provider documentation:
http://docs.vagrantup.com/v2/docker/basics.html

Obviously it's up to you, but this change will potentially
break less ppl

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,5 @@
+Vagrant.configure("2") do |config|
+  config.vm.provider "docker" do |d|
+    d.build_dir = "."
+  end
+end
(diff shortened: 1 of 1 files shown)


jb11211:
withdrawing for port forward fix

PR #943, opened 2015-03-31
Supports "to default" ANSII styles

sroze:
This PR supports to two `39m` and `49m` [1] ANSII codes that reset the foreground and background colors.

It simply reset the current styles; this can be improved in the future by managing explicitly foreground and background colors.

[1] http://pueblo.sourceforge.net/doc/manual/ansi_color_codes.html


bradrydzewski:
thanks!

PR #946, opened 2015-04-01
If email is unspecified, pass in -e ' ' to avoid having registry URL interpreted as email, which will fail cryptically.

hkjn:
If publish.docker.email is unspecified, special-case this to pass in -e ' '.

This avoids having the line end up as 'docker login -u foo -p bar -e someregistry', which is interpreted as 'log in user foo with pass bar and email someregistry to https://index.docker.io/v1/' (c.f. 'docker help login').

### DIFF
--- a/plugin/publish/docker.go
+++ b/plugin/publish/docker.go
@@ -94,8 +94,14 @@ func (d *Docker) Write(f *buildfile.Buildfile) {
 
 	// Login?
 	if d.RegistryLogin == true {
+		// If email is unspecified, pass in -e ' ' to avoid having
+		// registry URL interpreted as email, which will fail cryptically.
+		emailOpt := "' '"
+		if d.Email != "" {
+			emailOpt = d.Email
+		}
 		f.WriteCmdSilent(fmt.Sprintf("docker login -u %s -p %s -e %s %s",
-			d.Username, d.Password, d.Email, d.RegistryLoginUrl))
+			d.Username, d.Password, emailOpt, d.RegistryLoginUrl))
 	}
 
 	// Tag and push all tags


vaijab:
This does not work. See the error:

```
time="2015-04-09T09:45:34-04:00" level="fatal" msg="Error response from daemon: Registration: \"Wrong email format (it has to match \\\"[^@]+@[^@]+\\\\.[^@]+\\\")\""
```

hkjn:
Hmm, you're right, though I'm not sure what's going wrong; the command that should be produced works when I run it manually:
```
$ docker login -u someuser -p somepass -e ' ' http://some-registry
Login Succeeded
```

I first wanted to drop `-e` entirely if `email` is omitted, but that makes `docker login` request it interactively:
```
docker login -u username -p password https://registry:8000/v1/
Email: 
```

I tried adding a test case in my local clone (should have included it with this PR..) that verifies that omitting `email` produces `docker login -u username -p password -e ' ' https://registry:8000/v1/`, and the test passes:
https://github.com/hkjn/drone/commit/69ab9e999299fc2869af24167053c3e9cc1072ae

Any ideas for why the command doesn't do the right thing when drone runs it but appears to work manually would be welcome..

vaijab:
> if email is omitted, but that makes docker login request it interactively

Good catch. That means my change I was doing wouldn't work either.

> Any ideas for why the command doesn't do the right thing when drone runs it but appears to work manually would be welcome..

I am not 100% sure, but I think `echo` eats up quotes at  https://github.com/drone/drone/blob/master/shared/build/buildfile/buildfile.go#L57

hkjn:
Hmm, I'm not sure either, but I don't think the `BuildFile.WriteFile` method is being called here.. the `BuildFile.WriteCmdSilent` method just seems to call `WriteString` on the backing `bytes.Buffer` inside `BuildFile`., and the same object eventually writes its `Bytes()` to disk, without any further changes to the contents that I can see:
https://github.com/drone/drone/blob/master/shared/build/build.go#L528

I tried setting up a test case in `github.com/drone/drone/shared/build/build_test.go` as well (not submitting a PR since I had to comment out the `echo '#DRONE:%x'` line in `buildfile.go`, which produced diffs otherwise which I couldn't understand), which 
also verified that the file written to disk contains the intended line, with quotes intact:
```
docker login -u testuser -p testpassword -e ' ' https://test-registry
```

So TL;DR, I don't know why it's not working when actually run by drone; all manual and unit tests seem to say it should.. I could try changing it to double quotes (`-e " "`) instead, but that seems superstitious..

vaijab:
Thanks for looking in to that @hkjn. It is not a blocking bug I guess. I use private registry with drone, so I just put a dummy email address to `.drone.yml` and it works fine.

hkjn:
Yeah, same, so I guess it's no worse than before..

vaijab:
Agreed.

PR #955, opened 2015-04-06
WIP on adding after_success and after_failure

kirillseva:
The goal is to add these tags to drone.yml to run some commands on fail or success,
much like TravisCI

I've never written any Go before, and not yet familiar with Drone codebase,
so I'd welcome critique of the approach I'm taking


bradrydzewski:
with the 0.5 codebase this pr no longer applies. I'm not sure this is required with 0.5 and the new pipeline yaml syntax

PR #963, opened 2015-04-07
[WIP] add set-params command to cli

mohit:
Add set-params command to the cli that allows setting params for repos from a local file.


bradrydzewski:
@mohit can you re-submit here:
https://github.com/drone/drone-cli

we are migrating the CLI and API code to separate repositories and will be deleting the folders in this repo. Thanks!

PR #968, opened 2015-04-10
Add sequencing logic into the InsertBuild and test

benschumacher:
Appreciate review and comments on this effort.

I think it is worth considering a slight refactor of the [utility functions](https://github.com/drone/drone/blob/bolt/datastore/bolt/util.go) to take a `*bolt.Tx` -- a `nil` would indicate that they should create the `bolt.Tx`.

Either this, or BoltDB supporting nested transactions (AFAICT it does not) would simplify these changes, and allow them leverage the `insert` function. I think this signature would work: 

```go
func insert(db *DB, t *bolt.Tx, bucket, key []byte, v interface{}) error {
}
```

As an aside, is there a reason the `get`, `raw`, `update`, `insert`, and `delete` functions should not be defined as methods on the DB type? I suppose this is a style question, but given they all operate on a `*DB`, it seems a natural fit.

Thanks!

### DIFF
--- a/datastore/bolt/build.go
+++ b/datastore/bolt/build.go
@@ -2,6 +2,7 @@ package bolt
 
 import (
 	"bytes"
+	"encoding/binary"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


bradrydzewski:
Looks great, merging.

Good suggestions RE refactoring, I definitely think we could refactor to pass in the transaction as suggested. We may not need to pass in the database object.

Bolt offers a convenience function for creating a transaction and automatically managing commit and rollback which would remove a bunch of boilerplate from our code:

```
err := db.Update(func(tx *bolt.Tx) error {
    ...
    return nil
})
```

so we could maybe do something like this:

```
func (db *DB) InsertBuild(repo string, build *Build) error {
    return db.Update(func(tx *bolt.Tx) error {
        err := get(t, bucket, key, v)
        if err != nil {
           return err
        }
        insert(t, bucket, key, v)
        if err != nil {
           return err
        }
        return nil
    })
}

func insert(t *bolt.Tx, bucket, key []byte, v interface{}) error {
   ...
}
```

It cuts down on the boilderplate and we no longer need to worry about invoking `Commit` and `Rollback` which is a bit tedious.

benschumacher:
I'd seen that convenience function, too. Wasn't sure how far to go initially. I'll create an issue to track the refactor. Thanks.

PR #970, opened 2015-04-10
Rework BoltDB util functions to use bolt.Tx instead of DB

### LINKED ISSUE #969: Database: Refactor utility functions to leverage BoltDB wrappers
The current utility functions could be refactored to take a `*bolt.Tx` instead of `*DB`. This would allow them to be used directly with BoltDB's convenience functions.

An example implementation might look like: (from https://github.com/drone/drone/pull/968#issuecomment-91615322)

```go
func (db *DB) InsertBuild(repo string, build *Build) error {
    return db.Update(func(tx *bolt.Tx) error {
        err := get(t, bucket, key, v)
        if err != nil {
           return err
        }
        insert(t, bucket, key, v)
        if err != nil {
           return err
        }
        return nil
    })
}

func insert(t *bolt.Tx, bucket, key []byte, v interface{}) error {
   ...
}
```

benschumacher:
All existing tests pass. No new tests written. Fixes #969

### DIFF
--- a/datastore/bolt/build.go
+++ b/datastore/bolt/build.go
@@ -15,7 +15,11 @@ import (
 func (db *DB) GetBuild(repo string, build int) (*common.Build, error) {
 	build_ := &common.Build{}
 	key := []byte(repo + "/" + strconv.Itoa(build))
-	err := get(db, bucketBuild, key, build_)
+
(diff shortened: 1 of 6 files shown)


bradrydzewski:
looks great

PR #972, opened 2015-04-13
Use HipChat API ver 2 for notifications

907th:
This PR includes 2 fixes actually:
* jsonmessage was moved from github.com/docker/docker/utils to [github.com/docker/pkg/jsonmessage](https://github.com/docker/docker/tree/master/pkg/jsonmessage). I've just fixed the dependency.
* HipChat API v1 [is deprecated](https://www.hipchat.com/docs/), but github.com/andybons/hipchat (used by Drone) does not support v2. So I've implemented v2 using native http.PostForm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 NOTES.txt
 drone.sublime-project
 drone.sublime-workspace
+Vagrantfile
 .vagrant
 
 *~
--- a/plugin/notify/hipchat.go
+++ b/plugin/notify/hipchat.go
@@ -2,8 +2,10 @@ package notify
 
 import (
 	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
 
-	"github.com/andybons/hipchat"
 	"github.com/drone/drone/shared/model"
 )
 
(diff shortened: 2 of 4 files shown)


907th:
@bradrydzewski I think that the build is failing because Drone CI have cached `docker` repo at `/var/cache/drone/src/github.com/docker/docker`.  Am I right? The build should be OK with the latest `docker` sources.

bradrydzewski:
the problem was (and I'm not sure if it is fixed) that I was unable to `go get ./...` docker because some of the underlying files rely on a script being executed to generate some `.go` files, which caused failures.

so at the moment we rely on cloning a fixed Docker version by running `make docker`. By cloning, as opposed to `go get` it prevented Go from trying to compile and thus won't fail due to the missing files that need to be generated.

it was a total hack

maybe we split into two PRs? One that addresses Hipchat and a second that tries to fix the Docker dependency issue? Might simplify things

907th:
@bradrydzewski Ok, I'll split it tomorrow

907th:
I have splitted it into #973 and #974

PR #974, opened 2015-04-14
Fix dependency: github.com/docker/docker/utils was moved to pkg/jsonmessage

907th:
`utils.DisplayJSONMessagesStream` was moved from `github.com/docker/docker/utils` to [`github.com/docker/pkg/jsonmessage`](https://github.com/docker/docker/tree/master/pkg/jsonmessage). I've just fixed the dependency.


907th:
@bradrydzewski I have removed `make docker` step and it is building with the latest Docker sources with no errors now :)

PR #981, opened 2015-04-17
publish/docker: login should happen early

vaijab:
This fixes an issue where base image which drone is using to build an
image for publishing comes from a private repo.


vaijab:
Any update on this please?

dangra:
`docker login` must happen before `docker build` because of `--pull` right?

I guess your base repository (FROM directive) is private and in the same registry that you are publishing built images.

It still won't work if registry for base image is different to registry used to publish the built image but not a common case.

vaijab:
That's right @dangra

dangra:
+1 to merge :)

Issue #959, opened 2015-04-07
Database Repo Foreign Keys


bradrydzewski:
Deleting a repository should remove all associated keys, params, builds, build tasks and logs:
https://github.com/drone/drone/blob/bolt/datastore/bolt/repo.go#L69

PR #985, opened 2015-04-20
Working on drone/drone#959

benschumacher:
Cleaning up build traces of repo when it is deleted. Some of the cleanup steps may not be necessary for buckets that have been rolled up into the main build document.

### DIFF
--- a/datastore/bolt/repo.go
+++ b/datastore/bolt/repo.go
@@ -132,23 +132,43 @@ func (db *DB) SetRepoKeypair(repo string, keypair *common.Keypair) error {
 
 // DelRepo deletes the repository.
 func (db *DB) DelRepo(repo *common.Repo) error {
-	//TODO(benschumacher) rework this to use BoltDB's txn wrapper
-
-	t, err := db.Begin(true)
-	if err != nil {
-		return err
-	}
 	key := []byte(repo.FullName)
-	err = t.Bucket(bucketRepo).Delete(key)
-	if err != nil {
-		t.Rollback()
+
+	return db.Update(func(t *bolt.Tx) error {
+		err := t.Bucket(bucketRepo).Delete(key)
+		if err != nil {
+			return err
+		}
+		t.Bucket(bucketRepoKeys).Delete(key)
+		t.Bucket(bucketRepoParams).Delete(key)
+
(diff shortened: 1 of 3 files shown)


benschumacher:
Just FYI. I created a separate repo_del_test.go to avoid stepping on @oliveiradan's efforts for repo_test.go. These should be merged together.

datastore/bolt/repo.go
```diff
@@ -132,23 +132,43 @@ func (db *DB) SetRepoKeypair(repo string, keypair *common.Keypair) error {
...
+
+	return db.Update(func(t *bolt.Tx) error {
+		err := t.Bucket(bucketRepo).Delete(key)
+		if err != nil {
+			return err
+		}
+		t.Bucket(bucketRepoKeys).Delete(key)
+		t.Bucket(bucketRepoParams).Delete(key)
+
+		// should we just ignore these error conditions? or should
```
benschumacher: Whoops. I meant to clean this up a bit more, but how should we handle errors. In practice, I think BoltDB is unlikely to fail through this code path, but I wasn't sure if it was reasonable to rollback the entire "delete" operation if one part hiccuped.

datastore/bolt/repo.go
```diff
@@ -132,23 +132,43 @@ func (db *DB) SetRepoKeypair(repo string, keypair *common.Keypair) error {
...
-	t.Bucket(bucketRepoParams).Delete(key)
-	// TODO(bradrydzewski) delete all builds
-	// TODO(bradrydzewski) delete all tasks
-	return t.Commit()
+	})
+}
+
+// deleteTracesOfRepo cleans up build leftovers when a repo is removed
+func (db *DB) deleteTracesOfRepo(t *bolt.Tx, repoKey []byte) error {
+	err := error(nil)
```
benschumacher: re: Note above, this is a particularly egregious waste of code.

bradrydzewski:
great, we'll merge those test files in a follow-up

datastore/bolt/repo.go
bradrydzewski: I think we'll want to handle all errors (even though my initial code did not). I'm cool addressing this now, or coming back in a few weeks and making an effort to cleanup / harden the codebase

PR #990, opened 2015-04-23
SetBuildStatus() and SetBuildTask() index checking

oliveiradan:
1. server/builds.go:92 used SetStatus(). Replaced with SetBuildStatus().
2. Added a index check in SetBuildTask() and return an error in case the tasks number is > than build.tasks size.


bradrydzewski:
hey, it looks like you sent a PR for the wrong branch. These are commits from master

oliveiradan:
@bradrydzewski , sorry I didn't see it before, I will resubmit it for the correct one.

PR #991, opened 2015-04-23
Initial brush at API unit tests for badges

benschumacher:
Starting to get the framework in place for unit testing around the API. This requires a fair amount of set up, particularly if we want to leverages mocks. This is a first pass. Appreciate any feedback.

### DIFF
--- a/datastore/datastore.go
+++ b/datastore/datastore.go
@@ -1,14 +1,15 @@
 package datastore
 
 import (
+	"errors"
 	"io"
 
 	"github.com/drone/drone/common"
 )
 
 var (
-	ErrConflict = "Key not unique"
-	ErrNotFound = "Key not found"
+	ErrConflict    = errors.New("Key not unique")
+	ErrKeyNotFound = errors.New("Key not found")
 )
 
 type Datastore interface {
--- a/datastore/mock/mock.go
+++ b/datastore/mock/mock.go
@@ -1,0 +1,214 @@
+package mock
+
+import (
+	"io"
+
+	"github.com/drone/drone/common"
+	"github.com/drone/drone/datastore"
+)
+
+type Datastore struct {
+	Tokens map[string]*common.Token
+}
+
+func (db Datastore) User(string) (*common.User, error) {
+	return &common.User{}, nil
+}
+func (db Datastore) UserCount() (int, error) {
+	return 0, nil
+}
+
+// UserList returns a list of all registered users.
+func (db Datastore) UserList() ([]*common.User, error) {
+	return []*common.User{}, nil
+}
+
+// SetUser inserts or updates a user.
+func (db Datastore) SetUser(*common.User) error {
+	return nil
+}
+
+// SetUserNotExists inserts a new user into the datastore.
+// If the user login already exists ErrConflict is returned.
+func (db Datastore) SetUserNotExists(*common.User) error {
+	return nil
+}
+
+// Del deletes the user.
+func (db Datastore) DelUser(*common.User) error {
+	return nil
+}
+
+// Token returns the token for the given user and label.
+func (db Datastore) Token(user, label string) (*common.Token, error) {
+	key := string(user + "/" + label)
+
+	if val := db.Tokens[key]; val != nil {
+		return val, nil
+	} else {
+		return nil, datastore.ErrKeyNotFound
+	}
+}
+
+// TokenList returns a list of all tokens for the given
+// user login.
+func (db Datastore) TokenList(string) ([]*common.Token, error) {
+	return []*common.Token{}, nil
+}
+
+// SetToken inserts a new user token in the datastore.
+func (db Datastore) SetToken(*common.Token) error {
+	return nil
+}
+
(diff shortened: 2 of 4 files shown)


benschumacher:
FYI- Torn whether it made more sense to keep mocks closer to the code that's being tested, or closer to the thing they're mocking.... so I did both. ;-)

https://github.com/benschumacher/drone/blob/badge-api-test/datastore/mock/mock.go is the mock for the datastore. It's fairly complicated, since it has a lot of methods, so I thought it might be easier to maintain there. (Also isn't suffixed with _test, so it is part of all builds. This may be undesirable.)

On the other hand, the `MockSession` is much simpler and lives here: https://github.com/benschumacher/drone/blob/badge-api-test/server/mocks_test.go

Again. Appreciate any thoughts on which pattern makes more sense.

bradrydzewski:
Really like the idea of the separate mock package in `datastore/mock`. I think this approach is spot on. We won't be able to use a suffix `_test.go` because tests cannot reference _test files across packages (it would be nice, but isn't a blocker in my mind).

I definitely see your concern regarding a lot of setup / boilerplace code. Check out what the Docker guys are doing with this library. Maybe we can use this same `testify/mock` package as well?
https://github.com/samalba/dockerclient/blob/master/mockclient/mock.go

Instead of setting up a full router and middleware we can probably create the `Content` and manually add the mock objects, and then invoke the handler directly. For example:

```Go
// create gin context
w = httptest.NewRecorder()
c = &gin.Context{Engine: gin.Default()}
c.Writer = w

// add user to the context and invoke handler
u := &common.User{ Login: "octocat" }
c.Set("user", u)
GetUser(c)

// verify response code
g.Assert(w. Code).Equal(200)

// verify expected output
u_ := &common.User{}
json.NewDecoder(w.Body).Decode(u_)
g.Assert(u).Equal(u_) // should do a deep equals
```

This removes quite a bit of setup because we don't need to setup all the routes and middleware. We should (hopefully) be able to independently test middleware using a similar method.

benschumacher:
> Really like the idea of the separate mock package in datastore/mock. I think this approach is spot on. We won't be able to use a suffix _test.go because tests cannot reference _test files across packages (it would be nice, but isn't a blocker in my mind).

Right. `server.MockSession` might work well enough, as I don't think that is likely to get used outside the context of `server`, but the `datastore/mock` is more flexible. I'll refactor this.

>I definitely see your concern regarding a lot of setup / boilerplace code. Check out what the Docker guys are doing with this library. Maybe we can use this same testify/mock package as well?
> https://github.com/samalba/dockerclient/blob/master/mockclient/mock.go

I'll look into it. I probably should have searched a bit a mocking framework for Golang, but given how simple "objects" are, I wasn't sure it was necessary.

> Instead of setting up a full router and middleware we can probably create the `Context` and manually add the mock objects, and then invoke the handler directly.

I had considered this, but was hesitant to open up the internals of Gin since it is a third-party library, and its internal implementation is obviously subject to change. If you're comfortable with this, I'm happy to go ahead with that plan.

> This removes quite a bit of setup because we don't need to setup all the routes and middleware. We should (hopefully) be able to independently test middleware using a similar method.

Agreed. Figured we should create tests for the middlewares, as well, but they're relatively simple, so I was looking towards the bigger prize.

bradrydzewski:
> I'll look into it. I probably should have searched a bit a mocking framework for Golang, but given how simple "objects" are, I wasn't sure it was necessary.

I would have done the same thing. I try to use the standard library as much as possible before adding dependencies to my codebase. I only suggest this mock framework because after 3 years I still haven't found a great approach, and was hoping this framework might help since it seems to be working for the Docker guys.

Of course am open to new ideas or approaches, since I definitely didn't consider all the options.

This was one of the approaches we took with the original Drone codebase, although it still involved quite a bit of boilerplate:
https://groups.google.com/forum/#!starred/golang-nuts/9i01tuVo-1E

> I had considered this, but was hesitant to open up the internals of Gin since it is a third-party library, and its internal implementation is obviously subject to change. If you're comfortable with this, I'm happy to go ahead with that plan.

Good point. We can do version locking (once the Go team [defines a spec](https://groups.google.com/forum/#!topic/golang-dev/nMWoEAG55v8%5B201-225%5D)) which should help mitigate risk of things changing. Testing the http stuff often feels a bit hacky especially if you were coming from a language that had mature mvc frameworks.

benschumacher:
Just an FYI, this doesn't work so well:

> ```go
> // create gin context
> w = httptest.NewRecorder()
> c = &gin.Context{Engine: gin.Default()}
> c.Writer = w
> ```

Turns out that Gin defines a more complex interface for [`gin.ResponseWriter`](https://github.com/gin-gonic/gin/blob/develop/response_writer.go#L19) than what is implemented by the `httptest.NewRecorder`. I kind of believe that Gin is making this more complex. I've got code that works past this, but I'm not sold that it is quite what we want, either.

 I'm going to spin up a separate pull request with the change to make it easier to track differences.

bradrydzewski:
closing in favor of #994

PR #994, opened 2015-04-27
Alternative pattern for API unit tests (versus #991)

benschumacher:
Another pattern for doing API unit tests. This version introduces testify for mocks, though I haven't done much with them yet. Beyond this, we spin up the `gin.Context` directly and bypass a lot of the framework code in Gin.

This got a little complex, because Gin defines their own `gin.ResponseWriter`, so I created the (poorly named) `RecorderImpl` to serve as a shim. It's possible I'm just not doing something quite Go-right, but I had trouble coming up with another solution.


benschumacher:
Whoops. Meant to roll up the changes in the branch.

bradrydzewski:
I'm interested to collect more of your thoughts regarding `testify`. I don't mind the extra mock structures, since we would probably need something regardless of approach.

I'm sure I could submit a pull request to expose gin's internal `responseWriter` for improved testing or mocking. Otherwise, I can ask the guys that created [echo (nearly identical to gin)](https://github.com/labstack/echo) to see if they would be willing to add such functionality.

I'm not crazy about the idea of managing our own writer, but I think it could serve as a temporary workaround until we get something merged into `gin` or `echo`. We could probably shorten the custom `ResponseRecorder` by doing something like this:

```Go
type ResponseRecorder struct {
	*httptest.ResponseRecorder
}

func NewResponseRecorder() *ResponseRecorder {
	return &ResponseRecorder{httptest.NewRecorder()}
}

func (r *ResponseRecorder) Size() int                                    { return r.Body.Len() }
func (r *ResponseRecorder) Status() int                                  { return r.Code }
func (r *ResponseRecorder) WriteHeaderNow()                              { return }
func (r *ResponseRecorder) Written() bool                                { return r.Code != 0 }
func (r *ResponseRecorder) CloseNotify() <-chan bool                     { return nil }
func (r *ResponseRecorder) Hijack() (net.Conn, *bufio.ReadWriter, error) { return nil, nil, nil }
```

benschumacher:
More or less implemented as you commented above, @bradrydzewski. I went somewhat overboard on these, but was experimenting some. `server/badge.go` now has 100% coverage. :wink:

I believe this also provides full coverage of `ccmenu.go` as a pull through, too.

bradrydzewski:
Looking great. One minor comment is to check out this article:
https://github.com/golang/go/wiki/TableDrivenTests

You'll see table-driven tests in many Go projects and in the standard library as well. I think this would be a great use case for this pattern, since we are cycling through different badges / statuses / etc.

bradrydzewski:
I know you plan to keep working on this -- I hope you don't mind but I wanted to merge in the mean time.

bradrydzewski:
I refactored the code a bit to use table-driven tests:
https://github.com/drone/drone/blob/bolt/server/badge_test.go

This pattern is used heavily in the standard libraries' unit tests when testing many permutations.

PR #996, opened 2015-04-29
Dockerfile: Set default database driver and source

ianneub:
This will allow a default docker run to make use of the /var/lib/drone volume without having to set this manually.


MichielDeMey:
+1 for this nice default.

PR #998, opened 2015-04-29
About this commit:

oliveiradan:
1. server/login.go:49 (// TODO(bradrydzewski) return an error message instead). Added error message if authorization fails.
2. server/repos.go:178 (TODO(bradrydzewski) verify repo not exists). Added a checking for the repo and return an error in case it does not exist.

### DIFF
--- a/datastore/builtin/repo_test.go
+++ b/datastore/builtin/repo_test.go
@@ -1,8 +1,10 @@
 package builtin
 
 import (
+	"bytes"
 	"github.com/drone/drone/common"
 	. "github.com/franela/goblin"
+	"io"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -63,7 +65,8 @@ func TestRepo(t *testing.T) {
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "pending"})
-			db.SetLogs(testRepo, 1, 1, []byte("foo"))
+			//db.SetLogs(testRepo, 1, 1, []byte("foo"))
+			db.SetLogs(testRepo, 1, 1, io.Reader(bytes.NewBuffer([]byte("foo"))))
 
 			// first a little sanity to validate our test conditions
 			_, err = db.BuildLast(testRepo)
--- a/datastore/builtin/repo_test.go
+++ b/datastore/builtin/repo_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"github.com/drone/drone/common"
 	. "github.com/franela/goblin"
-	"io"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -66,7 +65,8 @@ func TestRepo(t *testing.T) {
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "pending"})
 			//db.SetLogs(testRepo, 1, 1, []byte("foo"))
-			db.SetLogs(testRepo, 1, 1, io.Reader(bytes.NewBuffer([]byte("foo"))))
+			//db.SetLogs(testRepo, 1, 1, io.Reader(bytes.NewBuffer([]byte("foo"))))
+			db.SetLogs(testRepo, 1, 1, (bytes.NewBuffer([]byte("foo"))))
 
 			// first a little sanity to validate our test conditions
 			_, err = db.BuildLast(testRepo)
--- a/datastore/builtin/repo_test.go
+++ b/datastore/builtin/repo_test.go
@@ -64,8 +64,6 @@ func TestRepo(t *testing.T) {
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "pending"})
-			//db.SetLogs(testRepo, 1, 1, []byte("foo"))
-			//db.SetLogs(testRepo, 1, 1, io.Reader(bytes.NewBuffer([]byte("foo"))))
 			db.SetLogs(testRepo, 1, 1, (bytes.NewBuffer([]byte("foo"))))
 
 			// first a little sanity to validate our test conditions
--- a/datastore/builtin/task.go
+++ b/datastore/builtin/task.go
@@ -16,10 +16,7 @@ func (db *DB) SetLogs(repo string, build int, task int, rd io.Reader) error {
 	if err != nil {
 		return err
 	}
-	//Added
-	//rdBuff := new(bytes.Buffer)
-	//rdBuff.ReadFrom(rd)
-	//log := rdBuff.Bytes()
+
 	log, err := ioutil.ReadAll(rd)
 	if err != nil {
 		return err
--- a/datastore/builtin/task.go
+++ b/datastore/builtin/task.go
@@ -9,12 +9,17 @@ import (
 
 // SetLogs inserts or updates a task logs for the
 // named repository and build number.
-func (db *DB) SetLogs(repo string, build int, task int, log []byte) error {
+func (db *DB) SetLogs(repo string, build int, task int, rd io.Reader) error {
 	key := []byte(repo + "/" + strconv.Itoa(build) + "/" + strconv.Itoa(task))
 	t, err := db.Begin(true)
 	if err != nil {
 		return err
 	}
+	//Added
+	rdBuff := new(bytes.Buffer)
+	rdBuff.ReadFrom(rd)
+	log := rdBuff.Bytes()
+
 	err = t.Bucket(bucketBuildLogs).Put(key, log)
 	if err != nil {
 		t.Rollback()
--- a/datastore/builtin/task.go
+++ b/datastore/builtin/task.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"github.com/boltdb/bolt"
 	"io"
+	"io/ioutil"
 	"strconv"
 )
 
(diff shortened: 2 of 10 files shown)


server/login.go
```diff
@@ -44,14 +44,15 @@ func GetLogin(c *gin.Context) {
 		getLoginOauth2(c)
 	}
 
+	login := ToUser(c)
```
bradrydzewski: if the login fails, attempting to fetch the user will cause a panic

server/login.go
```diff
@@ -44,14 +44,15 @@ func GetLogin(c *gin.Context) {
...
+
 	// exit if authorization fails
-	// TODO(bradrydzewski) return an error message instead
 	if c.Writer.Status() != 200 {
+		log.Errorf("could not get authorization for %s.", login.Login)
+		c.Redirect(303, "/login#error=access_denied")
 		return
 	}
 
-	login := ToUser(c)
```
bradrydzewski: we need this to happen after checking the error

server/repos.go
```diff
@@ -175,7 +175,12 @@ func PostRepo(c *gin.Context) {
 		httputil.GetURL(c.Request),
 	)
 
-	// TODO(bradrydzewski) verify repo not exists
+	_, err_ := store.Repo(name)
+	if err_ != nil {
+		//c.Fail(409, err_)
```
bradrydzewski: remove comment

server/repos.go
```diff
@@ -175,7 +175,12 @@ func PostRepo(c *gin.Context) {
 		httputil.GetURL(c.Request),
 	)
 
-	// TODO(bradrydzewski) verify repo not exists
+	_, err_ := store.Repo(name)
+	if err_ != nil {
```
bradrydzewski: in this case, we want to fail if there is no error (ie `err == nil`) since this means the repository already exists. Also, we should use `err` instead of `err_`

server/repos.go
```diff
@@ -175,7 +175,12 @@ func PostRepo(c *gin.Context) {
...
+		//c.Fail(409, err_)
+		c.Fail(409, fmt.Errorf("Conflict"))
```
bradrydzewski: in this case we can use `c.String(409, "Repository already exists")` ... just shortens the code a bit

server/login.go
oliveiradan: Ok. Do we want to show the user whose authorization was denied. if so, we will need to get another statement inside the block, like: https://gist.github.com/oliveiradan/292460dac1dca1dcef60

server/repos.go
oliveiradan: done.

server/repos.go
oliveiradan: done.

server/repos.go
oliveiradan: done.

server/login.go
bradrydzewski: ToUser(c) will not contain a user object if the login fails. So `failedAuthorization := ToUser(c)` would still cause a panic in a failure scenario. Also, no need to redirect since the below functions would have already redirected:

```
	switch {
	case settings.Service.OAuth == nil:
		getLoginBasic(c)
	case settings.Service.OAuth.RequestToken != "":
		getLoginOauth1(c)
	default:
		getLoginOauth2(c)
	}
```

oliveiradan: Ok, I think I got it then.

server/repos.go
```diff
@@ -175,7 +175,11 @@ func PostRepo(c *gin.Context) {
 		httputil.GetURL(c.Request),
 	)
 
-	// TODO(bradrydzewski) verify repo not exists
+	_, err := store.Repo(name)
```
bradrydzewski: this should be store.Repo(owner+"/"+name) since that is the format we are using

datastore/builtin/task.go
```diff
@@ -9,12 +9,17 @@ import (
...
 // named repository and build number.
-func (db *DB) SetLogs(repo string, build int, task int, log []byte) error {
+func (db *DB) SetLogs(repo string, build int, task int, rd io.Reader) error {
 	key := []byte(repo + "/" + strconv.Itoa(build) + "/" + strconv.Itoa(task))
 	t, err := db.Begin(true)
 	if err != nil {
 		return err
 	}
+	//Added
+	rdBuff := new(bytes.Buffer)
```
bradrydzewski: you can use `ioutil.ReadAll(rd)` as a preferred shorthand

datastore/builtin/repo_test.go
```diff
@@ -63,7 +65,8 @@ func TestRepo(t *testing.T) {
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "pending"})
-			db.SetLogs(testRepo, 1, 1, []byte("foo"))
+			//db.SetLogs(testRepo, 1, 1, []byte("foo"))
+			db.SetLogs(testRepo, 1, 1, io.Reader(bytes.NewBuffer([]byte("foo"))))
```
bradrydzewski: the `bytes.Buffer` implements the `io.Reader` interface, so I'm not sure if you need to wrap it in `io.Reader`

server/queue.go
```diff
@@ -3,7 +3,7 @@ package server
 import (
 	"encoding/json"
 	"io"
-	"io/ioutil"
+	//"io/ioutil"
```
bradrydzewski: remove comments

server/queue.go
```diff
@@ -170,13 +170,18 @@ func PushLogs(c *gin.Context) {
...
 	// instead of a byte array so that we can buffer the write and so that
 	// we avoid unnecessary copies of the data in memory.
-	logs, err := ioutil.ReadAll(io.LimitReader(c.Request.Body, 5000000)) //5MB
-	defer c.Request.Body.Close()
-	if err != nil {
-		c.Fail(500, err)
-		return
-	}
-	err = store.SetLogs(repo.FullName, bnum, tnum, logs)
+	var maxBuffToRead int64 = 5000000 // 5MB.
```
bradrydzewski: in this case we can use `const` instead of `var`. Note that magic numbers aren't necessarily frowned upon in Go, but I think in this case it is fine

bradrydzewski:
great! added some comments above. Also, is there a way to flatten / rebase this commit? It picked up a bunch of my changes over the past few days, making it a bit difficult to review the PR

datastore/builtin/repo_test.go
```diff
@@ -63,7 +64,9 @@ func TestRepo(t *testing.T) {
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "success"})
 			db.SetBuild(testRepo, &common.Build{State: "pending"})
-			db.SetLogs(testRepo, 1, 1, []byte("foo"))
+			//db.SetLogs(testRepo, 1, 1, []byte("foo"))
```
bradrydzewski: remove comments

datastore/builtin/task.go
```diff
@@ -4,17 +4,27 @@ import (
...
 // SetLogs inserts or updates a task logs for the
 // named repository and build number.
-func (db *DB) SetLogs(repo string, build int, task int, log []byte) error {
+func (db *DB) SetLogs(repo string, build int, task int, rd io.Reader) error {
 	key := []byte(repo + "/" + strconv.Itoa(build) + "/" + strconv.Itoa(task))
 	t, err := db.Begin(true)
 	if err != nil {
 		return err
 	}
+	//Added
```
bradrydzewski: remove comments :)

datastore/builtin/task_test.go
```diff
@@ -30,13 +32,15 @@ func TestTask(t *testing.T) {
 
 		g.It("Should set Logs", func() {
 			db.SetRepo(&common.Repo{FullName: testRepo})
-			err := db.SetLogs(testRepo, testBuild, testTask, testLogInfo)
+			//err := db.SetLogs(testRepo, testBuild, testTask, testLogInfo)
```
bradrydzewski: remove comments :)

datastore/mock/mock.go
```diff
@@ -291,7 +291,9 @@ func (m *Datastore) LogReader(_a0 string, _a1 int, _a2 int) (io.Reader, error) {
 
 	return r0, r1
 }
-func (m *Datastore) SetLogs(_a0 string, _a1 int, _a2 int, _a3 []byte) error {
+
+//func (m *Datastore) SetLogs(_a0 string, _a1 int, _a2 int, _a3 []byte) error {
```
bradrydzewski: remove comments :)

server/queue.go
```diff
@@ -170,13 +169,18 @@ func PushLogs(c *gin.Context) {
...
-	logs, err := ioutil.ReadAll(io.LimitReader(c.Request.Body, 5000000)) //5MB
-	defer c.Request.Body.Close()
-	if err != nil {
-		c.Fail(500, err)
-		return
-	}
-	err = store.SetLogs(repo.FullName, bnum, tnum, logs)
+	const maxBuffToRead int64 = 5000000 // 5MB.
+
+	//logs, err := ioutil.ReadAll(io.LimitReader(c.Request.Body, 5000000)) //5MB
```
bradrydzewski: remove comments :)

server/repos.go
```diff
@@ -172,6 +172,16 @@ func PostRepo(c *gin.Context) {
 	name := c.Params.ByName("name")
 
 	// TODO(bradrydzewski) verify repo not exists
+	link := fmt.Sprintf(
+		"%s/api/hook",
+		httputil.GetURL(c.Request),
+	)
```
bradrydzewski: it looks like there may have been a merge conflict? The creation of the link variable is repeated above and below. The link variable at 212 is correct and includes the token where this code (above) should be probably be removed

bradrydzewski:
merged rebased version of this commit from @benschumacher . thanks for all the changes!

PR #1007, opened 2015-05-08
Rebase of @oliveiradan's work in PR #998 

benschumacher:
Rebase of @oliveiradan's work in PR #998.

Fixing TODO comments in code

1. server/login.go:49 (// TODO(bradrydzewski) return an error message instead). Added error message if authorization fails.
2. server/repos.go:178 (TODO(bradrydzewski) verify repo not exists). Added a checking for the repo and return an error in case it does not exist.
3. server/queue.go:170:  // TODO (bradrydzewski) change this interface to accept an io.Reader. All references to the API change been in question SetLogs() have been modified.
4. remote/github/github.go:106  // Fixed a crash in case *repo_.Language is nil , when de-referencing it. This could happen when a repo only has a readme, so github hasn't set the language yet.
5. ./server/queue.go:170:  // TODO (bradrydzewski) change this interface to accept an io.Reader. All references to the API change been in question SetLogs() have been modified.
6. .remote/github/github.go:106  // Fixed a crash in case *repo_.Language is nil , when de-referencing it. This could happen when a repo only has a readme, so github hasn't set the language yet.

### DIFF
--- a/datastore/builtin/repo_test.go
+++ b/datastore/builtin/repo_test.go
@@ -1,6 +1,7 @@
 package builtin
 
 import (
+	"bytes"
 	"github.com/drone/drone/common"
 	. "github.com/franela/goblin"
 	"io/ioutil"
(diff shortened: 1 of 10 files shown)


bradrydzewski:
> Should we just pass the Body directly and let the datastore determine if it need to use a io.LimitReader?

I think that would probably make sense. It would make it part of the datastore implementation and therefore could be customized based on implementation. We can address in a follow-up PR

Thanks for rebasing!

PR #1010, opened 2015-05-12
enable to specify image with hostname

masarakki:
for using private docker repository in `services` .

### DIFF
--- a/shared/build/util.go
+++ b/shared/build/util.go
@@ -68,7 +68,10 @@ func parseImageName(image string) (owner, name, tag string) {
 	}
 
 	parts := strings.Split(name, "/")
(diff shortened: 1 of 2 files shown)


shared/build/util.go
```diff
@@ -68,6 +68,10 @@ func parseImageName(image string) (owner, name, tag string) {
 	}
 
 	parts := strings.Split(name, "/")
+	if len(parts) == 3 {
```
macb: why not just have an else?

```
if len(parts) == 2 {
 		owner = parts[0]
 		name = parts[1]
 } else {
		owner = fmt.Sprintf("%s/%s", parts[0], parts[1])
		name = parts[2]
}
```

PR #1018, opened 2015-05-14
Allow running script for swift plugin

unicell:
This change allows you to run scripts to prepare the release before uploading to swift.


bradrydzewski:
we've moved to a new plugin model in 0.4, where plugins are external docker container. They are stored in separate projects. We have a placeholder for a Swift plugin here if you are interested in helping out! https://github.com/drone-plugins/drone-swift

PR #1019, opened 2015-05-14
New set of Unit Tests for Build, based on 0.4.0

oliveiradan:


datastore/builtin/build_test.go
```diff
@@ -0,0 +1,207 @@
...
+		})
+
+		g.It("NewBuildstore()", func() {
+			repo := common.Repo{
+				UserID: 1,
+				Owner:  "oliveiradan",
+				Name:   "drone-test1",
+			}
+			//Add repo
+			_err1 := rs.AddRepo(&repo)
```
bradrydzewski: do we need to add repositories here?

datastore/builtin/build_test.go
```diff
@@ -0,0 +1,207 @@
...
+			g.Assert(_err3 == nil).IsTrue()
+			g.Assert(repo.ID).Equal(getrepo.ID)
+
+			//Add build
+			build := common.Build{
+				ID:       1,
+				CommitID: 1,
+				State:    "success",
+			}
+			_err1 = bs.SetBuild(&build)
```
bradrydzewski: only use underscores when the variable name conflicts with another variable or a package. For example, if I import the `url` package I cannot have a variable called `url` since they will  clash, so I might choose to name it `url_`

so generally speaking there should be very few veriables that are suffixed with an underscore

datastore/builtin/build_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+		g.It("Build()", func() {
+			repo := common.Repo{
+				UserID: 1,
+				Owner:  "oliveiradan",
+				Name:   "drone-test1",
+			}
+			//Add repo
+			_err1 := rs.AddRepo(&repo)
+			_err2 := rs.SetRepo(&repo)
```
bradrydzewski: can we remove?

datastore/builtin/build_test.go
```diff
@@ -0,0 +1,207 @@
...
+		g.BeforeEach(func() {
+			db.Exec("DELETE FROM blobs")
+			db.Exec("DELETE FROM builds")
+			db.Exec("DELETE FROM commits")
+			db.Exec("DELETE FROM repos")
+			db.Exec("DELETE FROM stars")
+			db.Exec("DELETE FROM tasks")
+			db.Exec("DELETE FROM tokens")
+			db.Exec("DELETE FROM users")
+		})
```
bradrydzewski: we probably don't need to delete data from all tables, since we are only modifying the `builds` table

datastore/builtin/build_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+		g.It("NewBuildstore()", func() {
+			//Add build
+			build := common.Build{
+				ID:       1,
+				CommitID: 1,
+				State:    "success",
+			}
+			err1 := bs.SetBuild(&build)
```
bradrydzewski: although the test is passing, it isn't really testing the intent of the function. The intent of the function is that it will update an existing build in the datastore.

ideally this would insert a build into the datastore, update it, and then fetch it from the datastore to ensure it was updated.

see the user update test for reference:
https://github.com/drone/drone/blob/0.4.0/datastore/builtin/users_test.go#L31

datastore/builtin/build_test.go
```diff
@@ -0,0 +1,148 @@
...
+			g.Assert(err1 == nil).IsTrue()
+		})
+
+		g.It("Build()", func() {
+			build_list := []*common.Build{
```
bradrydzewski: the names used in `g.It` should read more like sentences. The reason is that when tests fail it will print a nice, human readable sentence to the console describing what broke.

you can see an example of what this looks like here:
https://raw.githubusercontent.com/franela/goblin/master/goblin_output.png

see the user tests for reference:
https://github.com/drone/drone/blob/0.4.0/datastore/builtin/users_test.go#L31

datastore/builtin/build_test.go
```diff
@@ -0,0 +1,148 @@
...
+			//
+			err1 := cs.AddCommit(&commit1)
+			g.Assert(err1 == nil).IsTrue()
+			build, err2 := bs.Build(1)
+			g.Assert(err2 == nil).IsTrue()
+			g.Assert(build.ID == 1).IsTrue()
+		})
+
+		g.It("BuildSeq()", func() {
+			build_list := []*common.Build{
```
bradrydzewski: use camel case (instead of snake case)

bradrydzewski:
great, thanks!

PR #1020, opened 2015-05-18
Unit tests for PostToken API

benschumacher:
Initial table-based unit tests for Token creation.

### DIFF
--- a/pkg/server/token.go
+++ b/pkg/server/token.go
@@ -6,22 +6,14 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
 
-	"github.com/drone/drone/pkg/server/session"
 	common "github.com/drone/drone/pkg/types"
 )
 
 // POST /api/user/tokens
 func PostToken(c *gin.Context) {
-	settings := ToSettings(c)
+	sess := ToSession(c)
 	store := ToDatastore(c)
 	user := ToUser(c)
-
-	// if a session secret is not defined there is no way to
-	// generate jwt user tokens, so we must throw an error
-	if settings.Session == nil || len(settings.Session.Secret) == 0 {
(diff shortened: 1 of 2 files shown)


benschumacher:
There is an issue with the invariants in this code, and I took a guess at fixing them, but I may have gone the wrong way. Here's the brief description of the changes to `token.go`:

1.  `server.ToSession(c)` uses gin's `MustGet()`, making a `session.Session` required
2. `PostToken()` handles `settings.Session` being unset/insufficient by [returning an error](https://github.com/drone/drone/pull/1020/files#diff-40b24cc3eaef42cbc89055e5aadc3364R21)

I made a big change that allows the `session.Session` to be created if it doesn't exist, but it may have been simpler to just move the `sess := ToSession(c)` to come after the if condition above. I *believe* that would be sufficient to more closely match our invariants for the remainder of the method.

benschumacher:
(I think I was working from the assumption that `session.Session` mapped to a user's cookie-based session, but upon further reflection and review, this assumption appears to be incorrect.)

bradrydzewski:
hey Ben, if I understand the issue correctly, could the unit test instead invoke `SetSession` to add the session to the context?

benschumacher:
It does, but I want trying to cover branches, too.

My read of the functionality was that if you don't set `settings.Setting.Session.Secret` you wouldn't have a valid `session.Session`. Therefore, to exercise different branches, the table that drives the test is flexible enough to allow for this behavior. Does that make sense?

bradrydzewski:
`settings.Setting.Session.Secret` will fallback to the oauth secret for signing tokens

bradrydzewski:
https://github.com/drone/drone/blob/0.4.0/pkg/settings/settings.go#L121

benschumacher:
Alright. I think test and code match expectations based on changes I missed. Cheers!

PR #1028, opened 2015-05-21
Docker.sock

pdericson:
No expectations of this being merged, just wanted to put my use case out there in case it prompts a better solution :)

### DIFF
--- a/shared/build/build.go
+++ b/shared/build/build.go
@@ -389,8 +389,8 @@ func (b *Builder) run() error {
 	}
 
 	// link docker.sock
-	hostpath = "/var/run/docker.sock"
-	volume = hostpath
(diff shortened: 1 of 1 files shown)


bradrydzewski:
@pdericson I definitely appreciate the pull request. We have an official Docker plugin that will be launched with Drone 0.4 (on or around June 7) using Docker-in-Docker. You can see it here:
https://github.com/drone-plugins/drone-docker

also the updated yaml format will let trusted builds mount their own volumes directly in the yaml:
https://github.com/drone-plugins/drone-docker and
https://github.com/drone/drone/issues/902#issuecomment-77308459

I would definitely love to hear your feedback in a few weeks when 0.4 alpha is released

PR #1032, opened 2015-05-28
Replace pkg/settings with pkg/config

benschumacher:
This is working well enough for me. I took a little liberty with my interpretation of what you wanted w/the agents functionality. I know you're adding Docker Swarm support, as well, but I assumed that the agents functionality should remain...?

I also made the plugins list optional... I think if you want to make it a required value, you'd want to apply after TOML and envconfig are both processed.

Haven't fully tested this, and it'd make sense to write some UTs for it, but it does build and run for me now. I suppose I could have updated the README, too, to reflect the adjusted field names... I'll look at that in the morning.


benschumacher:
Crud. Totally failed to update existing tests, too. Will look at this in the morning, as well. Cheers.

bradrydzewski:
@benschumacher thanks! one minor ask. Can you add the `applyDefaults` function from the old `pkg/settings` to the new `pkg/config`?

benschumacher:
Certainly. I have a handful more changes I haven't pushed either, like actually reading from the `path` provided, instead of just using the local (from literal) `drone.toml`. I'll get those done later today and update the PR.

bradrydzewski:
thanks Ben!

bradrydzewski:
I can't figure out how to upgrade `dpkg`, but until I do, the `.deb` file won't be auto uploaded to S3. So for now please download and install 0.4 from source

PR #1042, opened 2015-06-04
Adds newline at end of file

goober:
I am getting the following error when I try to install drone through docker by executing:

**Relevant part of Dockerfile**
```
# Download Drone.io
RUN wget http://downloads.drone.io/0.4.0/drone.deb
RUN dpkg -i drone.deb
```
**Error**
```bash
dpkg: error processing archive drone.deb (--install):
 name of conffile (starting `/etc/drone/drone.toml') is too long (>1000 characters)
Errors were encountered while processing:
 drone.deb
```

According to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=108196#13 the error occurs when you have missed a newline at the end of the `conffiles`


bradrydzewski:
hmm, I'm seeing an error message at the end of the build

```
dpkg-deb --build dist/drone
dpkg-deb: error: conffile `/etc/drone/drone.toml' does not appear in package
make: [dist] Error 2 (ignored)
```

bradrydzewski:
I agree with the newline fix. We may also need to upgrade `dpkg` as well:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=281562

goober:
Unfortunately that addition is out of my knowledge. Maybe @benschumacher have some input on this?

bradrydzewski:
I can take care of it, in the mean time I'll merge

bradrydzewski:
ok, mystery solved. I was excluding a file in my .gitignore that was expected to be packaged. This has been resolved. I haven't tried installing 0.4 from the debian file, so it may still require some tweaking, but at least it is auto-generating and publishing to s3

goober:
Thanks for your quick reply. I will try with the latest `.deb`

PR #1044, opened 2015-06-05
0.4.0: update Dockerfile implementation

JeanMertz:
* fix $GOPATH usage
* improve layer caching

The caching is not *perfect* yet (I can't get `ADD . .` to stay equal between builds), but it's close enough.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+bin/
+cmd/drone-server/drone_bindata.go
+dist/
+doc/
+
+.dockerignore
+.drone.yml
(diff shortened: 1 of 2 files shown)


JeanMertz:
I added an empty `drone.toml` file to satisfy the requirement of this file being specified. I do think this is an implementation that should be removed, but this works for now:

```bash
% docker run --rm -e DRONE_DATABASE_DRIVER=postgres drone/drone-server
# panic: Could not get DB version: dial tcp 127.0.0.1:5432: connection refused
```

PR #1051, opened 2015-06-08
shared/build/images.go: add go1.4 tag

kshen0:
Noticed there is a Go 1.4 image available on [Docker hub](https://registry.hub.docker.com/u/bradrydzewski/go/tags/manage/) but that image is not listed yet in `images.go`.


bradrydzewski:
thanks!

PR #1055, opened 2015-06-11
ignore GH "push" event when ref is deleted

JeanMertz:


bradrydzewski:
awesome. great catch, thanks!

bradrydzewski:
great catch!

PR #1056, opened 2015-06-12
adapt drone.toml file to 0.4.0 version

djamelfel:


bradrydzewski:
ah great catch, thanks!

PR #1065, opened 2015-06-16
update Godeps for arm architecture

djamelfel:
update gin-gonic/gin and his dependance in order to build drone on arm architecture


bradrydzewski:
0.4 branch now use `GO15VENDOREXPERIMENT` instead of Godeps so no longer applicable

PR #1067, opened 2015-06-19
slack: implement on_change

etix:
Add on_change to push slack notifications only in case of status change (like the email plugin) but still allowing to enforce on_failure to always notify when it fails but only once when it succeeds after a failure.


kri5:
:+1:

predominant:
:+1:

kri5:
Anything blocking this PR to be merged?

bradrydzewski:
merging this now, however, please note that the 0.4 branch (next release) moves all plugins to external projects. This change will therefore need to be ported to the 0.4 release

PR #1079, opened 2015-07-02
add oracle driver

letusfly85:
I added oracle image to shared build images.

I customized oracle startup script to specify hostname and to change hostname in 'tnsnames.ora' and 'listner.ora' below.

https://registry.hub.docker.com/u/letusfly85/oracle-xe/


bradrydzewski:
in preparation for 0.4 release, we no longer need to keep a map of services. You can simply add oracle to the new 0.4 compose section of the Yaml.

PR #1093, opened 2015-07-11
Fixes for db connection and github signup

simonoff:
Fix database connection
Fix logging into Github: Add setting default scope even if it not set in config file.


tboerger:
LGTM

bradrydzewski:
thanks!

Issue #849, opened 2015-01-26
Move privileged flag to yaml file


bradrydzewski:
I would like to move the `Privileged` field to the Docker section of the yaml file:

```
docker:
  privileged: true
  net: host
```

And then change the `Privileged` field to `Trusted`:

```diff
type Repo struct {
-	Privileged  bool   `meddler:"repo_privileged"   json:"privileged"`
+	Trusted     bool   `meddler:"repo_privileged"   json:"trusted"`
```

Note that for compatibility reasons we should leave the database field alone (as `repo_privileged`) to avoid complicating our migrations.

The reasons for this change is that we are adding features (like Volume support) that require elevated privileges in order to enable. Enabling volumes, however, does not mean that someone wants to run the build in a Docker container with privileged mode. This is why I think it makes sense to start differentiating between `trusted` and `privileged`

dave-tucker:
_TL;DR: I don't think flags to elevate container permissions should be present in the YAML unless they can be encrypted._
_Trust should be applied to a build not a repo (unless the flag says ALL builds are trusted)._

 - - -
I'm not sure I like exposing any of the Docker privileges in the YAML file.
I'd be concerned about somebody submitting a PR that changed the permissions and used this to compromise the host system. `root` in the container has pretty much the same permissions as `root` on your Drone server.

Yes, it's possible (like is done today) to not build PR's with these Docker privileges... but consider the case where I have tests that *need* elevated privileges, to access Netlink or `iptables` for example.
In order to keep our master branch ready to ship, we need to build every PR.
But If we build every PR, we'll expose out system to unnecessary risk.
We're in a Catch 22 situation.

My suggestion would be to use the finer grained permissions available in Docker - `--cap-add`, `--cap-drop`, and `--device` - which are much safer. I'm of the opinion that containers booted using these permissions should be safe enough to be used for PR's *but* it's up to the administrator to define what level of risk they are willing to take (this is where trust comes in).

As such, I'm :+1: for separating between `trusted` and `privileged`. I would however argue that I don't think `trust` is as binary as `true|false` when applied to a Repo. I think that the concept of `trusted` should be applied to a `Build` and based on something like https://github.com/drone/drone/issues/838 that would enable us to work out our trust level from the remote's permissions and/or via system wide matrix (which is also configurable).

An example policy might be:
- Don't trust a build with where the container will run with `--privileged=true` unless the user has Admin permission on ACME Corp GitHub organization.
[...]

PR #1098, opened 2015-07-14
trusted builds from private repo or non pull requests can run under privileged mode

shawnzhu:
fixes issue #849 in 0.3.0.

Notice that this introduces a break change where privileged mode is determined by both `privileged` mode in repo setting and `privileged` in `.drone.yml`.

### DIFF
--- a/server/worker/docker/docker.go
+++ b/server/worker/docker/docker.go
@@ -102,15 +102,6 @@ func (d *Docker) Do(c context.Context, r *worker.Work) {
 		log.Printf("Error parsing YAML for %s/%s, Err: %s", r.Repo.Owner, r.Repo.Name, err.Error())
 	}
 
-	// append private parameters to the environment
-	// variable section of the .drone.yml file, iff
-	// this is not a pull request (for security purposes)
-	if params != nil && (r.Repo.Private || len(r.Commit.PullRequest) == 0) {
-		for k, v := range params {
-			script.Env = append(script.Env, k+"="+v)
-		}
-	}
-
 	// TODO: handle error better?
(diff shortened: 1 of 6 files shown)


bradrydzewski:
@shawnzhu thanks for putting some time into this. I think we can find a way to meet your needs, however, I do have some security concerns. The below change means that any public repository using privileged mode would be open to a malicious pull request. I'm not sure we want to expose existing Drone projects to a new attack vector when they may be relying on the existing behavior.

```diff
-Privileged: (b.Privileged && len(b.Repo.PR) == 0)
+Privileged: (b.Privileged && script.DockerPrivileged(b.Build.Docker)),
```

I therefore suggest the following change:

```diff
-Privileged: (b.Privileged && len(b.Repo.PR) == 0)
+Privileged: (b.Privileged && (len(b.Repo.PR) == 0 || b.Repo.Private)),
```

The above change enables privileged mode for pull requests if the repository is private. I think we can safely assume that team members with access to private projects can be trusted (otherwise they wouldn't have access). This also continues to prevents an attacker from issuing a malicious pull request to an open source repository. Hopefully this change would still meet your needs ...

Lastly, I'd prefer not to introduce breaking changes in 0.3 yaml. The `tty` change that we merged did not impact existing projects. Moving `privileged` to the yaml impacts every project using privileged mode. I'd prefer to push breaking changes to the next release unless absolutely necessary.

Please let me know your thoughts. I am very appreciate you rolled up your sleeves and dug into the code :)

tboerger:
What about privileged mode for pull requests if the author is a well known user for drone? :)

PR #1101, opened 2015-07-18
Force worker directory to use slash. Needed to use Drone on windows

richard-julien:
I'm completely new to Go but really interested to contribute to the Drone project. I have a devenv on Windows and this modification permits to correctly setup the git clone remote path and have a fully working solution with boot2docker.


bradrydzewski:
awesome, looking forward to windows support

Issue #1102, opened 2015-07-19
Invalid Docker API Version


arobson:
## Issue
As of Docker 1.7.1, a minimum HTTP API is enforced ([see changelog](https://github.com/docker/docker/blob/e0728035fe09bb6fc4fea03c57a3c241dc055fa8/CHANGELOG.md#remote-api)). Drone is intermittently using an invalid API version specifier when calling the Docker daemon's HTTP API. I suspect that with Docker 1.7.0, this didn't matter. Anyone on Docker 1.7.1 will see specific critical operations fail and make it impossible to generate a build.

## Symptoms
Drone is unable to build _anything_ in the latest 0.3.0 for me. In  the UI, I typically see:
 * `Error: Unable to pull image IMAGE. Bad Request`
 * `Bad Request`

no matter what images I try to use, whether I have already pulled them, nothing works. The console shows:
 `TIMESTAMP Error building COMMIT, Err: Error: Unable to pull image IMAGE. Bad Request`

If I look at the Docker log, I notice that Drone is not using a consistent version number in the path. I see alternation between `v1.9` and `v1.19`. If you `curl` the Docker HTTP API with anything greater than `v1.19`, you'll get the following response:
`client is too old, minimum supported API version is 1.12, please upgrade your client to a newer version`

A search for that error string will show you that it was introduced as of 1.7.1 which is now what you get from their apt repo now.

## Code
Looks like [line 24 in shared/build/docker/client.go](ttps://github.com/drone/drone/blob/90bf8130f8108b8262aa32047870f7d2da6dc372/shared/build/docker/client.go#L24) is the problem.

## Environment
Just as a sanity check, here's the environment I've run across this in:
Ubuntu 14.04
Docker 1.7.1
Drone 0.3.0-alpha (obtained via debian package from: http://downloads.drone.io/master/drone.deb)

arobson:
For what it's worth - I was able to verify that the patch I submitted fixes the issue I was having and I can now get builds from Drone :smile:

bradrydzewski:
Will older daemons complain if we use 1.19? Wondering if we should use 1.12 for max compatibility, based on the error message.

I'm traveling but can merge from my phone if you don't mind testing

arobson:
@bradrydzewski - happy to update the PR to put 1.12 in place of 1.19. I'll also see if I can get an older docker version to work with this.

Poogles:
Any news on this?

bradrydzewski:
[...]

PR #1103, opened 2015-07-19
Change docker version to v1.19 from invalid v1.9

arobson:
This is a very simple attempt to address issue #1102 that I reported earlier. I am not a Go dev but I updated the constant and the related tests and made sure that the changes didn't break existing tests.


bradrydzewski:
thanks!

Issue #1066, opened 2015-06-19
Update data model to support matrix builds, build numbers


bradrydzewski:
This is a design document to finalize the data model in `0.4`. Having a solid data model is extremely important with the introduction of plugins, which require data stability. When plugins are introduced, structural changes the data model will be difficult and infrequent, since it could impact hundreds of plugin repositories.

Each data structure below will map directly to a database table. Some data structures have nested objects (one-to-one relationships, such as Build.Commit.Author) which we can also include in the same table (no need for joins). Nested data does not include one-to-many data structures, which are instead broken out into separate structs and tables. The map data structures, however, will be serialized as JSON and stored in a single column.

Hoping to finalize this over the weekend. **Please note** that the below code is pseudo-code. The final implementation will not use embedded anonymous structures.

**User**

Represents the user object

```Go
type User struct {
	ID     int64
	Login  string
	Token  string // access token
	Secret string // refresh token (optional)
	Email  string
	Avatar string
	Active bool
	Admin  bool
}
```

**Repos**

Represents a repository

```Go
type Repo struct {
	ID       int64
	UserID   int64 // foreign key. associate the repo with creator's github access token
	Owner    string
	Name     string
	FullName string
	Link     string // link to repo in remote system (ie github.com/drone/drone)
	Remote   string // clone url
	Branch   string // default branch
	Private  bool
	Trusted  bool
	Timeout  int64

	Keys struct {
		Public  string
		Private string
	}
	Hooks struct {
		PullRequest bool // enable or disable pull request hooks
		Push        bool // enable or disable push hooks
		Tags        bool
	}

	Params map[string]string // private params not stored in yaml
}
```

**Builds**

Represents a build. Includes commit data and optional pull request data. The optional pull request data includes the base of the working branch, to which the pull request is being merged. This is important for testing pull requests in non-github systems

```Go
type Build struct {
	ID       int64
	RepoID   int64
	Number   int
	Status   string
	Started  int64
	Finished int64

	// Head commit. This would be used to clone pushes
	Commit struct {
		Sha       string
		Ref       string
		Branch    string
		Message   string
		Timestamp string
		Remote    string
		Author struct {
			Login string
			Email string
		}
	}

	// optional, nil if not pull request
	PullRequest struct {
		Title       string
		Number       int

		// Base commit (ie target of pull request). We would clone the Base commit
		// first, then merge the Head commit, and then run our tests. This would be
		// used with Bitbucket and GitLab, since they don't provide a ref for pull requests
[...]

PR #1108, opened 2015-07-23
Start working on specs implementation

Bugagazavr:
- I think we need generate oauth transport on remote side
- also AllowedOrganizations and OpenRegistration moved to remote and called from login via accessor methods
- By default i set sqlite3 driver
- Scope must be moved to settings and by default we set `repo,repo:status,user:email`


pkg/remote/remote.go
```diff
@@ -45,6 +51,27 @@ type Remote interface {
...
+
+	// GetOpen returns boolean field with enabled or disabled
+	// registration.
+	GetOpen() bool
+
 	// Default scope for remote
 	Scope() string
 }
+
+func New(conf *config.Config) (Remote, error) {
```
letusfly85: It seems not to have switch other driver info...
How can we get `DRONE_REMOTE_DRIVER`'s Remote struct? :sweat_drops:

pkg/remote/builtin/github/github.go
```diff
@@ -34,12 +37,14 @@ type GitHub struct {
 
 func New(conf *config.Config) *GitHub {
```
bradrydzewski: we're actually getting rid of the config file all together and moving to a driver-based model. There is a version of this implemented in the database:
https://github.com/drone/drone/blob/0.4-database/pkg/store/store.go#L16

We should do something like `remote.Load(os.Getenv("REMOTE_DRIVER))` which should find the appropriate driver and return a valid remote connection. The remote connection should retrieve its data from environment variables defined here:
http://readme.drone.io/docs/setup/#/github.md

Let me know if that makes sense :)

pkg/remote/builtin/github/github.go
```diff
@@ -281,6 +290,25 @@ func (g *GitHub) push(r *http.Request) (*common.Hook, error) {
 	return &common.Hook{Repo: repo, Commit: commit}, nil
 }
 
+// ¯\_(ツ)_/¯
+func (g *GitHub) Oauth2Transport(r *http.Request) *oauth2.Transport {
```
bradrydzewski: wow, this is a great idea

pkg/remote/builtin/github/helper.go
```diff
@@ -58,7 +58,7 @@ func GetUserEmail(client *github.Client) (*github.User, error) {
 	// WARNING, HACK
 	// for out-of-date github enterprise editions the primary
 	// and verified fields won't exist.
-	if !strings.HasPrefix(*user.HTMLURL, DefaultURL) && len(emails) != 0 {
+	if !strings.HasPrefix(*user.HTMLURL, defaultURL) && len(emails) != 0 {
```
bradrydzewski: I think we can leave this as a constant (`DefaultURL`) because basically we are saying that if github enterprise, and no email, allow the user to proceed.

bradrydzewski:
thanks for tackling this. complete agree with the overall decision.  I do have specific ideas for how I want to register and load drivers. I tried to explain in the comments, but let me know if I did a bad job :)

pkg/remote/builtin/github/github.go
letusfly85: >we're actually getting rid of the config file all together and moving to a driver-based model.

I see, thank you.
So, would the return type of `remote.Load(os.Getenv("REMOTE_DRIVER)) `  be `Remote`?

I will see remote.
And I'm looking forward to get GitLab and Gogs driver models.

bradrydzewski: yep, exactly!

so a remote package (ie `/pkg/remote/github`) would register a `DriverFunc`:

```Go
func init() {
	remote.Register(NewGithubDriver)
}

func NewGithubDriver() (remote.Remote, error) {
	// return github.GitHub with data populated from environment variables
}
```

The `remote` package would keep track of all registered drivers:

```Go
type DriverFunc func() (Remote, error)

func Register(name string, driver DriverFunc) {
	if driver == nil {
		panic("remote: Register driver is nil")
	}
	if _, dup := drivers[name]; dup {
		panic("remote: Register called twice for remote driver " + name)
	}
	drivers[name] = driver
}
```

So when Drone starts we could do something like:

```Go
remote.Load(os.GetEnv("REMOTE_DRIVER"))
```

letusfly85: So, if I want to add GitLab or Gogs drivers,
do I have to code like below?

> `/pkg/remote/gitlab`

```Go
func NewGitLabDriver() (remote.Remote, error) {
    // return gitlab.GitLab with data populated from environment variables
}
```

> `/pkg/remote/gogs`

```Go
func NewGogsDriver() (remote.Remote, error) {
    // return gogs.Gogs with data populated from environment variables
}
```

bradrydzewski: yep, exactly

bradrydzewski: and most of that code already exists in the 0.3 (master) branch. Just needs to be adapted to 0.4

letusfly85: > and most of that code already exists in the 0.3 (master) branch. Just needs to be adapted to 0.4

cool!! :smile:  Can I add gitlab adapter to 0.4?
We want to really use GitLab (or Gogs).

letusfly85: Sorry, we of course wait RFC fixed :sweat_drops: 

https://github.com/drone/drone/issues/1066

bradrydzewski: I think the data model is 97% locked down. There may be some minor tweaks, but overall I'm pretty confident. So yes, we can definitely start integrating GitLab.

We'll probably want to get this PR merged first, since it modifies the `Remote` interface and has some changes that you'll need

Bugagazavr: You want call manually `os.GetEnv` instead `envconfig`?

bradrydzewski: hmm, great question. I was initially thinking of doing something like this so that we could provide default values:

```Go
// create github remote and set required fields
remote := GitHub{}
remote.API = EnvString("GITHUB_HOST", DefaultHost)

// helper function to get environment variable OR default value
func EnvString(name, default) (v string) {
    v = os.Getenv(name)
    if len(v) == 0 {
      v = value
    }
}
```

I'm not opposed to using the `envconfig` library, however.  So whichever method you prefer :)

Bugagazavr: I think, for now we can use envconfig, maybe later we can move to something else.

pkg/remote/remote.go
Bugagazavr: Yes i have a plan to create gitlab, but a little bit later

pkg/remote/builtin/github/github.go
Bugagazavr: Well, i ported the driver system from store, so we need some corrections in specs ( http://readme.drone.io/docs/setup/#/github.md )

need to add `DRONE_GITHUB_API`, if we use gtihub hosted solution, api placed on subdomain, by default we can detect this if `DRONE_GITHUB_HOST` specified, but `DRONE_GITHUB_API` not, also we can show warning for this.

Bugagazavr:
so, next, i think i can start implement gitlab driver

letusfly85:
@Bugagazavr 

I'm looking forward to gitlab, and have one question about authorization, 

it seems that drone 0.4-database uses oauth2.

I tried gitlab driver yesterday in my branch, the authorization failed...
I found below gitlab production.log on debug mode,

```ruby
Doorkeeper::Application Load (2.5ms)  SELECT  `oauth_applications`.* FROM `oauth_applications`  WHERE `oauth_applications`.`uid` = '12d21c3297d1d0c04d5be46424e196948b8544604b873d773a7b1ee0291a2927' LIMIT 1
   (2.7ms)  SELECT COUNT(*) FROM `projects` INNER JOIN `namespaces` ON `projects`.`namespace_id` = `namespaces`.`id` WHERE `namespaces`.`owner_id` = 2 AND `namespaces`.`type` IS NULL
  User Load (2.7ms)  SELECT  `users`.* FROM `users`  WHERE `users`.`email` = 'my.mail@***.co.jp'  ORDER BY `users`.`created_at` DESC, `users`.`id` DESC LIMIT 1
  BroadcastMessage Load (2.6ms)  SELECT  `broadcast_messages`.* FROM `broadcast_messages`  WHERE (ends_at > '2015-07-24 16:02:15' AND starts_at < '2015-07-24 16:02:15')  ORDER BY `broadcast_messages`.`created_at` ASC, `broadcast_messages`.`id` ASC LIMIT 1
```


should we setting somthing on $GITLAB_HOME/config/gitlab.yml for oauth??
it seems default oauth is doorkeeper.

letusfly85:
@Bugagazavr 

If you have a time, please see.

https://github.com/Bugagazavr/drone/pull/1

I tried gitlab..

Bugagazavr:
Thanks, i'll try to see this today.

One tip: gitlab has different oauth scopes,  if i rememer correctly, uses
only `repo` scope
25 июля 2015 г. 11:54 пользователь "Shunsuke Wada" <notifications@github.com>
написал:

letusfly85:
@Bugagazavr 

Thanks, replying.

I remove other scope from `gitlab.go`

and some functions cannot be implement on gitlab.go for remote.go, I think.
because such `Status` doesn't seem to get from https://github.com/Bugagazavr/go-gitlab-client libraries.

If you have notes, please tell me.thank you.

tboerger:
PLEASE integrate gogs side by side with gitea :)

Bugagazavr:
@tboerger gitea supports oauth?

tboerger:
Currently it's still the same as Gogs as it's a fork and changes will maybe follow later on...

Bugagazavr:
So, i think gitea can be added, but as backport gogs from `0.3` ( no auth, no private repository, no deploy keys ).

tboerger:
Tell me what we need for gitea and i can push the implementation ;)

Bugagazavr:
> Tell me what we need for gitea and i can push the implementation ;)

- deploy keys
- oauth optionally

bradrydzewski:
yep, even better than deploy keys, in 0.4 we can clone a repository using the oauth token + netrc file. This allows us to more easily clone submodules, or npm install / go get / install dependencies using version control urls

letusfly85:
@Bugagazavr 

Thank you for replying, https://github.com/Bugagazavr/drone/pull/1, I modified, please see it.

bradrydzewski:
kick ass!

PR #1110, opened 2015-07-25
Citadel cluster support [not finished]

simonoff:


pkg/runner/builtin/runner.go
```diff
@@ -283,11 +235,20 @@ func cname(job *types.Job) string {
...
+		stats := r.Manager.ClusterStats()
+		log.Errorf("Cluster stats:")
+		log.Errorf("CPU: %s", stats.Cpus)
+		log.Errorf("Reserved CPU: %s", stats.ReservedCpus)
+		// If available cpus
+		if stats.Cpus - stats.ReservedCpus > 0 {
+			w := q.Pull()
+			q.Ack(w)
+			log.Errorf("Poll queue")
+			go r.Run(w)
```
bradrydzewski: I'll need to dig deeper, but we should make sure that this function can recover in case of a panic, typically done with `defer recover()`.

pkg/runner/builtin/runner.go
```diff
@@ -283,11 +235,20 @@ func cname(job *types.Job) string {
 
 func (r *Runner) Poll(q queue.Queue) {
 	for {
```
bradrydzewski: we can use a channel here with `select`  and `time.After`

```
select {
    case <-time.After(time.Second * 10):
        ...
    }
```

I think this is the one place citadel falls short, because it doesn't allow us to block until something is available. I think the initial implementation, however, will work and we can continue to improve :)

pkg/runner/builtin/runner.go
```diff
@@ -213,45 +161,49 @@ func (r *Runner) Run(w *queue.Work) error {
...
-			Job:     job,
-			Env:     w.Env,
-			Plugins: w.Plugins,
-		}
-		in, err := json.Marshal(work)
-		if err != nil {
-			return err
-		}
-		workers[i].Notify(in)
-		break
```
bradrydzewski: tracing through the changes. how are notifications being triggered now? Also, how are they being triggered after all jobs are complete, and the build object has been updated?

bradrydzewski:
I think this is a great start, first of all, and I really appreciate you tackling this.

Please keep in mind that I'm not on my primary computer, and am reviewing this from the GitHub UI which may limit my ability. So I pre-apologize if I missed something or am incorrect in some way :)

From what I can tell, the change from build scheduling -> job scheduling introduces race conditions. Jobs are added to the queue, and the queue is processed concurrently in goroutines as they are pulled from the queue (channel). The `Run` function reads and updates the `Build` and `Job` status across goroutines. This means if multiple jobs are running at the same time, they will be accessing the same `Build` and `Job` objects concurrently across goroutines. In Go, this is not a thread safe operation and will lead to race conditions and panics if the timing is just right.

These are two codeblocks that stood out:

https://github.com/simonoff/drone/blob/0.4-database-docker-cluster/pkg/runner/builtin/runner.go#L150:L162
https://github.com/simonoff/drone/blob/0.4-database-docker-cluster/pkg/runner/builtin/runner.go#L65:L72

This is a really nice tool that we can use to help detect and reduce races:
http://blog.golang.org/race-detector

This is one of the reasons I initially coded `Build` as the unit of work instead of `Job`, because it is just more difficult to get right. I do agree that we WANT `Job` as the unit of work, as you've implemented. At the same time, I wonder if it makes sense to tackle this in steps? Maybe citadel + `Build` as unit of measure, and then in next release we tackle `Job`? Just a though .... interested to hear your thoughts

pkg/runner/builtin/runner.go
simonoff: I'm not finished notification part yet :)

simonoff:
The possible fix for such race conditions is making changes to Build though channel too. In any case i will fix notifications first and after will think how to fix race conditions.

bradrydzewski:
We'll also need to do the same for jobs, since those are being accessed across multiple goroutines when looping through the list to view and possibly update status.

There can be no reads from Build or Job (or Jobs) when they are being updated. That even includes things like serializing the JSON payload to send to Docker. So you essentially need to lock before reading and writing to any of those objects.

I also am interested how the notifications will work. Notifications can only be sent once all jobs are complete and the build has been updated. We need to make sure that a worker is available to trigger the notification immediately. We want to avoid a situation where all jobs complete, those workers get re-allocated to a different build, and then the notification container has nowhere to execute. We wouldn't want, for example, a slack notification being triggered 30 minutes after the build is complete because all workers had been re-allocated.

I spent about 3 days weighing the pros and cons of Build vs Job as a unit of work, and sketched out a bunch of designs, and ultimately implemented something simple that I knew we could release and improve over time. I'm definitely looking forward to a fresh perspective, but I do think there are some difficult challenges that need to be solved here. So good luck and let me know how I can help! :)

pkg/runner/builtin/runner.go
bradrydzewski: also, one other question, and this might require some changes to citadel (not sure). I'm wondering if the following will be an issue and how we can prevent ...

build #1 is submitted, all workers are full, goes into 10 second backoff
build #2 is submitted, worker is available, build #1 is sleeping
build #2 executes before build #1

simonoff:
You know all that stuff will be easy if we will use Redis as a storage. In any case just now if drone process die all works in queue will be lost.

simonoff:
In any case i will try to fix race condition if it will be hard then i will back Build queue instead of Jobs queue.

simonoff:
Ah another question. Does drone support transactions in DB? We can do overhead in sql connection to make load of Build from DB.

pkg/runner/builtin/runner.go
simonoff: yes, I not thinked about this yet. It's like a hack for now. Is there is `retry` for Go?

bradrydzewski:
> You know all that stuff will be easy if we will use Redis as a storage.

I completely agree, however, the design constraint is that Drone cannot have any external dependencies. It needs to ship as a single binary file, batteries included. We are optimizing for end-user simplicity at the expense of some code complexity.

> In any case just now if drone process die all works in queue will be lost.

We currently mark running/started builds as failed when the server restarts. In practice, this isn't as big of a problem as you might think. A stable, production Drone installation doesn't frequently restart (or die) so I don't hear many complaints. I do agree, however, that we can and should improve this behavior over time. A better implementation would be to restore the queue and runners from the database on restart.

> In any case i will try to fix race condition if it will be hard then i will back Build queue instead of Jobs queue.

Hopefully I'm just being overly concerned and it will be a piece of cake :)

PR #1123, opened 2015-08-05
add avatar url for gitlab

letusfly85:
we also want to see avatar when using gitlab.

### DIFF
--- a/pkg/remote/builtin/gitlab/gitlab.go
+++ b/pkg/remote/builtin/gitlab/gitlab.go
@@ -73,6 +73,7 @@ func (r *Gitlab) Login(token, secret string) (*common.User, error) {
 	user.Email = login.Email
 	user.Token = token
 	user.Secret = secret
+	user.Avatar = r.URL + "/" + login.AvatarUrl
 	return &user, nil
 }
 
--- a/pkg/remote/builtin/gitlab/gitlab.go
+++ b/pkg/remote/builtin/gitlab/gitlab.go
@@ -73,7 +73,12 @@ func (r *Gitlab) Login(token, secret string) (*common.User, error) {
 	user.Email = login.Email
 	user.Token = token
 	user.Secret = secret
-	user.Avatar = r.URL + "/" + login.AvatarUrl
+
+	if strings.HasPrefix(login.AvatarUrl, "http") {
+		user.Avatar = login.AvatarUrl
+	} else {
+		user.Avatar = r.URL + "/" + login.AvatarUrl
+	}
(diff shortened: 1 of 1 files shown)


bradrydzewski:
@letusfly85 thanks! can you verify that we

bradrydzewski:
@letusfly85 quick question about this line:

```diff
+user.Avatar = r.URL + "/" + login.AvatarUrl
```

Are you sure we need to concatenate `r.URL` and `login.AvatarUrl`? I took a look at the GitLab API [documentation](https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/users.md) and it appears that `AvatarUrl` was the fully qualified URL:

```
[
  {
    "id": 1,
    "username": "john_smith",
    "name": "John Smith",
    "state": "active",
    "avatar_url": "http://localhost:3000/uploads/user/avatar/1/cd8.jpeg",
  }
]
```

letusfly85:
@bradrydzewski 

I got relative path from gitlab api, I'll ask to gitlab team.

Thank you :smile:

bradrydzewski:
great thanks :)

Bugagazavr:
sample

Bugagazavr:
i think relative and non-relative url specified in config

bradrydzewski:
@Bugagazavr thanks for the tip!

Maybe it makes sense to see if `strings.HasPrefix(login.AvatarUrl, "http")` and if `true` we use the value as-is, and if `false` we construct the url as currently implemented in this pull request?

letusfly85:
@bradrydzewski @Bugagazavr 

Thank you! I'll try to fix it.

letusfly85:
@bradrydzewski @Bugagazavr 

I also think that prefix check is good idea.
however, if some events happens to gitlab by relative path, I think we have to consider about it.. :sweat_drops:

bradrydzewski:
@letusfly85 when a user authenticates it should update the URL. This means that if the avatar URL changes Drone should update the next time the user authenticates.

I also think we can also provide a database update statement (as a gist) to bulk update avatar URLs if there is a change to the gitlab URL format, hostname, etc. This probably happens pretty infrequently, so a manual work-around like this is probably good enough for now :)

letusfly85:
@bradrydzewski 

Thank you for replying.
 
It seems infrequently, and if users can upload manually, there is no problem.

So, I think that this PR will be for first authenticates and logged in.

bradrydzewski:
awesome, thanks!

letusfly85:
Thank you!

PR #1131, opened 2015-08-10
Probe for push tag events

Bugagazavr:
In `0.4` we implement this, keep `0.3` working with new client changes


bradrydzewski:
thanks!

PR #1134, opened 2015-08-12
Allow any case/seq to be used for “[CI SKIP]”

elia:
Eg.

[ci skip]
[skipci]
[skip ci]
[CI SKIP]

### DIFF
--- a/server/handler/hook.go
+++ b/server/handler/hook.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"log"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/drone/drone/plugin/remote"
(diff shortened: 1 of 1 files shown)


server/handler/hook.go
```diff
@@ -43,7 +44,9 @@ func PostHook(c web.C, w http.ResponseWriter, r *http.Request) {
 	// in some cases we have neither a hook nor error. An example
 	// would be GitHub sending a ping request to the URL, in which
 	// case we'll just exit quiely with an 'OK'
-	if hook == nil || strings.Contains(hook.Message, "[CI SKIP]") {
+	const ci_skip = `\[(?i:ci *skip|skip *ci)\]`
```
Bugagazavr: const should be declared outside function, go uses camelCase instead snake_case

bradrydzewski:
agreed, looks great, just need to change to camel case :)

elia:
thanks, changed to camelCase and avoided the constant altogether

bradrydzewski:
great, thanks!

PR #1138, opened 2015-08-14
remove unrequired import

kpacha:
the "strings" package is not used so the build is broken


kpacha:
duplicated (#1137). Closing

Great job, by the way!

Issue #412, opened 2014-08-05
more logging for s3 publish


tj:
trying my go 1.3 image, seems to build and create the tarball just fine, same s3 config as before (which was working) however all I get is:

```
$ make build
go get ./...
go build -o build/red-range
./build/red-range --version
0.3.2
tar -zcf build.tgz build
rm -fr build
publishing to Amazon S3 ...
```

and then it's marked as failed. logs don't show anything interesting, appears to complete in the logs actually, but the file is definitely not in s3

```
Successfully built fa13be4eb194
copying repository to /var/cache/drone/src/github.com/segmentio/red-range
starting build 
temp directory is /tmp/drone
removing build container
removing build image
```

PR #1139, opened 2015-08-14
S3 plugin: check dependencies (and improve output)

kpacha:
as @tj pointed out in #412, the lack of logs drives to some annoying build errors. So this PR tries to improve the output of the plugin by checking if `pip` is present.

Also, it checks if the `awscli` package is already available in the container in order to avoid the installation step and use a docker image as small as possible.

Notice that it doesn't use `which`, therefore it doesn't add more dependencies to the container.


bradrydzewski:
we've moved to a new plugin model in the 0.4 branch where plugins are themselves docker containers. As a result, we avoid this issue all together :)

check out https://github.com/drone-plugins/drone-s3 for more info

PR #1141, opened 2015-08-16
Gitlab notifications

kpacha:
send build status updates to gitlab as comments on the related commit

### DIFF
--- a/plugin/notify/gitlab.go
+++ b/plugin/notify/gitlab.go
@@ -1,0 +1,134 @@
+package notify
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/Bugagazavr/go-gitlab-client"
+	"github.com/drone/drone/plugin/remote/gitlab"
+	"github.com/drone/drone/shared/model"
+)
+
+type Gitlab struct {
+	SkipVerify bool `yaml:"skip_verify,omitempty"`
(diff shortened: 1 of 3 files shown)


plugin/notify/gitlab.go
```diff
@@ -0,0 +1,134 @@
...
+
+type GitlabClient interface {
+	SendRepoCommitComment(id string, sha string, body string) (*gogitlab.CommitComment, error)
+}
+
+const (
+	StatusPending = ":raised_hand:"
+	StatusSuccess = ":thumbsup:"
+	StatusFailure = ":thumbsdown:"
+	StatusError   = ":exclamation:"
```
chengweiv5: Add 'Started/Running' status?

kpacha:
@chengweiv5: the Started/Running status is not implemented because this plugin doesn't update the gitlab CI_Server status (https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/models/project_services/ci_service.rb#L44). It just sends a comment and I think it would be too verbose.

For an actual integration, I'd rather build a new gitlab CI_Service.

partikus:
:+1:

bradrydzewski:
cool, merging this now, but just a heads up that 0.4-alpha.1 is going to release next week with a new plugin model and this will need to be manually ported over

PR #1142, opened 2015-08-17
Adding caching

donny-dont:


donny-dont:
Still need to add the override but wanted to get this out there first as I had a couple questions about things.

pkg/yaml/parse.go
```diff
@@ -66,7 +67,7 @@ func ParseOpts(raw string, opts *Opts) ([]*common.Config, error) {
 }
 
 // helper funtion to parse a yaml configuration file.
-func ParseSingle(raw string, opts *Opts) (*common.Config, error) {
+func ParseSingle(raw string, opts *Opts, r *common.Repo) (*common.Config, error) {
 	conf := &common.Config{}
```
bradrydzewski: I'm thinking we take it one step further, and remove transforms from the yaml parser. We could split these into separate calls:

```
conf := yaml.ParseSingle(injected)
err := yaml.Lint(conf)
err := transform.Defaults(conf, repo)
```

but we can save this task for a later time. More important to get the feature working first, and then we can invest more time into some additional refactoring, if we even think it still makes sense.

I also think, as you mentioned, we can remove the `opts` all together and then just run `transform.Defaults` and optionally `transform.Safe` depending on if `repo.Trusted==true` or not.

Issue #1145, opened 2015-08-18
Secure environment variables encrypted in Yaml


bradrydzewski:
Currently we have the ability to store private environment variables in the database. I'd like to expand this capability to store encrypted environment variables in the yaml itself, similar to travis.

```yaml
secure:
  docker_user: U2FsdGVkX19g96OdOIRPtVNGzPP1CgGlxRxRVDvlVHZKYrulqPLlYFG3TfUw53Gk
  docker_pass: U2FsdGVkX18B3tY76uPxLe2tA6UzaBrVMMsArAUS4bo=
```

the secure parameters are decrypted at runtime and injected into your build yaml, similar to how we inject private variables (entered in the UI) or matrix parameters:

```yaml
deploy:
  docker:
    email: $$docker_user
    password: $$docker_pass
```

Unlike travis, injecting the private parameters as environment variables is not sufficient since publish, deploy and notify sections (which are typically the only sections that require such secrets) are executed outside of the build container and expect very specific parameters in the yaml. This is a result of the new 0.4 plugin model.

So for now we need to inject using the same syntax as matrix and private variables. I will spend some time, however, thinking up alternate approaches ...

bradrydzewski:
cc @shawnzhu

bradrydzewski:
note that this implementation uses AES256 encryption because all parameters are encrypted via the endpoint. We use a per-repo encryption key (repo.Hash as the secret) that never leaves drone or is exposed to any end user or any build.

I'm hesitant to use rsa encryption because the [`rsa.EncryptPKCS1v15`](http://golang.org/pkg/crypto/rsa/#EncryptPKCS1v15) function is considered insecure. The benefit of RSA is that we could encrypt with only the public key, and it could be done outside of Drone on a developers laptop, without sending private data to the Drone server. At the same time, Drone has the private key and can always decrypt, so I'm not sure it really buys much more security.

tboerger:
And how do we get the encrypted value that have to be placed in the yaml?

shawnzhu:
@tboerger the idea is creating a new sub-command for drone CLI to get the encrypted version:

```shell
$ drone encrypt <owner/repo-name> <sensitive-value>
U2FsdGVkX18B3tY76uPxLe2tA6UzaBrVMMsArAUS4bo=
```

tboerger:
great if users can access the cli command standalone :)

shawnzhu:
[...]

PR #1154, opened 2015-08-20
AES -> RSA-OAEP

shawnzhu:
proposal to #1145

encrypts secured variables via repo public key while decrypted via repo private key.

notice that it uses sha256 checksum of repo Hash for RSA-OSAP encryption/decryption, so repo Hash is never leaked to client side.

### DIFF
--- a/pkg/utils/sshutil/sshutil.go
+++ b/pkg/utils/sshutil/sshutil.go
@@ -4,7 +4,9 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/pem"
+	"hash"
 
 	"github.com/drone/drone/Godeps/_workspace/src/code.google.com/p/go.crypto/ssh"
 )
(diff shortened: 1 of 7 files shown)


bradrydzewski:
> notice that it uses sha256 checksum of repo Hash for RSA-OSAP encryption/decryption, so repo Hash is never leaked to client side.

really smart idea!

pkg/yaml/secure/secure.go
```diff
@@ -1,79 +1,36 @@
...
 	"github.com/drone/drone/Godeps/_workspace/src/gopkg.in/yaml.v2"
+
+	common "github.com/drone/drone/pkg/types"
+	"github.com/drone/drone/pkg/utils/sshutil"
 )
 
 // Parse parses and returns the secure section of the
 // yaml file as plaintext parameters.
-func Parse(key, raw string) (map[string]string, error) {
+func Parse(repo *common.Repo, raw string) (map[string]string, error) {
```
bradrydzewski: since the data model has changed a bit (and may still change) I'd like to avoid sending the repo model to the `secure` package. Do you think we could do this instead?

```diff
-func Parse(repo *common.Repo, raw string)
+func Parse(privateKey, hash, raw string)
```

pkg/utils/sshutil/sshutil.go
```diff
@@ -38,15 +40,35 @@ func MarshalPrivateKey(privkey *rsa.PrivateKey) []byte {
...
 }
 
-// helper function to encrypt a plain-text string using
+// Decrypt is helper function to encrypt a plain-text string using
 // an RSA public key.
-func Decrypt(privkey *rsa.PrivateKey, secret string) (string, error) {
-	msg, err := rsa.DecryptPKCS1v15(rand.Reader, privkey, []byte(secret))
-	return string(msg), err
+func Decrypt(hash hash.Hash, privkey *rsa.PrivateKey, secret string) (string, error) {
+	decoded, err := base64.StdEncoding.DecodeString(secret)
```
bradrydzewski: wondering if we should use `base64.URLEncoding` vs `base64.StdEncoding`. I don't have a strong preference, and am fine leaving it as-is, but didn't know if there were any pros / cons either way

bradrydzewski:
added some minor comments, but this is looking really good. If you don't mind making a few adjustments, I think we can get it merged right away. Thanks so much ... really great work.

pkg/yaml/secure/secure.go
shawnzhu: sure. fixing

pkg/utils/sshutil/sshutil.go
shawnzhu: from godoc of `encoding/base64`:
> StdEncoding is the standard base64 encoding, as defined in RFC 4648.
> URLEncoding is the alternate base64 encoding defined in RFC 4648. It is typically used in URLs and file names.

I guess there's no difference when encoding RSA key file into PEM format.

PR #1159, opened 2015-08-22
GitLab service interface

Bugagazavr:
Allow to work with gitlab as service interface

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1184


Bugagazavr:
Currently gitlab not support it, but if gitlab merge my MR, this can be merged too

bradrydzewski:
@Bugagazavr cool. In the mean time, we should capture enough data in the `Build` object to clone without merge ref. This is what the `Build` object looks like when we receive a pull request from GitHub which includes enough do manually clone origin and upstream ...

```json
{
  "ID": 10,
  "number": 3,
  "status": "success",
  "started_at": 1440022534,
  "finished_at": 0,
  "head_commit": {
    "sha": "ededa54d3968051ef18ae3bebe2903bd1bb061d0",
    "ref": "refs\/pull\/1\/merge",
    "branch": "fix\/clone-into-non-empty-dir",
    "timestamp": "2015-08-19 22:04:07.457360262 +0000 UTC",
    "remote": "https:\/\/github.com\/donny-dont\/drone-git.git",
    "author": {
      "login": "donny-dont"
    }
  },
  "pull_request": {
    "number": 1,
    "title": "Use git init instead of git clone",
    "base_commit": {
      "sha": "7138816569c347afef5f62d899a1677b59037fd6",
      "ref": "master",
      "remote": "https:\/\/github.com\/drone-plugins\/drone-git.git",
      "author": { }
    }
  }
}
```

just wanted to throw out as an option in case it takes a while for your pull request to be accepted by GitLab. Also, we'll need similar logic for Stash, Bitbucket, etc anyway.

Bugagazavr:
How you want to clone source repository without ref pathspec?

bradrydzewski:
using the above json example, we could do something like this:

```sh
git fetch origin # this would pull from drone/drone
git fetch https://github.com/donny-dont/drone-git.git fix/clone-into-non-empty-dir
git checkout FETCH_HEAD
```

I'm sure we can refine the above to even clone the specific sha. Everything we need to clone the pull request should be in the payload. If you look at the sample json you'll see the `head_commit` for a pull request contains most of the relevant data, including the remote url, branch and sha.

The ref doesn't have to be `refs/pull/1/merge`. That is something we do specifically for GitHub. We don't have to use the same approach for other remote systems. For Bitbucket, we'll need to use the actual ref from the merge request commit

Bugagazavr:
Github, Gitlab, Stash support refs, Bitbucket - not, Your case works only for public repos, but not for:

- Private -> Private
- Private -> Public

Ofc if you use netrc, maybe you can have access to some private repositories, but maybe not, there is no guarantee that source repo will be accessible

bradrydzewski:
> Github, Gitlab, Stash support refs, Bitbucket - not

Sorry, I think I misunderstood. I thought that refs weren't supported in GitLab until your pull request was merged

bradrydzewski:
I would, however, like to split this into two separate pull requests. Let's merge the gitlab.go changes right away to enable pull request functionality. I would like to understand and discuss some of the other REST and database-related changes

Bugagazavr:
> Sorry, I think I misunderstood. I thought that refs weren't supported in GitLab until your pull request was merged

Since `7.14.0` supported, if my pull request has been merged:
  - Drone can activate self as gitlab service via gitlab API
  - Drone no need add deploy key and add manually hook
  - Drone can clone repository with repository token as `drone-ci-token:drone-repo-token`

>  Let's merge the gitlab.go changes right away to enable pull request functionality. I would like to understand and discuss some of the other REST and database-related changes

thats a breaking functionality, after merge this we need reactivate, all builds to get it works

**Database changes**: 

1. added a sql queries to get last build from pull request number 
2. added a sql query to get build from sha and branch.

**REST changes**: 

1) added method for gitlab, to get current build status protected by repository token

```
/api/repos/:owner/:name/pr/:number
```

2) added method, to got a redirect to build page from pull request number

```
/redirect/:owner/:name/pr/:number
```

3) added method, to get a redirect to build page from she and branch

```
/redirect/:owner/:name/pr/:number
```

**Remote changes**:

1. Removed hook and deploy key manipulations
2. Added service API instead hook and deploy key ( faster, easy )
3. Added support for push tag ( requires reactivation, because by default in 0.4.0 will be disabled ) and merge request events
4. Gitlab remote tests

I have a plan come to september ( from 1st to 7th ) merge window, if not, i create a pull request to drone with only push tags and pull requests support

pkg/store/builtin/build_sql.go
```diff
@@ -596,6 +596,79 @@ WHERE build_repo_id = ?
 AND build_number = ?
 `
 
+const stmtBuildSelectPullRequestNumber = `
+SELECT
```
bradrydzewski: since the build_sql.go is auto-generated, let's move these non-generated queries to `build.go` so they don't overwritten. I haven't published the sql generator yet, but I'll try to do that next week :)

also, I will probably adjust some of the non-generated queries I wrote to use the following:

```Go
const stmtBuildSelectPullRequestNumber = stmtBuildSelectList + `
WHERE build_repo_id = ?
AND build_pull_request_number = ?
ORDER BY build_number DESC
LIMIT 1
`
```

... the benefit to the above is that we don't have to worry about column changes in our non-generated queries

cmd/drone-server/drone.go
```diff
@@ -185,6 +185,14 @@ func main() {
 			repo.GET("/logs/:number/:task", server.GetLogs)
 			// repo.POST("/status/:number", server.PostBuildStatus)
 		}
+
+		// Routes for external services
+		repoExternal := repos.Group("")
+		{
+			repoExternal.Use(server.SetRepo())
+
+			repoExternal.GET("/pr/:number", server.GetPullRequest)
```
bradrydzewski: let's do `/pulls/:number` to more closely align with GItHub, if you don't mind :)

also, this can be put in the above group. The reason is that none of the middleware (security, etc) will be applied unless it sits in its own security group.

```diff
			repo.GET("/logs/:number/:task", server.GetLogs)
-		}
-
-		// Routes for external services
-		repoExternal := repos.Group("")
-		{
-			repoExternal.Use(server.SetRepo())
-			repoExternal.GET("/pr/:number", server.GetPullRequest)
+			repo.GET("/pulls/:number", server.GetPullRequest)
+		}
```

we'll also be able to add `repoExternal.GET("/pulls", server.GetPullRequests)` which I know @donny-dont would like as well

cmd/drone-server/drone.go
```diff
@@ -236,6 +244,14 @@ func main() {
 		auth.POST("", server.GetLogin)
 	}
 
+	redirects := r.Group("/redirect")
+	{
+		redirects.Use(server.SetDatastore(store))
+		redirects.Use(server.SetRepo())
+
+		redirects.GET("/:owner/:name/commit/:sha", server.RedirectSha)
```
bradrydzewski: I'd like to move this function up w/ the repo group instead if you don't mind. And also use `/s/commit/commits/g`

```diff
repo.GET("/pulls/:number", server.GetPullRequest)
+repo.GET("/commits/:sha", server.GetCommit)
```

pkg/server/redirect.go
```diff
@@ -0,0 +1,61 @@
...
+		branch = repo.Branch
+	}
+
+	build, err := store.BuildSha(repo, sha, branch)
+	if err != nil {
+		c.Redirect(301, "/")
+		return
+	}
+
+	c.Redirect(301, fmt.Sprintf("/%s/%s/%d", repo.Owner, repo.Name, build.ID))
```
bradrydzewski: I added some notes about slightly altering the route. Since we have the build, let's just to `c.Json` instead of `c.Redirect` since I think this may be a userful endpoint for others as well

bradrydzewski:
Thanks for the explanation, sorry I didn't mention this previously, but there is a ton of great work here!  I've added some (hopefully) minor notes in the code review section. Once those are complete we are good to merge. Also, you get bonus points for porting the unit tests back over to the 0.4 branch :)

I'll take care of updating the swagger file to include the new endpoints once this is complete. It is a pain in the ass to update, and I don't want to punish you after submitting such a nice pull request :)

cmd/drone-server/drone.go
Bugagazavr: But this need to access from CI without user session, the same mechanism used in hook handler

bradrydzewski: ok, for now, let's just make the path changes (`s/pr/pulls/g`) and I'll separately work on an improved token implementation so that we can combine the two at some point in the future.

cmd/drone-server/drone.go
Bugagazavr: but this is a internal pull request number in repo scope

bradrydzewski: I guess I'm still a bit confused by all of this ... it seems we need to add a ton of GitLab-specific code to Drone? I guess I don't understand what all of this code does since I don't use GitLab. Can you please further explain in the comments why all of this is required? Thanks!

Bugagazavr: Ok. GitLab doesn't provide any method to set merge request status from CI, gitlab know about service ( url patterns, sended data ), and try to fetch build status. After status has been fetched, gitlab shows build status.

screen shot 2015-08-24 at 21 43 15

As we see, we have a link "View build details", after click we need to drone build page ( but 0.4.0 not uses SHA ). After click we go to redirect handler, and handler searches commit with sha and branch in repos scope, and finally build find, we redirect user to frontend build page.

pkg/server/redirect.go
Bugagazavr: This handler for redirection only, `/:owner/:name/commit/:sha`, we use this to redirect user when they click to "View build details"

cmd/drone-server/drone.go
bradrydzewski: great, thanks for the explanation!  Is it possible for GitLab to query the API prior to redirecting? GitLab could hit the Drone API, fetch the build by commit to get the build number, and then route the user to the appropriate URL?

Bugagazavr: Build url created before gitlab fetch build status :(

bradrydzewski: how is the build URL created? Is this something we can pass back in the Hook response?

Bugagazavr: As you can see, all possible build states generated when CI detected

https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/views/projects/merge_requests/widget/_heading.html.haml#L3-13

and after initialized javascript https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/views/projects/merge_requests/widget/_heading.html.haml#L27-29

javascript create request to gitlab **merge request controller**m to method **ci_status** 

https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/controllers/projects/merge_requests_controller.rb#L179

After, called current **ci_service** ( for example drone ) method **merge_request_status** 

https://gitlab.com/bugagazavr/gitlab-ce/blob/72b32c8e8d82438f5981e3a7075b6c841077c394/app/models/project_services/drone_ci_service.rb#L43

And ajax got a status, than javascript make a visible, previously generated container with status

bradrydzewski: how do they handle Jenkins integration or Travis integration, which both use build numbers?

Bugagazavr: Jenkins it's a part Enterprise edition, i don't know, travis doesn't support gitlab

bradrydzewski: Relying on sha or pull request number does have some consequences. This makes eventually replacing sqlite with Bolt a bit more complicated since we would need to manually maintain indexes of commit sha and pull request numbers.

I would like to explore all other options if possible. Although it sounds like there may be no other options :(

martinsefcik: @Bugagazavr 
Jenkins integration is only in Enterprise edition, but source code is publicly visible. So you can check, how it is implemented:
https://gitlab.com/gitlab-org/gitlab-ee/blob/master/app/models/project_services/jenkins_service.rb

Bugagazavr: @bradrydzewski so jenkins uses sha too https://gitlab.com/gitlab-org/gitlab-ee/blob/master/app/models/project_services/jenkins_service.rb#L55

Bugagazavr: @martinsefcik thanks

Bugagazavr:
@bradrydzewski so i have finally complete this PR.

but @DouweM requests to merge this PR before gitlab merge own ( https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1215  ), this breaks current compatibility, and can be used after september 22 gitlab 8.0 release. Need to add description about this and create current drone snapshot if someone wanna to play with drone and gitlab less then 8.0 (:

pkg/remote/builtin/gitlab/gitlab.go
```diff
@@ -259,20 +230,66 @@ func (r *Gitlab) Hook(req *http.Request) (*common.Hook, error) {
...
+	hook.Repo.Link = re.ReplaceAllString(parsed.ObjectAttributes.Target.HttpUrl, "$1")
+	hook.Repo.Clone = parsed.ObjectAttributes.Target.HttpUrl
+	hook.Repo.Branch = "master"
+
+	hook.Commit = &common.Commit{}
+	hook.Commit.Message = parsed.ObjectAttributes.LastCommit.Message
+	hook.Commit.Sha = parsed.ObjectAttributes.LastCommit.Id
+	hook.Commit.Remote = parsed.ObjectAttributes.Source.HttpUrl
+
+	if parsed.ObjectAttributes.Source.HttpUrl == parsed.ObjectAttributes.Target.HttpUrl {
```
DouweM: It's probably safer to compare `source_project_id` and `target_project_id`.

Bugagazavr: yep, it most simple compare then:

```go
itarget.Name == source.Name && target.Namespace == source.Namespace
```

DouweM: @Bugagazavr :+1:

pkg/remote/builtin/gitlab/gitlab.go
```diff
@@ -259,20 +230,66 @@ func (r *Gitlab) Hook(req *http.Request) (*common.Hook, error) {
...
+		hook.Commit.Timestamp = time.Now().UTC().Format("2006-01-02 15:04:05.000000000 +0000 MST")
 	}
 
+	hook.Commit.Author = &common.Author{}
+	hook.Commit.Author.Login = parsed.ObjectAttributes.LastCommit.Author.Name
+	hook.Commit.Author.Email = parsed.ObjectAttributes.LastCommit.Author.Email
+
+	hook.PullRequest = &common.PullRequest{}
+	hook.PullRequest.Number = parsed.ObjectAttributes.IId
+	hook.PullRequest.Title = parsed.ObjectAttributes.Description
```
DouweM: Do we want `.Description` or `.Title`?

pkg/remote/builtin/gitlab/gitlab.go
```diff
@@ -259,20 +230,66 @@ func (r *Gitlab) Hook(req *http.Request) (*common.Hook, error) {
...
+	hook.Commit.Remote = parsed.ObjectAttributes.Source.HttpUrl
+
+	if parsed.ObjectAttributes.Source.HttpUrl == parsed.ObjectAttributes.Target.HttpUrl {
+		hook.Commit.Ref = fmt.Sprintf("refs/heads/%s", parsed.ObjectAttributes.SourceBranch)
+		hook.Commit.Branch = parsed.ObjectAttributes.SourceBranch
+		hook.Commit.Timestamp = time.Now().UTC().Format("2006-01-02 15:04:05.000000000 +0000 MST")
+	} else {
+		hook.Commit.Ref = fmt.Sprintf("refs/merge-requests/%d/head", parsed.ObjectAttributes.IId)
+		hook.Commit.Branch = parsed.ObjectAttributes.SourceBranch
+		hook.Commit.Timestamp = time.Now().UTC().Format("2006-01-02 15:04:05.000000000 +0000 MST")
```
DouweM: Can't you get this from `parsed.ObjectAttributes.LastCommit`?

DouweM: Also, I don't think this and the line above need to be in both branches of the if/else statement.

pkg/remote/builtin/gitlab/gitlab.go
```diff
@@ -259,20 +230,66 @@ func (r *Gitlab) Hook(req *http.Request) (*common.Hook, error) {
...
 
-	if len(parsed.After) == 0 || parsed.TotalCommitsCount == 0 {
-		return nil, nil
-	}
+	switch parsed.ObjectKind {
+	case "merge_request":
+		if parsed.ObjectAttributes.State != "reopened" && parsed.ObjectAttributes.MergeStatus != "unchecked" ||
+			parsed.ObjectAttributes.State != "opened" && parsed.ObjectAttributes.MergeStatus != "unchecked" {
+			return nil, nil
+		}
```
DouweM: This check is now unnecessary since GitLab already applies it.

DouweM:
@Bugagazavr https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1215 implements access over HTTP with `drone-ci-token`/`<token>` as username/password, but I don't see that used anywhere here. What am I missing?

Bugagazavr:
@DouweM you use incorrect basic auth look this:

```
#git clone http://drone-ci-token:121212@localhost:3000/gitlab-org/gitlab-ci.git
Cloning into 'gitlab-ci'...
remote: Counting objects: 9723, done.
remote: Compressing objects: 100% (3570/3570), done.
remote: Total 9723 (delta 5862), reused 9717 (delta 5859)
Receiving objects: 100% (9723/9723), 2.50 MiB | 0 bytes/s, done.
Resolving deltas: 100% (5862/5862), done.
Checking connectivity... done.
```

pkg/remote/builtin/gitlab/gitlab.go
```diff
@@ -259,20 +230,66 @@ func (r *Gitlab) Hook(req *http.Request) (*common.Hook, error) {
...
+func mergeRequest(parsed *gogitlab.HookPayload, req *http.Request) (*common.Hook, error) {
+	var hook = new(common.Hook)
+
+	re := regexp.MustCompile(".git$")
+
+	hook.Repo = &common.Repo{}
+	hook.Repo.Owner = req.FormValue("owner")
+	hook.Repo.Name = req.FormValue("name")
+	hook.Repo.FullName = fmt.Sprintf("%s/%s", hook.Repo.Owner, hook.Repo.Name)
+	hook.Repo.Link = re.ReplaceAllString(parsed.ObjectAttributes.Target.HttpUrl, "$1")
```
DouweM: If you want, you can add `web_url` to the project's `hook_attrs` in your MR. The method already exists and it's already returned by the API. It's a little cleaner than doing this replacement here :)

DouweM:
@Bugagazavr I understand how Basic Auth works, but I don't see the `drone-ci-token` username used anywhere in this PR?

Bugagazavr:
in this PR no, because it works on nettc method and we need to pass repository data to netrc method ( it requires a change github remote plugin, because used in one interface ), this pull request is too large i planed do this after this merged.

DouweM:
@Bugagazavr All right, I hope you'll get everything ready and merged into Drone before GitLab 8.0 is released :)

pkg/remote/builtin/gitlab/gitlab.go
Bugagazavr: Im not a sure how timestamp has been parsed, ok im try to use it

pkg/remote/builtin/gitlab/gitlab.go
Bugagazavr: looks like hook does not contains `web_url` key

DouweM: Yeah, I was suggesting that you can add it to the GitLab `hook_attrs` method on `Project`, so that you can use it here.

Bugagazavr:
@DouweM thanks for the review, I have tried to fulfill all your wishes (:

so, i have find a bug in drone model, i have plan to cover drone model tomorrow, now i send minor fixes to gitlab

Bugagazavr:
@bradrydzewski hi brad, any news, on this?

bradrydzewski:
@Bugagazavr sorry for the delay on my end. The PR looks awesome. I do have 1 final request, and I promise we can merge :)

I'd like to take the gitlab-specific routes and wrap them in a `/gitlab` router group. This is similar to what the Jenkins implementation does as well. For example:

```
	gitlab := r.Group("/gitlab")
	{
		// all gitlab-specific routes here
	}
```

I'd also like to move all GitLab routes into a `server/gitlab.go` file. The primary reason is that I'm worried that myself (or someone else) won't know what these routes are for, and will end up changing or removing them. By wrapping in a `/gitlab` router group, and by putting in a `server/gitlab.go` file it will help caution anyone from changing.

I hope that makes sense, and that it isn't too big of a change. Thanks again!

Bugagazavr:
@bradrydzewski done

bradrydzewski:
awesome, thanks!!!!!

PR #1166, opened 2015-08-26
Fix: DRONE_SERVER_PORT isn't valid for drone 0.4

chengweiv5:
Signed-off-by: Chengwei Yang <yangchengwei@qiyi.com>


bradrydzewski:
fixed in newer 0.4 Dockerfile

PR #1178, opened 2015-09-01
Fix hook check logic

chengweiv5:
Signed-off-by: Chengwei Yang <yangchengwei@qiyi.com>


bradrydzewski:
great catch, thanks so much!

PR #1179, opened 2015-09-01
Make.go improvements

tboerger:


make.go
```diff
@@ -283,19 +356,44 @@ func clean() error {
...
+// run is a helper function that executes commands
+// and assigns stdout and stderr targets
+func run(command string, args ...string) error {
+	cmd := exec.Command(command, args...)
+
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	trace(cmd.Args)
+	err := cmd.Run()
```
bradrydzewski: we can tighten this up -- `return cmd.Run()`  :)

make.go
```diff
@@ -55,15 +53,46 @@ func main() {
 
 type step func() error
 
+func executeDeps() error {
+	bindata := run(
+		"go",
```
bradrydzewski: we could probably use a slice, similar to the css files:

```
deps := []string{"github.com/jteeuwen/go-bindata/...", "golang.org/x/tools/cmd/cover" }
for _, dep := range deps {
   err := run("go", "get", "-u", dep)
   if err != nil {
    return err
  }
}
```

make.go
```diff
@@ -145,104 +174,148 @@ func styles() error {
...
 	}{
-		{"./bin/drone-agent", "drone/drone-agent"},
-		{"./bin/drone-server", "drone/drone"},
+		{
+			"bin/drone-agent",
+			"drone/drone-agent"},
+		{
+			"bin/drone-server",
+			"drone/drone"},
 	}
```
bradrydzewski: can you run `gofmt` on the code? the wrapping seems a bit inconsistent is all:

```
+		{
+			"bin/drone-server",
-			"drone/drone"},
+			"drone/drone",
+		}
``

make.go
tboerger: Haha, yes... you are right. I will fix that.

make.go
tboerger: Good idea, I will do that.

bradrydzewski:
just some minor comments above, but otherwise great work. Would you also mind updating the build instructions in the README? https://github.com/drone/drone/tree/0.4.0#cloning-building-running

tboerger:
> just some minor comments above, but otherwise great work. Would you also mind updating the build instructions in the README? https://github.com/drone/drone/tree/0.4.0#cloning-building-running

Sure, I will address that in the next commit.

tboerger:
@bradrydzewski I have addressed all the comments. Should be fine now :)

PR #1181, opened 2015-09-02
Added clone mode for gitlab, use token mode as default

Bugagazavr:
- added clone_mode parameter for gitlab
- added strategies to generate `.netrc` file

I have move hash method to separate package, because it used in gitlab package too


bradrydzewski:
looks good

PR #1194, opened 2015-09-09
fix docker tls

vtolstov:
if user provide tls certs we need use it in all cases

### DIFF
--- a/pkg/runner/builtin/runner.go
+++ b/pkg/runner/builtin/runner.go
@@ -44,10 +44,38 @@ type Runner struct {
 	Updater
 }
 
+func newDockerClient() (dockerclient.Client, error) {
+	var tlc *tls.Config
+
+	// create the Docket client TLS config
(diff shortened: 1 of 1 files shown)


pkg/runner/builtin/runner.go
```diff
@@ -240,7 +240,27 @@ func (r *Runner) Run(w *queue.Work) error {
 }
 
 func (r *Runner) Cancel(job *types.Job) error {
-	client, err := dockerclient.NewDockerClient(DockerHost, nil)
+	var tlc *tls.Config
```
Bugagazavr: this code repeated twice, maybe you can move that to one function?

vtolstov: I think about this (3 times repeated =)), may be we need something like newDockerClient() function...

vtolstov:
@Bugagazavr i'm fix pr.

tboerger:
:+1:

bradrydzewski:
thanks!

PR #1196, opened 2015-09-10
Dockerfile: use go 1.4 base image

olalonde:


MichielDeMey:
We could even upgrade to 1.5 at this point. ;)
The GC has improved a lot since.

bradrydzewski:
We've got this fixed in the 0.4 branch:

```
FROM centurylink/ca-certs
EXPOSE 8000
ENTRYPOINT ["/drone_static"]
```

We are using `golang:1.5` in our `.drone.yml` file and statically compiling the binary to include sqlite3. We are then adding the binary directly to the image instead of compiling inside. The `centurylink/ca-certs` image is based on `scratch` and is around 250kb. The resulting Drone image is now under 20mb.

PR #1201, opened 2015-09-15
Fix Marathon deployments

kpacha:
use PUT method instead POST so it creates a new app (if it doesn't already exist) or upgrades it


kpacha:
@bradrydzewski, the build failed because a timeout during the dependency installation... please, ignore the fail or relaunch the build process

taik0:
+1 
Without this patch the marathon plugin cannot be used.

bradrydzewski:
thanks! just re-ran but it looks like it just might need to be go-formatted in order to pass the build

kpacha:
Fixed!

kpacha:
There is a bug in the order of the parameters...

PR #1203, opened 2015-09-16
Deliver drone UI from subfolder

tboerger:
To be able to use multiple instances of drone within different subfolders I have started this pull request. In my case i want to ship drone through `https://ci.example.com/github` and `https://ci.example.com/gitlab`, to be able to do that we need the option to respond to these paths.

It's currently WIP because we need to find a proper way how to handover the folder to the angular application. As you can see i have added an `ng-root` attribute to the `body`. What do you prefer to access this attribute?

Please comment @bradrydzewski, I really need this feature :)

### DIFF
--- a/cmd/drone-server/drone.go
+++ b/cmd/drone-server/drone.go
@@ -37,6 +37,7 @@ var conf = struct {
 		addr string
 		cert string
 		key  string
+		root string
 	}
 
 	docker struct {
@@ -70,6 +71,7 @@ func main() {
 	flag.StringVar(&conf.server.addr, "server-addr", ":8080", "")
 	flag.StringVar(&conf.server.cert, "server-cert", "", "")
 	flag.StringVar(&conf.server.key, "server-key", "", "")
+	flag.StringVar(&conf.server.root, "server-root", "/", "")
 	flag.StringVar(&conf.remote.driver, "remote-driver", "github", "")
 	flag.StringVar(&conf.remote.config, "remote-config", "https://github.com", "")
 	flag.StringVar(&conf.database.driver, "database-driver", "sqlite3", "")
@@ -208,11 +210,13 @@ func main() {
 
 	r.SetHTMLTemplate(index())
 	r.NoRoute(func(c *gin.Context) {
-		c.HTML(200, "index.html", nil)
+		c.HTML(200, "index.html", gin.H{
+			"root": conf.server.root
+		})
 	})
 
-	http.Handle("/static/", static())
-	http.Handle("/", r)
+	http.Handle(conf.server.root + "static/", static())
+	http.Handle(conf.server.root, r)
 
 	if len(conf.server.cert) == 0 {
 		err = http.ListenAndServe(conf.server.addr, nil)
--- a/cmd/drone-server/drone.go
+++ b/cmd/drone-server/drone.go
@@ -215,7 +215,7 @@ func main() {
 		})
 	})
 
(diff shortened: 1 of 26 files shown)


cmd/drone-server/static/index.html
```diff
@@ -1,18 +1,18 @@
...
+  <link rel="stylesheet" href="{{.root}}static/styles/reset.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/fonts.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/alert.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/blankslate.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/list.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/label.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/range.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/switch.css" />
+  <link rel="stylesheet" href="{{.root}}static/styles/main.css" />
+  <link rel="icon" href="{{.root}}static/favicon.png">
```
Ablu: I am not sure about any side effects but wouldn't `<base href="{{.root}}" />` be a bit better here?

http://www.w3schools.com/tags/att_base_href.asp

Just my 2 cent. Only randomly saw this pullrequest and thought I would comment. I am not involved in the drone development. So please only listen to the ones who are :P

bradrydzewski:
agreed with @Ablu, I believe that is required per the angular docs:

> if you configure $location to use html5Mode ( history.pushState ), you need to specify the base URL for the application with a <base href="">

also, unfortunately, this change is a bit more complex than just serving the static assets. I believe the oauth code needs to change to use the correct callback URL, the build runner needs to use the correct system URL, the repositories creation needs to use the correct hook URL, and maybe other changes that I haven't considered.

tboerger:
I just missed the base ref, I will replace that. Than i can already see if the UI behaves correctly.

cmd/drone-server/drone.go
```diff
@@ -208,11 +210,13 @@ func main() {
...
 	r.NoRoute(func(c *gin.Context) {
-		c.HTML(200, "index.html", nil)
+		c.HTML(200, "index.html", gin.H{
+			"root": conf.server.root,
+		})
 	})
 
-	http.Handle("/static/", static())
-	http.Handle("/", r)
+	http.Handle(conf.server.root + "static/", static())
```
tboerger: @brad do you prefer that or strings join?

bradrydzewski:
I think most of the server-side changes should be done by adding code to `httputil` to get the base. We already use this for `X-FORWARDED-FOR` to get the hostname when sitting behind a proxy. I'd like to use the same mechanism for multiple drone instances. Drone sits behind a reverse-proxy and we use a standard `X-*` header to designate the base

tboerger:
There are still some files missing that needs to get fixed with the linking. I thought about a directive to replace the path, but that works not for all cases:

* [ ] views/builds/index/toolbar.html
* [ ] views/repos/index/toolbar.html
* [ ] views/users/toolbar.html
* [ ] views/profile/toolbar.html
* [ ] views/repos/edit.html
* [ ] services/feed.js

tboerger:
Seem to work fine with github! Tested at https://ci.webhippie.de/github/webhippie/drone-test/1

tboerger:
Replaced by #1238

PR #1210, opened 2015-09-23
Build Docker image with Go 1.5.1

dtan4:
Now Drone 0.4 requires Go 1.5+ to build a binary. I modified Dockerfile to use Go 1.5.1 for building.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.4.2
+FROM golang:1.5.1
 
 ENV DRONE_SERVER_PORT :80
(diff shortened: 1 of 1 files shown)


bradrydzewski:
we're instead using the `ca-certs` base image (200kb image with just certs installed) and staticly compiling Drone and sqlite outside the container, and adding the binary. This creates the smallest possible image (16mb)  :smile: 

```
FROM centurylink/ca-certs
EXPOSE 8000

ENV DATABASE_DRIVER=sqlite3
ENV DATABASE_CONFIG=/var/lib/drone/drone.sqlite

ENTRYPOINT ["/drone_static"]
```

PR #1213, opened 2015-09-24
Update Dockerfile

desimone:
Prefer the shorter, more ubiquitous equivalent of `>/dev/null 2>&1` to  `1> /dev/null 2> /dev/null` .

Forgive my OCD twitch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ADD . /gopath/src/github.com/drone/drone/
 WORKDIR /gopath/src/github.com/drone/drone
 
 RUN apt-get update
-RUN apt-get -y install zip libsqlite3-dev sqlite3 1> /dev/null 2> /dev/null
+RUN apt-get -y install zip libsqlite3-dev sqlite3 >/dev/null 2>&1
 RUN make deps build embed install
 
 EXPOSE 80


Dockerfile
```diff
@@ -9,7 +9,7 @@ ADD . /gopath/src/github.com/drone/drone/
 WORKDIR /gopath/src/github.com/drone/drone
 
 RUN apt-get update
-RUN apt-get -y install zip libsqlite3-dev sqlite3 1> /dev/null 2> /dev/null
+RUN apt-get -y install zip libsqlite3-dev sqlite3 >/dev/null 2>&1
```
dangra: then what about using the even shorter syntax instead `&>/dev/null` 

Source from [bash manpage](http://www.gnu.org/software/bash/manual/bash.html#Redirecting-Standard-Output-and-Standard-Error)

> 3.6.4 Redirecting Standard Output and Standard Error
>
> This construct allows both the standard output (file descriptor 1) and the standard error output (file descriptor 2) to be redirected to the file whose name is the expansion of word.
>
> There are two formats for redirecting standard output and standard error:
>
>     &>word
>
> and
>
>     >&word
>
> Of the two forms, the first is preferred. This is semantically equivalent to
>
>     >>word 2>&1
>

desimone: `&>/dev/null` is not portable, though that's probably irrelevant here.

dangra: I think the real problem is that google/golang image is based on debian and it uses dash (not bash) as primary shell, so `&>/dev/null` won't work.

dangra:
IMHO this is a pointless exercise as real development is happening on [branch 0.4.0](https://github.com/drone/drone/tree/0.4.0/) where [Dockerfile](https://github.com/drone/drone/blob/0.4.0/Dockerfile) changed a lot.

desimone:
agreed.

dangra:
:+1:

PR #1231, opened 2015-10-09
fix(router): correct for get gitlab status

antonikonovalov:
-/repos/gitlab/go/builds/test-drone-4/commits/e3064b865e445718d453afb94838e6ccf3454b24
+/api/gitlab/go/builds/test-drone-4/commits/e3064b865e445718d453afb94838e6ccf3454b24

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -158,6 +158,9 @@ func normalize(h http.Handler) http.Handler {
 
 		parts := strings.Split(r.URL.Path, "/")[1:]
 		switch parts[0] {
+		case "gitlab":
+                        parts = append([]string{"", "api"}, parts...)
+                        r.URL.Path = strings.Join(parts, "/")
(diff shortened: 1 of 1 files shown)


antonikonovalov:
@bradrydzewski by default now request from gitlab solve to /repos/...
it's not correct by current state of code

bradrydzewski:
great catch, we can keep the internal route as `/gitlab` instead of `/api/gitlab` and then adjust to look something like this:

```diff
-case "settings", "api", "login", "logout", "", "authorize", "hook", "static":
+case "settings", "api", "login", "logout", "", "authorize", "hook", "static", "gitlab":
```

and this:

```diff
-gitlab := e.Group("/api/gitlab/:owner/:name")
+gitlab := e.Group("/gitlab/:owner/:name")
```

bradrydzewski:
I made the suggested changes locally. Thanks though for identifying

PR #1235, opened 2015-10-13
sudo issue fixed

ipedrazas:
Depending on the docker image, when running as root the sudo commands fails. This makes that the docker installation fails as well.

This is just a quick fix that avoids the root/sudo problem.


bradrydzewski:
we just launched the next version of Drone which include a plugin model. Plugins are Docker containers that are completely self-contained with everything they need to run. The Docker plugin is found here http://addons.drone.io/docker/

PR #1237, opened 2015-10-15
When cloning pull requests, use specific branch

pheuter:
### Problem

When running CI against a pull request, `master` will be cloned instead of the pull request branch. This leads to problems like when the master branch has submodules on different versions than the pull request branch, and causes unecessary failures when running tests.

### Proposed solution

When cloning during PR build, clone specific branch just like with generic builds.


thyming:
+1

runjoerun:
+100

PR #1238, opened 2015-10-16
Deliver drone UI from subfolder

tboerger:
**This pull requests replaces https://github.com/drone/drone/pull/1203**

To be able to use multiple instances of drone within different subfolders I have started this pull request. In my case I want to ship drone through `https://ci.example.com/github` and `https://ci.example.com/gitlab`, to be able to do that we need the option to respond to these paths.

It's currently WIP because we need to find a proper way how to integrate the sub path into the authorization. Generally it works already pretty fine, I just need to fix the login functionality.


tboerger:
@bradrydzewski this PR should work now. Any suggestion to improve it is welcome :D

tboerger:
Close for now, added an issue to keep it tracked.

PR #1241, opened 2015-10-19
Fix MySQL init script

psytale:
* Fix a few failing statements in script
* Changed max length of owner/repo name to 255 (limit of 100 in github and 255 in gitlab). MySQL can't handle a unique field of more than 767 bytes by default

### DIFF
--- a/shared/database/mysql/1_init.sql
+++ b/shared/database/mysql/1_init.sql
@@ -18,9 +18,9 @@ CREATE TABLE users (
 CREATE TABLE repos (
  repo_id            INTEGER PRIMARY KEY AUTO_INCREMENT
 ,repo_user_id       INTEGER
-,repo_owner         VARCHAR(500)
-,repo_name          VARCHAR(500)
-,repo_full_name     VARCHAR(1000)
+,repo_owner         VARCHAR(255)
+,repo_name          VARCHAR(255)
+,repo_full_name     VARCHAR(511)
 ,repo_avatar        VARCHAR(500)
 ,repo_link          VARCHAR(1000)
 ,repo_clone         VARCHAR(1000)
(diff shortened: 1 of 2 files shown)


psytale:
Think something funky is happening due to "keys" being a reserved word in mysql. Keys are getting inserted to the DB but when a container is run it doesn't have any keys injected. Perhaps a table rename is necessary?

psytale:
Assuming https://github.com/drone/drone/blob/0.4.0/model/key.go#L38 is silently failing

bradrydzewski:
typically quoting the table name in insert / update / delete / select statements would resolve this. The insert / update statements automatically quote table and column names thanks to the library we are using, but not select statements.

I've been testing mostly with sqlite, but prior to release we'll need to add unit tests for mysql and postgres to catch these sort of issues.

But in the mean time, I'd put my money on quoting the table name fixing the issue.

psytale:
Just unsure how quoting the name would play with psql & sqlite. Assume
it'll be fine. I'll play then update PR tomorrow.

bradrydzewski:
I'll get unit tests setup for the database today ... it is just a matter of adding the mysql and postgres configuration to the `.drone.yml` and invoking the unit tests with environment variables. This will help ensure that any changes we make work across database platforms.

I'm going to go ahead and merge this now, but let's definitely get this fixed in a follow-up pull request

bradrydzewski:
quick update -- unit tests are now being run for postgres and mysql https://github.com/drone/drone/commit/7f9ef94f475884a6f73c4978f7d1d836e1f73cae#diff-3216dfff0ed3e301453e6799e8c367e2

I went ahead with some minor updates to handle `keys` in msyql, as you suggested, so the build would pass. I think before releasing we should rename to `repo_keys` to avoid the issue all together. I typically try to group database changes together in a single commit, to avoid constant migrations. So I'm thinking I can make this change end of week to include any other changes to come up.

Issue #860, opened 2015-01-29
Gogs Private Repos


bradrydzewski:
The current Gogs implementation works for public (open source) repositories only. We need a way to upload ssh keys to the repository in order to clone. Ideally we would also have a way to see if the key already exists, and either update or delete and re-upload.

We'll need the Gogs team to expose this functionality.

For reference, this is the code we use for GitHub:
https://github.com/drone/drone/blob/master/plugin/remote/github/helper.go#L233:L255

@compressed thoughts?

bradrydzewski:
For reference, I just logged https://github.com/gogits/gogs/issues/878

andyjeffries:
Looking forward to having this functionality working.  I'm a real newbie in Golang, but if I can be of any help testing it, let me know.

compressed:
Thanks for logging this. Yep, it is needed.

On the https side, I worked on a PR (already merged) so we can now clone private repos as well using the drone token: https://github.com/gogits/gogs/pull/845

We should try to support both ssh and https methods when doing this work for private repos.

andyjeffries:
Don't suppose you have any blog post/markdown file on how to setup drone with gogs do you?

bradrydzewski:
we do have some docs for Gogs, although not as complete as our GitHub docs (for example)
http://readme.drone.io/setup/config/gogs/

csu:
+1

Ant59:
+1

Kruemmelmonster:
:+1:
just asking is this going to fix http/s cloning aswell?
because in my setup i rely on it

Ant59:
Is there an ETA on this? It's a really important missing feature. Most people using self-hosted Git/CI want private repos, else we'd be using GitHub.

Kruemmelmonster:
i am working on a workaround on this issue keep looking into here later

Kruemmelmonster:
i found a temporary fix wich has a serious security problem (not a problem for me because my drone installation is not accessible anywhere outside my network).
You can see the Username and Password in the build log

I just edited the database (in my case sqlite) and changed the field `repo_clone_url` and `repo_git_url` to have my credentials in them like `http://username:password@gitserver/user/repo`

bradrydzewski:
@Ant59 I agree this is important, however, I'm a bit oversubscribed at the moment. I'm looking for someone from the community to step up and implement Gogs private repository support. I've labeled the issue accordingly.

[...]

PR #1244, opened 2015-10-20
Changes to SSH for private repo in gogs

dolanor:
Hi, I needed to access my private gogs repos from my own docker, therefore I just added some code from other providers (gitlab, github). There were some answers in drone/drone#860 that helped me to understand.

It works for me. But I don't know if there are other specific things I should know and code about.

But simply changing the clone url to SSH and manually copying the ssh keys from drone to gogs worked flawlessly.

And… Thanks for drone! Pretty great project!

### DIFF
--- a/plugin/remote/gogs/gogs.go
+++ b/plugin/remote/gogs/gogs.go
@@ -121,6 +121,10 @@ func (r *Gogs) GetRepos(user *model.User) ([]*model.Repo, error) {
 			},
 		}
 
+		if repo.Private {
+			repo.CloneURL = repo.SSHURL
+		}
+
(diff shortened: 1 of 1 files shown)


bradrydzewski:
we also need to register an ssh key with the Gogs server:
https://github.com/drone/drone/blob/master/plugin/remote/gogs/gogs.go#L138:L153

but we actually support private Gogs repositories in 0.4 using `https` clone and a `.netrc` file. Since 0.4 branch will be merged soon we won't be making changes to 0.3 (master). So I definitely recommend taking a look at 0.4. 

see http://readme.drone.io/docs/setup/index.html for setup instructions

PR #1245, opened 2015-10-21
RFC7239 : Standard header for SSL proxy forwarding

dolanor:
Hi,
Having misconfigured my nginx headers, my gogs instance didn't get the right webhook URL. It was registered with the "http://" scheme.
Trying to understand how to fix it, I found http://stackoverflow.com/questions/13111080/what-is-a-full-specification-of-x-forwarded-proto-http-header . Since this RFC dates from 2014, I guess it could be a good thing to comply to it.

Keeping the old header to it could also be done, but I guess it is better to move forward, or else we should even implement all the possibilities in the wild : http://stackoverflow.com/questions/16042647/whats-the-de-facto-standard-for-a-reverse-proxy-to-tell-the-backend-ssl-is-used

What do you think?

### DIFF
--- a/shared/httputil/httputil.go
+++ b/shared/httputil/httputil.go
@@ -5,6 +5,15 @@ import (
 	"strings"
 )
 
+func hasHttpsForwarded(r *http.Request) bool {
+	forwardedHeader := r.Header["Forwarded"]
+	for _, w := range forwardedHeader {
+		strings.Contains(w, "proto=https")
+		return true
+	}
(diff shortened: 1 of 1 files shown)


bradrydzewski:
Very cool, I wasn't aware of RFC7239. FYI we'll need to backport these changes to the `0.4` branch. The `0.4` branch introduces major changes (plugins, matrix, etc) and the branches have diverged too much to merge.

PR #1248, opened 2015-10-22
Added missing dependencies ( gin, cache, sql-migrate , dockerclient ,…

rahal:
There is some dependencies missing in the makefile.


bradrydzewski:
this needs to be run with `export GO15VENDOREXPERIMENT=1` to take advantage of Go vendoring. You can also look at the `.drone.yml` to see how everything works.

rahal:
I see. Thank you.

PR #1252, opened 2015-10-23
RFC7239 : Standard header for SSL proxy forwarding

dolanor:
Done modifications to include RFC7239 in 0.4.0.
Compared to my 0.3.0 branch, I added test and a more generic way of parsing the "Forwarded" header and getting the values.
I also deleted the X-Real-IP code for getting the host, since the RFC7239 should handle it with the first `for=` options of the list. And it also has the `host=` options that could be used directly.
But for me, X-Host and X-Real-IP are different. One is for getting the original http request from the client before all the redirection. And X-Real-IP is the IP from the original client. Those are 2 totally opposite ends.
But all of that is based on what I learned on the web. And seriously, there are not so many documentation on all those non standard extensions except X-Forwarded-For and X-Forwarded-Proto. So maybe my assertions are wrong.


bradrydzewski:
I'd like to check for the existence of `Forwarded` before parsing since it is something that we'll be doing on every request, and attaching to the context. I'm in the process of moving from `httputil` to a `location` middleware.

So maybe something like this instead:

```diff
func GetHost(r *http.Request) string {
	switch {
	case len(r.Host) != 0:
		return r.Host
	case len(r.URL.Host) != 0:
		return r.URL.Host
+	case len(r.Header.Get("Forwarded")) != 0:
+		return parseForwardedHeader(r)
	case len(r.Header.Get("X-Forwarded-For")) != 0:
		return r.Header.Get("X-Forwarded-For")
	case len(r.Header.Get("X-Host")) != 0:
		return r.Header.Get("X-Host")
	case len(r.Header.Get("XFF")) != 0:
		return r.Header.Get("XFF")
-	case len(r.Header.Get("X-Real-IP")) != 0:
-		return r.Header.Get("X-Real-IP")
	default:
		return "localhost:8000"
	}
}
```

Initially I liked the way you parsed into a struct, but upon more thought I'm a bit torn. This will generate extra garbage for the gc on every page request to aggregate data we may not need. Just returning the string we ask for may be better.

```diff
+type forwardedHeader struct {
+	For   []string
+	Proto string
+	By    []string
+	Host  string
+}
```

Also I'm wondering if we need the `strings.Split` since the `Header` is of type `type Header map[string][]string`. It would be interesting to verify if Go splits the header automatically when it parses the http request and creates the header object.

If it doesn't, maybe it is better to use `r.Header.Get("Forwarded")` instead of `r.Header["Forwarded"]` which returns a string instead of an array, to avoid the extra iteration.

Thanks for looking into all of this, really appreciate it!

dolanor:
Actually, Go will create a `[]string` only if there are multiple header of the same type

```
Forwarded: for=1.2.3.4; for=12.34.56.78
Forwarded: for=4.3.2.1; proto=https
```
It will create a 
```
map[string][]string{
Forwarded: []string{
"for=1.2.3.4; for=12.34.56.78",
"for=4.3.2.1; proto=https"
```
So, not really helpful, but still could be an issue because the norm doesn't block this.

I spent some time to find a library that would parse that kind of data, but couldn't really find one that suited my needs. The closer I could get from was : https://github.com/golang/gddo/blob/9f5f5a58539570ac98c1e905a015569ea1bf4963/httputil/header/header.go#L135-L163
But still didn't do what I wanted. So I went in for a quick'n'dirty.

But you raise a good question with the GC. I will try to modify it in the next days though. I'm hitting bed now :)
Though I might create a mini library out of that because it's weird nothing exist already…

dolanor:
I'm not sure if you've seen, but I updated the branch to take your changes into account.
Tell me if it's good for you.

shared/httputil/httputil.go
```diff
@@ -5,6 +5,30 @@ import (
...
+				val = append(val, value)
+			}
+		}
+	}
+	return
+}
+
+func hasHttpsForwarded(r *http.Request) bool {
+	if parseForwardedHeader(r, "proto")[0] == "https" {
+		return true
```
bradrydzewski: could this lead to an index out of bounds exception if `parseFormwardedHeader` returns an empty array due to lack of `["Forwarded"]` header?

dolanor: In fact. I corrected that, added some tests, rebased and pushed again!

Issue #1274, opened 2015-10-28
Migration failed: table users already exists


wolfgang42:
I have a Docker container running Drone, which I just tried to migrate by running
```bash
docker stop droneci
docker rm droneci
docker pull drone/drone:0.4
/usr/bin/docker run --name droneci \
	# Some arguments omitted for clarity
	-e 'DRONE_DATABASE_DATASOURCE=/var/lib/drone/drone.sqlite' \
	-v '/srv/drone:/var/lib/drone' \
	drone/drone:0.4
```

However, the new container now outputs the following to stdout and quits:
```
time="2015-10-28T21:05:25Z" level=info msg="using database driver sqlite3" 
time="2015-10-28T21:05:25Z" level=info msg="using database config /var/lib/drone/drone.sqlite" 
time="2015-10-28T21:05:25Z" level=error msg="table users already exists" 
time="2015-10-28T21:05:25Z" level=fatal msg="migration failed" 
```

I found #224, which seems similar but has been fixed already.

The version in the `migration_version` table is 3.
The `users` table does indeed exist, and has an entry for me in it (I was the only user).

bradrydzewski:
We recently upgrade from Drone 0.3-alpha to 0.4-beta. This was a breaking release and is not backward compatible with 0.3. The database and yaml have changed significantly and need to be updated to work with the latest version.

To use the 0.4 release you'll need to start with a new database and update all your yaml files. The move from alpha to beta signals stability moving forward with future releases.

bradrydzewski:
for reference, these are the new instructions to run Drone. Please note that the environment variables have changed as well, so make sure you check out the new docs:

http://readme.drone.io/setup/
http://readme.drone.io/build/

If you have any questions or issues with the upgrade please reach out in our gitter room, https://gitter.im/drone/drone

PR #1276, opened 2015-10-29
Add link from installation to upgrade instructions

wolfgang42:
This is a followup to #1274, where I didn't notice that 0.4 was a breaking release.
The upgrading documentation mentions this, but I was looking on the installation instructions page and missed the link in the sidebar.
Hopefully this change will prevent someone else from making the same mistake in the future.


bradrydzewski:
we moved the docs to a separate repo at https://github.com/drone/docs but feel free to send any improvements or clarifications. Thanks!

PR #1285, opened 2015-10-30
PowerShell SHA256 command now emits lowercase

### LINKED ISSUE #1277: .drone.yml checksum must be lowercase, but PowerShell emits uppercase
The documentation on [secret variables](http://readme.drone.io/build/secrets.html) suggests the following PowerShell command to get the checksum for the file:
```powershell
$ Get-FileHash .\.drone.yml -Algorithm SHA256
```

The problem is that `Get-FileHash` emits an uppercase hash, but Drone expects a lowercase one, so validating the checksum fails.
You can get a lowercase hash from PowerShell like so:
```powershell
$ (Get-FileHash .\.drone.yml -Algorithm SHA256).Hash.ToLower()
```

On the other hand, perhaps it would be better to accept both uppercase and lowercase checksums? I can submit a pull request for either solution, but I'm not sure which one is better.

wolfgang42:
Previously it emitted uppercase hashes, which drone-exec didn't like.

Closes #1277

### DIFF
--- a/docs/build/secrets.md
+++ b/docs/build/secrets.md
@@ -61,5 +61,5 @@ f63561783e550ccd21663d13eaf6a4d252d84147  .drone.yml
 Generate a checksum on Windows with powershell:
 
 ```
-$ Get-FileHash .\.drone.yml -Algorithm SHA256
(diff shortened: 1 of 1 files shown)


bradrydzewski:
thanks so much, really appreciate it !

PR #1287, opened 2015-10-30
correct cc menu link

macb:
@joonas noticed this when we were messing around with an internal drone deployment


bradrydzewski:
great catch, thanks so much.  I was just talking about cc menu in gitter ... good timing!

PR #1300, opened 2015-11-06
Display full name of repository in list

wolfgang42:
Consider the following:

Drone CI dashboard, with a search for "small-label-frontend" entered. Two different repositories have come up, both with the name "Small-Label-Frontend" and no way to tell the two apart.

One of these is the main repository; the other one is my personal fork. There's no way of telling which is which without hovering over them and looking at the URL.

Drone 0.3 did not have this problem, as it showed the full repository name, with group. This patch changes the dashboard in 0.4 to do the same thing.

### DIFF
--- a/template/amber/repos.amber
+++ b/template/amber/repos.amber
@@ -32,8 +32,7 @@ block content
                                     else
                                         img.avatar[src="/static/images/dummy.png"]
                                 div.card-block
-                                    h3.login #{$repo.Name}
-                                    div.full_name.hidden #{$repo.FullName}
+                                    h3.full_name #{$repo.FullName}
 
 block append scripts
     if len(Repos) != 0


gtaylor:
Yikes, I never thought of this. It would definitely break teams that do a fork + pull request workflow.

template/amber/repos.amber
```diff
@@ -22,8 +22,7 @@ block content
                             div.card-header
                                 img.avatar[src=$repo.Avatar]
                             div.card-block
-                                h3.login #{$repo.Name}
-                                div.full_name.hidden #{$repo.FullName}
+                                h3.full_name #{$repo.FullName}
```
bradrydzewski: does changing the css class to `full_name` alter the style? Not sure if we have a class for `full_name` in which case we should leave the class as `login`

wolfgang42: No, changing it from `.login` to `.full_name` does not alter the style. None of the styles for `.login` apply here (I'm not sure why it has that attribute set at all) and `.full_name` does not have any styling at all.

However the search box depends on  `#{$repo.FullName}` having a class of `full_name`.

bradrydzewski:
I'm wondering if the fix for this is to resolve the lack of avatar? What remote are you using that no avatar exits?  If you see the example below we use avatar to differentiate ownership:

screen shot 2016-04-07 at 1 42 21 pm

wolfgang42:
I'm using GitLab. I think even if it had identicons I'd prefer the name to be included as well. The search bar searches the full name, and it's confusing to be searching for something that doesn't show up. (Example: Imagine I have a fork of the [ski](https://www.npmjs.com/package/ski) package. Why do all of your repositories show up when I try to find it?)
Here's what it looks like with this patch:

drone-1300-patch

gdubicki:
+1

gdubicki:
Gitlab doesn't show avatars by default, @bradrydzewski :

screen shot 2016-06-07 at 12 33 49

bradrydzewski:
this is resolved in the new [drone-ui](https://github.com/drone/drone-ui) repository. See screenshot

screen shot 2016-07-08 at 3 46 04 pm

wolfgang42:
Looks nice! I assume this is only in the currently experimental 0.5 release?

tboerger:
> Looks nice! I assume this is only in the currently experimental 0.5 release?

That's correct, and since 0.4 got a freeze it won't get backported

PR #1304, opened 2015-11-06
Adding more documentation to the swagger

donny-dont:


bradrydzewski:
Thanks. I'm offline for a bit but will jump back on gitter a bit later

PR #1306, opened 2015-11-09
Add /etc/nsswitch.conf to docker container

### LINKED ISSUE #1269: drone 0.4 docker image can't use /etc/hosts
There is no `/etc/nsswitch.conf` in the image so I extended it and added one and it started working:

```
# /etc/nsswitch.conf
#
# Example configuration of GNU Name Service Switch functionality.
# If you have the `glibc-doc-reference' and `info' packages installed, try:
# `info libc "Name Service Switch"' for information about this file.

passwd:         compat
group:          compat
shadow:         compat

hosts:          files dns
networks:       files

protocols:      db files
services:       db files
ethers:         db files
rpc:            db files

netgroup:       nis
```

thomasf:
Fixes #1269 

IDK where you want this file to exist in the tree... I'll move it to where you would like it to be.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@
 
 FROM centurylink/ca-certs
 EXPOSE 8000
+ADD contrib/docker/etc/nsswitch.conf /etc/
 
 ENV DATABASE_DRIVER=sqlite3
 ENV DATABASE_CONFIG=/var/lib/drone/drone.sqlite
--- a/contrib/docker/etc/nsswitch.conf
+++ b/contrib/docker/etc/nsswitch.conf
@@ -1,0 +1,19 @@
+# /etc/nsswitch.conf
+#
+# Example configuration of GNU Name Service Switch functionality.
+# If you have the `glibc-doc-reference' and `info' packages installed, try:
+# `info libc "Name Service Switch"' for information about this file.
+
+passwd:         compat
+group:          compat
+shadow:         compat
+
+hosts:          files dns
+networks:       files
+
+protocols:      files
+services:       files
+ethers:         files
+rpc:            files
+
+netgroup:       nis


bradrydzewski:
sounds good, let's put in `contrib/docker/etc/nsswitch.conf` to mirror what we've got going with our debian packaging :smile:

thomasf:
done

bradrydzewski:
hey just a minor follow-up question. I noticed the below four variables had both `db` and `files` on a few different machines and versions of ubuntu, as well as in the Go unit tests [1]

```
protocols:      db files
services:       db files
ethers:         db files
rpc:            db files
```

Any idea what the difference is between `files` and `db files`? I'm just unfamiliar with this file and just wanted to double-check before merging. Thanks :smile: 

[1] https://github.com/golang/go/blob/master/src/net/nss_test.go#L28:L31

thomasf:
I read the man pages for nsswitch.conf before creating this pull reques and I removed the db from the template from a clean ubuntu install I referenced in the original issue. 

Since drone is built with netgo/cgo off I guess that it shouldn't matter much but from what I could understand but I removed the "db" stuff anyways.. 

From the man page:

```
The service specifications supported on your system depend on the presence of shared libraries, and are there-
       fore  extensible.   Libraries  called  /lib/libnss_SERVICE.so.X will provide the named SERVICE.  On a standard
       installation, you can use "files", "db", "nis", and "nisplus". 
```

thomasf:
I tested my setup with and without the "db" part.. Having said this,  I'm not entirely sure which the "db" provider of my normal OS actually might be.

thomasf:
Any specific reason why you are holding off merging this?

bradrydzewski:
I have been testing this and there seem to be no adverse impacts so we can merge. I also found the following documented workaround, which is a one-liner that can be included directly in the Dockerfile. Does this fix the issue? Or do we need the full file?

```
RUN echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf
```

PR #1307, opened 2015-11-10
Enable vendoring for github.com/go-swagger dependency

hkjn:
Building in a Docker container based on `golang` doesn't work at `427560a`: C.f https://github.com/drone/drone/commit/2d37d66e2d5ef0de221339a7d010fabf94b12d79


bradrydzewski:
thanks!

PR #1309, opened 2015-11-11
Publish section example for tag pushes

gtaylor:
Adds an example for execution limiting publishes to situations where tags are pushed.


bradrydzewski:
thanks!

PR #1315, opened 2015-11-11
sql: remove IF NOT EXISTS from mysql migrations

macb:
`CREATE INDEX IF NOT EXISTS` is not supported by mysql

```
time="2015-11-11T22:23:06Z" level=error msg="Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IF NOT EXISTS ix_build_repo ON builds (build_repo_id)' at line 1"
```


macb:
introduced in #1311

PR #1318, opened 2015-11-12
Add missing injected variables to doc

x-cray:
According to https://github.com/drone/drone-exec/blob/master/main.go we have also $$COMMIT_SHORT and $$BUILD_NUMBER injected. Not sure about $$REPO though, couldn't find where it's explicitly injected.


bradrydzewski:
hey there, `COMMIT_SHORT` is being deprecated in favor of `${COMMIT:8}. You are correct that REPO is not being injected either. Good catch

x-cray:
Fixed that.

bradrydzewski:
thanks!

PR #1334, opened 2015-11-20
Big round of improvements to the publish step docs.

gtaylor:
I struggled a bit to understand the publish step initially, so this PR includes everything that I felt would have helped me grok it more quickly. It is verbose, but I feel like this is probably one of the most important things for our users to have the best grasp of.

This was written right before bedtime, so the wording could probably be tightened up. However, this should be a strong improvement on what was there before. Hopefully it helps someone else!

### DIFF
--- a/docs/build/publish.md
+++ b/docs/build/publish.md
@@ -1,19 +1,110 @@
 # Publish
 
-Drone uses the `publish` section of the `.drone.yml` to configure publish steps. Drone does not have any built-in publish or artifact capabilities. This functionality is outsourced to [plugins](http://addons.drone.io). See the [plugin marketplace](http://addons.drone.io) for a list of official plugins.
+Drone uses the `publish` section in `.drone.yml` to configure publish steps.
+Publish steps are primarily used to do two things:
 
-An example configuration that builds a Docker image and publishes to the registry:
+1. Build a Docker image from a `Dockerfile`.
+2. Publish said Docker image to an image repository.
+
+## Publishing and plugins
+
+Drone does not have any built-in publish or artifact capabilities. 
+This functionality is outsourced to [plugins](http://addons.drone.io).
+Consequently, the behaviors of the various plugins can vary.
+
+See the [plugin marketplace](http://addons.drone.io) for a list of official 
+plugins and their documentation.
+
+## How the publish step interacts with the clone/build steps
+
+At the start of each build, Drone clones your repository in the
+[clone](clone.html) step.
+
+Once we have an up-to-date clone of your repository, the [build](build.html)
+step starts. A build container is fired up (based on the image you specify
+in the build step) and the cloned source is mounted as a volume. The end
+result is a subdirectory beneath `/drone` in your build container. From here, 
+you can run tests, compile binaries, pack single page apps, or do any other 
+prep work before creating a Docker image in the next step.
+
+Once the build step concludes, we move on to the publish step. While behavior
+can vary based on the plugin you choose, this is roughly equivalent to running
+`docker build` and `docker push` from within your build sub-directory within
+`/drone`. Your Dockerfile will typically COPY/ADD files from your build
+directory.
+
+The publish step is for building and pushing Docker images, optionally
+pulling in files from the [clone](clone.html) and [build](build.html) steps.
+
+## Basic examples
+
+The following example configuration uses the 
+[Docker publish plugin](http://addons.drone.io/docker/) to build and push a 
+Docker image to an image registry (Docker Hub by default):
 
 ```yaml
 publish:
   docker:
     username: kevinbacon
-    password: pa55word
+    password: someday
     email: kevin.bacon@mail.com
     repo: foo/bar
     tag: latest
     file: Dockerfile
 ```
+
+With the inclusion of the `registry` key, we can publish to other image 
+registries:
+
+```yaml
+publish:
+  docker:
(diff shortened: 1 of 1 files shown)


bradrydzewski:
@gtaylor I definitely think based on the amount of questions I get daily in the Gitter room, I've definitely missed the mark on the publish and deployment documentation, especially related to all things Docker. So thanks for taking the time to work on this.

**warning** This comment ended up turning into a rambling mess of ideas (sorry about that) but I think this is a good opportunity to address the overall structure of our documentation and issues related to finding relevant content. I will comment some on the publish documentation, however, I want to solve for some of the overarching issues we're seeing with the documentation as well.

---

> Publish steps are primarily used to do two things:
> 1. Build a Docker image from a `Dockerfile`.
> 2. Publish said Docker image to an image repository.

There are a bunch of other publish plugins, including `bintray`, `pypi`, `npm`, `s3`, `cloudfiles` and there will be many more added over the coming months. So I do think, as a general introduction to publish plugins, it may focus too heavily on the Docker use case.

There is also a bit of overlap between this documentation and the documentation at http://addons.drone.io/docker/. One of the big wins having the plugins documented alongside the code (each plugin has a `DOCS.md` file in its repository) is that pull requests to add or modify functionality almost always include updated docs. This was a big problem with the old version of Drone.

In fact, we see this exact issue with the git clone documentation. The plugin documentation (at http://addons.drone.io/git/) has stayed up-to-date, however, the clone documentation has fallen out-of-date (see http://readme.drone.io/build/clone.html).

Overall, here are some of the challenges I'm wondering how we can solve:

* Can we document the overall plugin process (like you've done) but with minimal overlap in plugin-specific documentation?
* Can we make plugin documents more discoverable?
* Can we make plugin documents feel less disjointed from the main documentation?

Not really related to this pull request, but I've also started creating simple, sample applications at https://github.com/drone-demos

The goal is to have a bunch of sample repositories that highlight how to use Drone with specific types of technologies. Some examples include testing Drone with a project that needs Postgres and testing Drone with a project that needs Selenium. Perhaps we could include a sample for building and publishing a Dockerfile with Drone.

Similar to the plugins, I'd like to include a `DOCS.md` file in each sample repository that heavily documents how everything works, similar to a blog post or tutorial. I can then auto-generate an entire site (similar to the plugin marketplace) that focuses on "How do I use Drone with ...". Like plugins, we'll need to make sure this is discoverable and doesn't feel disjointed from the main documentation.

Sorry for rambling on here, but I wanted to make sure I did a brain dump of everything I've been thinking related to documentation. Hopefully with the full picture, and if you're up for it, we can brainstorm the best way to go about make this all fit together.

gtaylor:
The main documentation vs plugin documentation is indeed a bit tough to solve. When I was first getting oriented, this did cause some confusion as plugins can be used to do *everything*.

I suspect this is going to end up being something we're going to have to tinker with, fail at, and repeat until we find the right balance. I took a first stab at it last night without applying the same philosophy to the rest of the doc pages, so this page looks a bit out of place right now. A few of the premises for what you see in this PR are:

* Context switching between pages in the documentation carries with it some overhead and distraction. At the same time, we can't cram everything on to each page. However, we can try to provide enough of a feel for how what we are looking at fits into the greater picture. I may have gone too detailed with this initial revision.
* Any examples in the main docs should be solely for the purpose of teaching concepts. They shouldn't aim to be a complete reference. Mine were definitely a bit docker-heavy, but I couldn't find the bintray plugin that was previously referenced. The user should understand the basic concepts in the main docs before getting too caught up looking at per-plugin docs. For example "Publish plugins can have different parameters", "Publish plugins interact with different services", and "Publish plugins are generally centered around image building/publishing". They really need to know these things before the per-plugin docs are going to reach maximum value.
* The main documentation should be light, topical, and conversational. By comparison, the reference/plugin documentation would spend more time documenting parameters, behaviors, interactions, showing or linking to examples, etc. Completing the circle, we should mention and link to some of these plugins in the main docs to help illustrate variety and show off different options.
* The main documentation should "sell" Docker by specifically showing off some of the novel things you can do with it. This may come at the expense of a little bit of example repetition between main docs and plugin docs. Once potential users land on the Doc page, we have a very short window of time to convince them that Drone is interesting and worth further exploration.

The outshoot of all of this was that we ended up with stuff like the "How the publish step interacts with the clone/build steps", which does duplicate some info that would eventually be in the overview section. The thought was to very briefly allude to the other steps, but not go into as much detail as the eventual, more thorough build process overview documentation would. A brief re-cap to save them diving into the overview again. 

As you had noted, there are also a good number of examples listed. Perhaps we could narrow this down somewhat and move the rest to the Docker plugin doc. The content on this particular page should, IMO, be centered around bringing that "ah-ha!" moment as efficiently as possible for the widest audience. The examples that were chosen were also aimed at trying to sell Docker as flexible, but maybe this was too much.

To sum it up, I think we should be OK with *some* repetition as long as it simplifies the learning/selling process. The "main" documentation should be enough to teach concepts and help people reach that point of understanding. The plugin docs will definitely go into far greater detail, and will have more reference-oriented material (along with relevant links to examples).

As far as the drone-demos repo, I think that's great. That would be a huge asset to everyone. I'd love to still see us include a handful of examples in the docs to help teach the concepts without context switching, but we could go into much greater detail and variety in drone-demos.

Apologies for the counter-ramble :)

gtaylor:
Superceded by drone/docs.

PR #1338, opened 2015-11-20
ability to replace repository private key

bradrydzewski:
this is the endpoint only -- `curl -X POST -d @id_rsa http://localhost:8000/api/repo/octocat/hello-world/key`


khirakawa:
Awesome :+1:

Issue #1345, opened 2015-11-24
Bitbucket repos with uppercase characters fail to activate


gtaylor:
I've got some mixedCase repositories from other groups on our Bitbucket org. We're trying to activate them to get some builds going on Drone. We're seeing something like this in the Drone stdout when trying to activate:
```
[500] Repository myorg/myRepo not found
```
It appears to be related to using Repo.Name instead of Repo.FullName. The former is just the repo name "myRepo", while confusingly the FullName is the full "myorg/myrepo". A few important things to note:

* FullName appears to always be lowercase. Whereas Name shows the same caps that are visible in the Bitbucket UI.
* It looks like the current remotes all assume a separate notion of an owner + repo name as separate args. Bitbucket assumes that you are always passing a FullName around.

We talked about a few potential resolutions, but there may be more:
* Make all remote code not assume the owner+repo args. Probably would mean passing around some combo string.
* Add some conditionals above the bitbucket remote that splits FullName on forward slash, passing the result in as ``owner`` and ``name``.
* Leave everything as-is, but lowercase Name for now. Fixes the vast majority of cases, but may not cover special chars.

I'm not entirely sure where to do any of these, but it sounded like we were leaning towards the lowercase solution as an acceptable stopgap. Where in the code might this go? I figured we'd probably want to avoid repeatedly calling lowercase, but wasn't sure how else to go about fixing this without doing so.

bradrydzewski:
@gtaylor I propose we should use the `full_name` to derive the `owner` and `name` fields. This is the easiest, least disruptive path forward to solve the issue.

first step to update https://github.com/drone/drone/blob/master/remote/bitbucket/helper.go#L12

```diff
	repo := model.Repo{
-		Owner:     from.Owner.Login,
-		Name:      from.Name,
+		Owner:     strings.Split(from.FullName, "/")[0],
+		Name:      strings.Split(from.FullName, "/")[1],
		FullName:  from.FullName,
		Link:      from.Links.Html.Href,
		IsPrivate: from.IsPrivate,
		Avatar:    from.Owner.Links.Avatar.Href,
		Kind:      from.Scm,
		Branch:    "master",
	}
```

and then update https://github.com/drone/drone/blob/master/remote/bitbucket/helper.go#L94

```diff
	return &model.RepoLite{
-		Owner:    from.Owner.Login,
-		Name:     from.Name,
+		Owner:     strings.Split(from.FullName, "/")[0],
[...]

PR #1351, opened 2015-11-26
Temporary bandaid to fix Bitbucket case sensitivity.

gtaylor:
As per #1345, Bitbucket repos that have uppercase letters are currently failing. This PR splits on FullName, which is lowercase. This will avoid the issues inherent with concatenating owner + Name (which can have mixed case which causes the API errors).

**Limitations:**
* All bitbucket repositories will now show up as lowercase in Drone.


bradrydzewski:
thanks :)

PR #1359, opened 2015-11-30
Restores the .deb structure in downloads.drone.io

asosso:
The current `drone.deb` address is `downloads.drone.io/drone/0.4` (master branch).
It's very different from other version, for example:

* Branch 0.4.0: `downloads.drone.io/0.4.0/drone.deb`
* Branch 0.3.1: `downloads.drone.io/master/drone.deb`

I suggest to restore `target: $$BRANCH/` or update it to `target: 0.4.0/`

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,7 +29,7 @@ publish:
     access_key: $$AWS_KEY
     secret_key: $$AWS_SECRET
     source: contrib/debian/drone.deb
-    target: drone/0.4
+    target: $$BRANCH/
     when:
       repo: drone/drone
 


bradrydzewski:
@asosso the reasons this hasn't been updated (or documented) is that we're deprecating the `.deb` distribution and pushing everyone toward a Docker-based installation.

asosso:
Ok, I didn't realize that.
Perhaps it would be better to remove this section from .drone.yml?

floatdrop:
@bradrydzewski we need to update 0.3.0 version to 0.3.1, but there is now way to do that. Where 0.3.1 deb package can be downloaded now? `downloads.drone.io/0.3.1/drone.deb` returns `403` response.

tboerger:
@floatdrop 0.3 have reached EOL so it's unsupported and you should reinstall 0.4

bradrydzewski:
@floatdrop 0.3.1 would need to be compiled from source and is tagged as https://github.com/drone/drone/tree/v0.3.1

floatdrop:
@bradrydzewski yep, I try'd, but failed. Downloading from `downloads.drone.io/master/drone.deb` should do the trick - now drone works with recent docker.

bradrydzewski:
removed the debian section from the yaml per suggested, closing issue

PR #1365, opened 2015-12-07
making remote driver error message more verbose

sokoow:


bradrydzewski:
closing since fixed in0.5

PR #1369, opened 2015-12-09
Fixed REMOTE_CONFIG

zumbrunnen:
Variable interpolation does not seem to work (Drone 0.4). Using the old sample code, Drone just sent `...?client_id=%24%7BCLIENT%7D&...` to GitHub, which resulted in a 404 error.

Further, quotes in variables lead to problems - e.g. with plug-in whitelisting. So, I've removed them.

I think, many people copy the sample `dronerc` and will run into this. So it would be good to start with a working config file.


zumbrunnen:
I assume this isn't of any interest, so I'm closing my PR. KTHXBYE :smile_cat:

PR #1370, opened 2015-12-11
skip build with any combination of "skip" and "ci"

glaszig:
a commit message containing any case-insensitive variant of the above two words wrapped in square brackets will skip the build.
examples: `[ci skip]`, `[skip CI]`

this reintroduces pr #1134 and additionally logs the matched skip instruction.

i have the code, you have the tests since i don't have sufficient testing infrastructure in place.
let me know if anything's stupid.

### DIFF
--- a/controller/hook.go
+++ b/controller/hook.go
@@ -43,9 +43,9 @@ func PostHook(c *gin.Context) {
 
 	// skip the build if any case-insensitive combination of the words "skip" and "ci"
 	// wrapped in square brackets appear in the commit message
-	skipMatches := skipRe.FindStringSubmatch(build.Message)
-	if len(skipMatches) > 0 {
-		log.Infof("ignoring hook. %s found in %s", skipMatches[0], build.Commit)
+	skipMatch := skipRe.FindString(build.Message)
+	if len(skipMatch) > 0 {
+		log.Infof("ignoring hook. %s found in %s", skipMatch, build.Commit)
 		c.Writer.WriteHeader(204)
 		return
 	}
--- a/controller/hook.go
+++ b/controller/hook.go
@@ -20,6 +20,8 @@ import (
 	"github.com/drone/drone/yaml/matrix"
 )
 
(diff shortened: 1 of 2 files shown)


controller/hook.go
```diff
@@ -38,10 +39,12 @@ func PostHook(c *gin.Context) {
 		return
 	}
 
-	// a build may be skipped if the text [CI SKIP]
-	// is found inside the commit message
-	if strings.Contains(build.Message, "[CI SKIP]") {
-		log.Infof("ignoring hook. [ci skip] found for %s")
+	// skip the build if any case-insensitive combination of the words "skip" and "ci"
+	// wrapped in square brackets appear in the commit message
+	skipRe := regexp.MustCompile(`\[(?i:ci *skip|skip *ci)\]`)
```
bradrydzewski: we can move this to a global variable to only compile on init

controller/hook.go
```diff
@@ -38,10 +39,12 @@ func PostHook(c *gin.Context) {
...
+	skipMatches := skipRe.FindStringSubmatch(build.Message)
+	if len(skipMatches) > 0 {
+		log.Infof("ignoring hook. %s found in %s", skipMatches[0], build.Commit)
 		c.Writer.WriteHeader(204)
```
bradrydzewski: if we always take the first match (ie `match[0]`) we can use `skipMatch := skipRe.FindString(build.Message)` ... not that the existing implementation is wrong ... just a short-hand :smile:

bradrydzewski:
thanks! added a minor comment, but otherwise looks good

glaszig:
i'm still pretty new the golang and its stdlib so, thanks for the input. you're right, of course. fixed everything. let me know if it's ok, i'd like to squash the commits before you merge.

bradrydzewski:
@glaszig thanks!  the build is failing due to a recent change with one of the utilities we `go get` to generate our api documentation. I've opened an issue with the project https://github.com/go-swagger/go-swagger/issues/166

PR #1374, opened 2015-12-12
Move gitlab client inside drone

Bugagazavr:


Bugagazavr:
@bradrydzewski looks like something going wrong with go-swagger

martinsefcik:
Don't you want to use this gitlab client which has 100% coverage of Gitlab API ?
https://github.com/xanzy/go-gitlab

Bugagazavr:
@martinsefcik no, because we not need all 100% gitlab API, but embedded API allow contribute to drone without intermediate PR's into client library.

martinsefcik:
If the client has 100% API coverage, you probably don't need to do PR :)
Also Drone still uses non-embended go-github, go-gogs clients and I think there is no problem with that.

Bugagazavr:
@martinsefcik this problem, because i maintain current client library ( go-gitlab-client ). For example, we need integrate some bleeding edge gitlab feature/fix, gitlab with this feature/fix not released, we need wait 22 day of the month, before we send our feature/fix to https://github.com/xanzy/go-gitlab, and then we need wait when we merged

Bugagazavr:
And for example thats client not covers 100% gitlab API, https://godoc.org/github.com/xanzy/go-gitlab no API for commit statuses

martinsefcik:
Yes, you are right, because status API is quite new feature. 
But I still think it's not good programming approach to include dependencies directly to drone. Then all libraries/dependencies used by drone can be included to drone because they require some PR when something is not working in that libraries ... I don't think it as good idea.
But it's not up to me to decide :)

Anyway I will submit PR to add status API support into xanzy/go-gitlab client.

Bugagazavr:
@bradrydzewski ping

bradrydzewski:
@Bugagazavr I'm good w/ moving it inside Drone similar to what we've done with Bitbucket. If we do this, I'd like to remove all un-used code from the GitLab library, so that it is only functions used by Drone itself. This will make it easier for us to maintain.

@martinsefcik I'm also good with switching to xanzy/go-gitlab if the library has all the features required by Drone. In the short term, we should merge this PR because @Bugagazavr is no longer supporting our GitLab integration and he may not be able to maintain his library going forward. Once this PR is merged we can start looking at xanzy/go-gitlab

Bugagazavr:
@bradrydzewski done, i'l remove many useless data from structs

Issue #1324, opened 2015-11-16
RFC Make Build Runner Configurable / Pluggable


bradrydzewski:
We should be able to make the build runner (drone-exec) configurable / extendable / swappable. There are plenty of reason why teams may want to completely replace (or wrap) drone-exec.

Some use cases that come to mind:

1. need custom logic to distribute a single build across multiple servers (like https://stripe.com/blog/distributed-ruby-testing)
2. need a custom yaml format or decryption process. The build runner (drone-exec) is responsible for parsing the yaml, so by extending drone-exec a team could modify, extend, or replace the yaml file
3. need single-use build servers. For example, you could wrap drone-exec with a custom build runner that creates a new server on google compute, runs the real drone-exec on that server, and then destroys the server

The first step is to replace these lines of code with something configurable via environment variable:
https://github.com/drone/drone/blob/master/engine/worker.go#L12:L24

While environment variables are typically not great for storing structured data, we can use a `dsn` style string for this. This string could use an inline yaml format:

```sh
BUILD_IMAGE="image: drone/drone-exec, command: [--clone, --cache, --build, --publish, --deploy]"
NOTIFY_IMAGE="image: drone/drone-exec, command: [--notify]"
```

The `dsn` string could be parsed by wrapping in curly braces like this:

```
imagev := fmt.Sprintf("{ %s }", os.Getenv("BUILD_IMAGE"))
yaml.Unmarshal([]byte(imagev), image)
```

This gives us the ability to provide structured data including the image, commands, entrypoint and default environment variables.

PR #1377, opened 2015-12-14
Make the build agent image name configurable

msteinert:
This should simplify things for people who want to hack on drone-exec.

### DIFF
--- a/engine/util.go
+++ b/engine/util.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"encoding/json"
+	"os"
 )
 
 func encodeToLegacyFormat(t *Task) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


engine/util.go
```diff
@@ -33,3 +34,11 @@ func encodeToLegacyFormat(t *Task) ([]byte, error) {
 	// }
 	return json.Marshal(t)
 }
+
+func getenvDefault(varname, value string) string {
```
tboerger: getEnvDefault :)

tboerger:
Maybe you want to add it for entrypoint as well?

Beside that a small NITPIK

engine/util.go
msteinert: I was following the style of the Go standard library:

https://golang.org/pkg/os/#Getenv

msteinert:
@tboerger I was thinking about entrypoint as well... But the implementation is a bit more complicated than this since it's a list. Maybe in a follow-on PR?

Without too much though, perhaps it could be a semi-colon separated list?

tboerger:
These were just my 2 cents... Maybe @bradrydzewski thinks differently about it

msteinert:
I just noticed this:

https://github.com/drone/drone/issues/1324

... plus some problems with this implementation. I'll take another crack at it later...

PR #1378, opened 2015-12-15
Make build runner configurable/pluggable

### LINKED ISSUE #1324: RFC Make Build Runner Configurable / Pluggable

msteinert:
Closes #1324

This change adds support for configuring the build, pull-request, and notify agent containers via the `BUILD_IMAGE`, `PULL_REQUEST_IMAGE`, and `NOTIFY_IMAGE` environment variables. The following inline YAML format is supported:

```
BUILD_IMAGE="image: drone/drone-exec, entrypoint: [/bin/drone-exec], cmd: [--clone, --cache, --build, --publish, --deploy]"
```

All of the keys are optional. If a key is omitted then the default value will be used.

If there is an error parsing YAML then the application will panic on startup.


bradrydzewski:
thanks! I'll take some time tomorrow to review a bit further and let you know if I have any other comments, but looks pretty good so far

bradrydzewski:
fixed in 0.5

PR #1383, opened 2015-12-22
Append reponame to ssh keys

### LINKED ISSUE #565: Add description to SSH keys
With possibly many projects and many deployment methods it might become handy to identify SSH public keys. To make things easier Drone should append a readable description to the key of each project.

nils-werner:
Currently drone SSH keys only show

    ssh-rsa AAA.....AAA

This PR appends a name to each repos SSH key so you can identify the key when you post it to a different service or `authorized_keys` file:

    ssh-rsa AAA.....AAA owner-repo@drone

Fixes #565

### DIFF
--- a/template/amber/repo_config.amber
+++ b/template/amber/repo_config.amber
@@ -68,7 +68,7 @@ block content
         div.row
             div.col-md-3 Public Key
             div.col-md-9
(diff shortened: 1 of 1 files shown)


tboerger:
LGTM

PR #1393, opened 2016-01-04
Corrects CC Menu url

donatj:
CC Menu URL was an empty placeholder. This completes it with the URL of the build.


tboerger:
LGTM

PR #1410, opened 2016-01-12
make gogs integration honor skip_verify

michaelgibson:
Currently the skip_verify option for the gogs client has no effect.
This change adds TLSClientConfig: &tls.Config{InsecureSkipVerify: true} 
to the Client's Transport.

### DIFF
--- a/remote/gogs/gogs.go
+++ b/remote/gogs/gogs.go
@@ -59,7 +59,7 @@ func (g *Gogs) Login(res http.ResponseWriter, req *http.Request) (*model.User, b
 		return nil, false, nil
 	}
 
-	client := gogs.NewClient(g.URL, "")
+	client := gogs.NewClient(g.URL, "", g.SkipVerify)
 
 	// try to fetch drone token if it exists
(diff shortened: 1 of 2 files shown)


tboerger:
:-1: Do not manipulate a vendored lib. This needs to be added to the upstream client first.

bradrydzewski:
@michaelgibson thanks for taking the time to send a patch! As mentioned, we should make the change upstream at [gogits/go-gogs-client](https://github.com/gogits/go-gogs-client) and then refresh the vendored copy

michaelgibson:
ok I think I see how things work here:
Looks like you already requested this change to the client.
https://github.com/gogits/go-gogs-client/issues/8

Which is now exposed via SetHTTPClient:
https://github.com/gogits/go-gogs-client/commit/8068507b500ea3ededecacc9d53c10da5f6ddd01#diff-0ec5cf868d4c7c045eb43e57a1b06419R37

I'll re-work it to use that instead and re-submit.
Thanks!

Issue #1426, opened 2016-01-20
Docker base image up for change?


cloudnautique:
Hi All, I wanted to ask if the community would be open to changing the base Docker image from "scratch" to alpine or busybox. I really like scratch, I do, but it limits extendability when the tools inside have an external dependency. For instance, if I need to wait for the existence of a configuration file that will contain all of my environment variables, I have no way to extend without making my own image because there is no shell.

 By creating a configuration container, that will generate the file with confd and share a new entrypoint script to source it, I could alter the behavior _WITHOUT_ modifying the upstream container and all of its goodness. By injecting configuration and alternate scripts, one does not need to worry about disturbing the core dependencies of the app.

something like:

```
drone:
  image: drone/drone:0.4
  entrypoint: /opt/scripts/entry
  volumes_from:
    - drone-config
...
drone-config:
  image: cloudnautique/drone-config:0.4
  volumes:
    - /opt/scripts
```

with a script in `/opt/scripts/` that looks like

```
#!/bin/bash
 while [ ! -e /etc/drone/dronerc ]; do
    sleep 1
 done

source /etc/drone/dronerc
exec /drone_static
```

Currently, something like this fails because there is no shell in drone:0.4. 

Alternatively, if the binary was placed in a folder instead of `/` the relationship could be flipped. So if /drone/bin/drone_static was where the binary lived, the container could be launched with `-v /drone` and picked up with volumes-from in the conf container. 

In general, I like the idea of having a shell for troubleshooting purposes, but I certainly see the benefits of `scratch`.

Thoughts? Alternate approaches? I will be happy to put in a PR one way or another.

bradrydzewski:
I would be open to busybox as the base layer since it only adds an extra 1mb to size. My primary concern with alpine is that it uses MUSL and therefore has some annoying incompatibilities when building drone in a standard Go environment that uses GLIBC.

cloudnautique:
Cool, do you run latest? Or like to pin versions of the FROM image?

bradrydzewski:
I'm find w/ latest

bradrydzewski:
does busybox have root ca's installed?

cloudnautique:
I'll take a look.

cloudnautique:
It did not have them, so I pulled down the previous images file.

bradrydzewski:
are there any special license considerations needed when using Busybox? The langauge is a bit vague and scary https://busybox.net/license.html

cloudnautique:
[...]

PR #1427, opened 2016-01-20
Change docker base image to busybox

cloudnautique:
This changes the base image to busy box so that it can be extended
for use with /bin/sh scripts, and basic troubleshooting.

In order to keep changes to a minimum, I pulled the ca-certificates file from the centurylink/ca-certs image source.

Discussed in #1426

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,12 @@
 #
 #     docker build --rm=true -t drone/drone .
 
-FROM centurylink/ca-certs
+FROM busybox
 EXPOSE 8000
 ADD contrib/docker/etc/nsswitch.conf /etc/
+
+# Pulled from centurylin/ca-certs source.
+ADD https://raw.githubusercontent.com/CenturyLinkLabs/ca-certs-base-image/master/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
 ENV DATABASE_DRIVER=sqlite3
 ENV DATABASE_CONFIG=/var/lib/drone/drone.sqlite


cloudnautique:
@bradrydzewski, any updates with this one?

bradrydzewski:
hey just a quick update -- I'm going to create a new `drone:0.5` tag next week and this change will be included in that tag, with `drone:0.4` continuing to use scratch. I'm being a bit overly paranoid, but this is just to avoid any unforeseen side-effects and keep 0.4 stable.

cloudnautique:
Sounds like a plan, thanks for the update. Is 0.5 going full release? Or is this the first step?

xiaods:
why not alpline?

gtaylor:
Alpine uses musl, which causes unexpected breakages with a non-trivial amount of software in the wild. It's been particularly nightmarish with our Python stuff.

xiaods:
https://hub.docker.com/r/frolvlad/alpine-glibc/

tboerger:
@gtaylor but that doesn't really matter for the drone image itself.

@xiaods currently is simply scratch, as we don't need much busybox should be enough to have at least a shell.

sylus:
Any status on getting this in? I unfortunately have to roll my own till can get a base image that can be extended :)

larstobi:
As a workaround until something like this gets merged you could copy the shell binary from an alpine container into the drone_container. Get to the paths like this, assuming you're using overlay storage:

    ls  /var/lib/docker/overlay/$(cat /var/lib/docker/image/overlay/layerdb/mounts/$(docker inspect -f '{{.Id}}' drone_agent)/mount-id)/merged

where "drone_agent" is the name of the drone container, which you could swap out with the name or reference to the alpine container with the shell binary.

cloudnautique:
@bradrydzewski, since Drone is a static binary, we could flip the change. Instead of putting the binary at `/` we could put the binary in a volume at '/drone/bin/` then users who want to do things like us can just do `--volume-from drone-container` with an alpine or busybox container. 

The change would have less risk, assuming that the code doesn't assume '/' :)

bradrydzewski:
closed by https://github.com/drone/drone/pull/2232

Issue #1411, opened 2016-01-12
implement skip_verify in gogs client.


michaelgibson:
Currently the skip_verify option for the gogs client has no effect.
This was initially due to a limitation in the go-gogs-client vendor api but has since been addressed.
https://github.com/gogits/go-gogs-client/commit/8068507b500ea3ededecacc9d53c10da5f6ddd01#diff-0ec5cf868d4c7c045eb43e57a1b06419R37

Just need to account for this in remote/gogs/gogs.go

PR #1433, opened 2016-01-21
Enable skip_verify functionality in gogs client

michaelgibson:
This change required updating the vendor client code for go-gogs-client.
I've tested this a bit and haven't seen any incompatibilities but may want to verify it won't break existing functionality.

Solves: https://github.com/drone/drone/issues/1411


bradrydzewski:
thanks!

Issue #1412, opened 2016-01-13
Drone is throwing up 500 on tag hooks for Gogs


cusspvz:
Drone is throwing up an 500 error whenever a tag hook is triggered by gogs.

`versions`
```
Drone 0.4.0
Gogs 0.8.10
```

`tag payload`
```json
{
  "secret": "hidden",
  "ref": "v1.2.6",
  "ref_type": "tag",
  "repository": {
    "id": 1,
    "name": "repository",
    "url": "https://gogs.company.com/project/repository",
    "ssh_url": "git@gogs.company.com:project/repository.git",
    "clone_url": "https://gogs.company.com/project/repository.git",
    "description": "",
    "website": "",
    "watchers": 13,
    "owner": {
      "name": "project",
      "email": "",
      "username": "project"
    },
    "private": true
  },
  "sender": {
    "login": "jose.moreira",
    "id": 1,
    "avatar_url": "https://gogs.company.com//avatar/6b2f1bb07eb368151c88af45764967df"
  }
}
```

`branch payload`
```json
{
  "secret": "hidden",
  "ref": "refs/heads/master",
  "before": "4d160779370bb6c974adb558b68d3703a1fa7059",
  "after": "a42e1bb384efffcc0d994a9c0e451e20d1ee61c9",
  "compare_url": "https://gogs.company.com/project/repository/compare/4d160779370bb6c974adb558b68d3703a1fa7059...a42e1bb384efffcc0d994a9c0e451e20d1ee61c9",
  "commits": [
    {
      "id": "a42e1bb384efffcc0d994a9c0e451e20d1ee61c9",
      "message": "Drone: fix yml\n",
      "url": "https://gogs.company.com/project/repository/commit/a42e1bb384efffcc0d994a9c0e451e20d1ee61c9",
      "author": {
        "name": "José Moreira",
        "email": "hidden",
        "username": "jose.moreira"
      }
    }
  ],
  "repository": {
    "id": 1,
    "name": "repository",
    "url": "https://gogs.company.com/project/repository",
    "ssh_url": "git@gogs.company.com:project/repository.git",
    "clone_url": "https://gogs.company.com/project/repository.git",
    "description": "",
    "website": "",
    "watchers": 13,
    "owner": {
      "name": "project",
      "email": "",
      "username": "project"
    },
    "private": true
  },
  "pusher": {
    "name": "jose.moreira",
    "email": "hidden",
    "username": "jose.moreira"
  },
  "sender": {
[...]

PR #1448, opened 2016-01-31
Manage tag hooks from gogs

bsauvajon:
As discussed in issue #1412 drone does not support tag event with gogs.
With this commit tag event are correctly managed, you just have to activate it on the webhook property on gogs and in the project property on drone.
Also fix a panic message logged when a branch was created


remote/gogs/helper.go
```diff
@@ -66,13 +66,29 @@ func buildFromPush(hook *PushHook) *model.Build {
 		hook.Repo.Url,
 		fixMalformedAvatar(hook.Sender.Avatar),
 	)
+
+	eventType := "Undefined"
+	message := "No message"
+
```
bradrydzewski: perhaps instead of dummy strings just use empty values:

```
var eventType string
var message string
```

remote/gogs/helper.go
```diff
@@ -66,13 +66,29 @@ func buildFromPush(hook *PushHook) *model.Build {
...
+	switch {
+	case hook.RefType == "tag":
+		eventType = model.EventTag
+		message = "Create Tag " + hook.Ref
```
bradrydzewski: for other systems, the message for the tag is the commit message for the sha being tagged. Is this available in the gogs request?

remote/gogs/helper.go
```diff
@@ -66,13 +66,29 @@ func buildFromPush(hook *PushHook) *model.Build {
...
+		message = hook.Commits[0].Message
+	}
+
 	return &model.Build{
-		Event:     model.EventPush,
+		Event:     eventType,
 		Commit:    hook.After,
 		Ref:       hook.Ref,
 		Link:      hook.Compare,
 		Branch:    strings.TrimPrefix(hook.Ref, "refs/heads/"),
```
bradrydzewski: for consistency with GitHub and others, I believe we should make the branch the tag value (ie v1.0.0). Also I believe in GitHub and others the ref for the tag is the full ref, `refs/tags/v1.0.0`

remote/gogs/helper.go
```diff
@@ -66,13 +66,29 @@ func buildFromPush(hook *PushHook) *model.Build {
...
+	message := "No message"
+
+	switch {
+	case hook.RefType == "tag":
+		eventType = model.EventTag
+		message = "Create Tag " + hook.Ref
+	case hook.RefType == "branch":
+		eventType = model.EventPush
+		message = "Create Branch " + hook.Ref
+	default:
```
bradrydzewski: should a branch creation event trigger a build in drone? with GitHub we ignore these, because otherwise it would trigger two builds, since we get an event for branch creation AND for the push

bradrydzewski:
thanks for taking the time to create and submit this patch! I added some notes for some potential tweaks we'd need to make, just to ensure consistency across all implementations (gogs, github, etc), as well as a couple of questions.

remote/gogs/helper.go
bsauvajon: For a tag creation there is no commit message in the payload, here is the example posted in issue #1412 : 

```
{
  "secret": "hidden",
  "ref": "v1.2.6",
  "ref_type": "tag",
  "repository": {
    "id": 1,
    "name": "repository",
    "url": "https://gogs.company.com/project/repository",
    "ssh_url": "git@gogs.company.com:project/repository.git",
    "clone_url": "https://gogs.company.com/project/repository.git",
    "description": "",
    "website": "",
    "watchers": 13,
    "owner": {
      "name": "project",
      "email": "",
      "username": "project"
    },
    "private": true
  },
  "sender": {
    "login": "jose.moreira",
    "id": 1,
    "avatar_url": "https://gogs.company.com//avatar/6b2f1bb07eb368151c88af45764967df"
  }

```}

remote/gogs/helper.go
bsauvajon: In gogs the ref sent for a tag creation is just the tag name, so I think here Branch will have the value of the tag. Maybe it would be eaisier to read if Branch was set before in the `switch`

remote/gogs/helper.go
bsauvajon: With more test, I see that gogs works like github, two builds are triggered. Yuo are true, branch creation event should definitiely be ignored

bradrydzewski:
I'm going to close this because the PR needs to be squashed. Right now it updates so many files that it crashes the GitHub UI preventing me from reviewing the change :)

bsauvajon:
Hum, I wanted to correct my email adress on the commits, but I didn't see it created a big mess on the PR.
I' will submit a new one soon, integrating your comments ;)

PR #1449, opened 2016-01-31
Gitlab avatars

Bugagazavr:


martinsefcik:
One more thing needs to be fixed ... missing avatars in Search suggestions popup, but I'm not sure if it's just Gitlab related issue or general drone issue.

Anyway, good job Bugagazavr!

PR #1453, opened 2016-02-01
Dot names fix

Bugagazavr:


remote/gitlab/client/util.go
```diff
@@ -5,8 +5,18 @@ import (
 	"strings"
 )
 
+var encodeMap = map[string]string{
+	".": "%252E",
```
tboerger: What about replacing slashes as before?

bradrydzewski:
LGTM just needs a CLA

remote/gitlab/client/util.go
Bugagazavr:
```go
value = url.QueryEscape(value)
```

Already do this

Bugagazavr:
@bradrydzewski looks like cla acepted, this because my first commit has been sent with wrong credentionals, I use ammed to fix this, but badge cla is old, just look to checks

Issue #1234, opened 2015-10-13
Improve Dashboard, Repo List, Feed


bradrydzewski:
One major change that comes with 0.4 is that we no longer sync your complete repository list in the database (for reasons described in #776). Instead we use the GitHub API to fetch your repository list and your repository permissions live.

One disadvantage to this approach is that we don't have a simple way (like a database join) to tell Drone to quickly give me a list of all repositories I have access to, or a build feed for all repositories I have access to.

I believe the solution is to fetch the list of all repositories from GitHub and then to create a really big `IN` statement. The `IN` operator will get passed the list of all repository names from GitHub as demonstrated in the following example:

```sql
select *
from repos
where repo_full_name IN (?,?,?,?,?,?.....................,?,?,?,?)
```

I even have a function already stubbed out in the code:

```Go
func GetRepoListOf(db meddler.DB, listof []string) ([]*Repo, error) {
	var repos = []*Repo{}
	var length = len(listof)
	var qs = make([]string, length, length)
	var in = make([]interface{}, length, length)
	for i, repo := range listof {
		qs[i] = "?"
		in[i] = repo
	}
	var stmt = "SELECT * FROM repos WHERE repo_id IN (" + strings.Join(qs, ",") + ")"
	var err = meddler.QueryAll(db, &repos, database.Rebind(stmt), in...)
	return repos, err
}
```

According to the documentation sqlite can hold a maximum of 999 parameters in the `IN` statement, and mysql and postgres [1] can hold tens of thousands of parameters. I remember doing this about 10 years ago to query an Oracle database with a few million rows and it worked quite well. So I'm pretty confident this approach is technically feasible, however, I still want to understand the performance implications on large datasets for the database vendors that we support.

Until I have benchmarks that support the above design, we implement the following workarounds:

1. The `/user/repos` endpoint returns a list of all repositories to which the user has authored a commit, based on data in the `builds` table
2. The `/user/feed` endpoint returns a list of all builds the user has authored

I am proactively logging this issue to explain why the behavior is currently different and might appear broken, to assure everyone this is just a temporary workaround, and that I hope to have the proposed changes in place prior to the final 0.4 release.

---

[1] consider described optimization for postgres https://www.datadoghq.com/blog/100x-faster-postgres-performance-by-changing-1-line/

bradrydzewski:
This is proving to be a bit more difficult than I anticipated. I very much like not storing user repository permissions in the database. Storing these in an in-memory cache with expiration has worked out really well.

[...]

PR #1460, opened 2016-02-04
Add master build list endpoint for wall

scottferg:
@bradrydzewski this adds a top level build list (similar to what we added to v3) that is necessary for drone-wall to operate.

### DIFF
--- a/controller/build.go
+++ b/controller/build.go
@@ -20,6 +20,15 @@ import (
 	"github.com/drone/drone/router/middleware/session"
 )
 
+func GetAllBuilds(c *gin.Context) {
+	builds, err := store.GetBuildMasterList(c)
+	if err != nil {
+		c.AbortWithStatus(http.StatusInternalServerError)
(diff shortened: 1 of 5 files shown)


bradrydzewski:
@scottferg I think you can use the existing `/api/user/feed` endpoint for this

scottferg:
Yup! Totally can. Serves me right for not reading the docs.

bradrydzewski:
actually, it isn't in the docs haha :)

bradrydzewski:
for reference, the reason this wasn't documented was I didn't want to advertise the endpoint until it was benchmarked in case it needed to be removed or tweaked. The benchmark was positive (you can see at https://github.com/drone/drone/issues/1234) so you should be good to rely on this endpoint going forward.

scottferg:
Ha cool. I don't feel like as big of an idiot. This will make getting the wall up to snuff much easier either way.

PR #1461, opened 2016-02-05
The latest drone upstream, which includes docker connection pooling

mikejihbe:


mikejihbe:
Disregard. Meant to push this to my fork, we're just catching up.

Issue #1462, opened 2016-02-05
Different variable injection on conditions in build YAML stanza


desktophero:
First `n00b to Drone`. I'll jump into the code to see what I can see. I wanted to get an issue opened in parallel. Totally convinced it's something I'm doing! :smile: but perhaps this is an edge case.

Issue I am seeing is that when I have a build on `push`, environment variables inject correctly. When I apply a condition to add steps on a `pull_request`, the variable injection does not happen.

Note: the data below is mocked up based on the yml files I have

`.drone.yml`
```yaml
build:
  always_run:
    image: <image_name>
    commands:
      - export KEY_NAME="$$OS_KEY_NAME"
      - env | sort
  pull_requests:
    image: <image_name>
    when:
      event: pull_request
    commands:
       - export KEY_NAME="$$OS_KEY_NAME"
       - env | sort
```
`secrets.yaml`
```yaml
environment:
  - OS_KEY_NAME=BOOM_GOES_THE_DYNAMITE
```

Build log for `push` (`always_runs`):
```
$ env | sort
OS_KEY_NAME=BOOM_GOES_THE_DYNAMITE
```

Build log for `pull_request` (`pull_requests`):
```
$ env | sort
OS_KEY_NAME=7OS_KEY_NAME
```

In the `push` scenario (labeled `always_run`), all the environment variables are a-okay. In the `pull_request` condition (labeled `pull_requests`), the environment variables have that `7` (sometimes `8`, sometimes `9`) and then the name of the variable instead of the value.

bradrydzewski:
Thanks for getting in touch and trying out drone! Secrets are not injected into pull requests. For these sort of questions please use discuss.drone.io or gitter.im/drone/drone since we like to reserve the issue log for confirmed bugs or feature requests. Thanks!

bradrydzewski:
forgot to mention this is the issue you'll want to subscribe to https://github.com/drone/drone/issues/1250 but otherwise, we'll look forward to seeing you in our gitter room!

desktophero:
I searched multiple ways to see if an issue was already out there and help flush out if this is expected behavior. Missed it (GHE search-foo fail). Thank you for the link!

bradrydzewski:
no worries! sorry I don't have an immediate fix for you, but hopefully we'll have that issue resolved in the coming weeks. Cheers!

PR #1464, opened 2016-02-05
Env variable doc update

desktophero:
I opened this issue earlier (https://github.com/drone/drone/issues/1462) ... after a couple of hours of WTF moments trying to discern my `pull_request` build option and then getting some (greatly appreciated) direction on the "why" the behavior is happening, it seemed like a good idea to submit some doc update to help others not encounter this and/or explain in docs why they are seeing something as designed. 

I hope this is helpful :smile:

### DIFF
--- a/docs/build/env.md
+++ b/docs/build/env.md
@@ -51,7 +51,7 @@ publish:
 A subset of bash string substitution operations are emulated:
 
 * `$$param` parameter substitution
-* `$${param}` parameter substitution (same as above) 
+* `$${param}` parameter substitution (same as above)
 * `"$$param"` parameter substitution with escaping
 * `$${param:pos}` parameter substition with substring
 * `$${param:pos:len}` parameter substition with substring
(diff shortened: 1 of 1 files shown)


desktophero:
@jackspirou ... from my issue earlier. Pulling you in for review as well

bradrydzewski:
to avoid duplicate details, perhaps we link from the environment section to secrets page? http://readme.drone.io/usage/secrets/#pull-requests:0d943d142dc6bcdd2e544b095a49f72b

environment variables are technically different than secrets, but if that is the default place that people are looking it would be great to link between the two

gtaylor:
Are these docs even used anymore? I thought they were pretty much all over here https://github.com/drone/docs except for the swagger.yml?

bradrydzewski:
@gtaylor good point, I clicked the link from gmail and didn't notice which repo we were in. There is one important piece of information that hasn't been ported over yet (using the cli to register remote worker nodes) but otherwise we can delete the docs section from this repo

@desktophero we've actually move the docs to https://github.com/drone/docs ... sorry for the confusion!

desktophero:
I can move/rewrite/enhance based on the feedback ... outside of that, cool with the content of PR?

gtaylor:
> There is one important piece of information that hasn't been ported over yet (using the cli to register remote worker nodes)

I wonder if we should even bother, given the push-to-pull conversion? I'm assuming there won't be a need to explicitly register the workers on the CLI after that?

desktophero:
Closing this ... I'll submit a new one based on the comments and feedback (and in the right repo). Thanks!

PR #1467, opened 2016-02-05
Add public mode

msteinert:
This change adds a public mode, which makes build logs visible to users who are not logged in. To activate public mode, set the environment variable `PUBLIC_MODE=true`

In my current setup (Gitlab & Drone running behind a corporate firewall) most users don't care about Drone or logging into Drone, they just want to get their changes merged. In the case of a failed build I would like to be able to link to the build logs without the "confusion" of having them log in.

This allows me to send a build log link to a staff member who is not be a Gitlab user.

### DIFF
--- a/router/middleware/session/repo.go
+++ b/router/middleware/session/repo.go
@@ -2,6 +2,7 @@ package session
 
 import (
 	"net/http"
+	"os"
 
 	"github.com/drone/drone/model"
 	"github.com/drone/drone/remote"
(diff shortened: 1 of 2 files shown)


bradrydzewski:
LGTM

bradrydzewski:
cc @chaseadamsio I think this was something you wanted as well

jackspirou:
Thanks @msteinert for this.  @bradrydzewski this will be helpful for us as well :)

chaseadamsio:
:+1: indeed I did. :) Thanks @bradrydzewski

PR #1468, opened 2016-02-06
Login page instead of 404 for non-logged in users on repos page

### LINKED ISSUE #1266: Add login redirect for 404s
If a user isn't logged in and follows a Drone link it will 404 ala GitHub. This can cause some confusion for users as they don't see anything and don't know they aren't logged in. It would be great to have a login prompt if they're not logged in to avert any questions about access.

kustosz:
Hi!

This pull request fixes #1266. Note that in order to implement the "go back after login" behavior, I needed to build some more magic on top of your router path normalization, if you see any better solution (or have any hints on how to improve this one), please let me know.

Cheers,
Marcin

### DIFF
--- a/router/middleware/session/repo.go
+++ b/router/middleware/session/repo.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/drone/drone/model"
 	"github.com/drone/drone/remote"
+	"github.com/drone/drone/shared/httputil"
 	"github.com/drone/drone/shared/token"
 	"github.com/drone/drone/store"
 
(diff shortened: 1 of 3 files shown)


bradrydzewski:
Thanks for taking a look at this. The behavior I'd like to see is that a 404 is displayed, but includes a link to login on the page instead of auto-redirecting. This more closely mirrors the GitHub behavior.

This should therefore end up being mostly template code with little (maybe no) Go code changes required, which means it will be compatible with the work we are doing on the new UI at drone/drone-ui

Thanks!

jhasse:
Any news on this? IMHO a 404 would be misleading, as it is possible that the page DOES exist.

bradrydzewski:
@jhasse 404 is the intended behavior because we don't want to leak the existence of a private repository. This is the same behavior you will see in GitHub for security and private reasons, which we are mirroring. There are changes to the UI to include a login link, which should help.

I'm going to close this PR because it won't get merged as implemented and we have a new UI that is being released with 0.5 that provides a login link in addition to the 404

PR #1482, opened 2016-02-12
resume waiting for the container to complete

macb:
This backports the resuming of tailing logs introduced in https://github.com/drone/drone-exec/pull/26 as well as the marking of incomplete builds as errors similar to https://github.com/drone/drone-exec/commit/7dc4524fc4c8115d1dcc6cc5027f75606a3069ea.

### DIFF
--- a/shared/docker/docker.go
+++ b/shared/docker/docker.go
@@ -2,8 +2,6 @@ package docker
 
 import (
 	"errors"
-	"io"
-	"io/ioutil"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/samalba/dockerclient"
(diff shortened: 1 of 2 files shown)


shared/docker/docker.go
```diff
@@ -1,10 +1,12 @@
 package docker
 
 import (
+	"errors"
 	"io"
 	"io/ioutil"
 
 	"github.com/samalba/dockerclient"
+	"github.internal.digitalocean.com/digitalocean/doge.git/log"
 )
```
bradrydzewski: I'm guessing the editor auto-completed the wrong logging library :)

bradrydzewski:
@macb while working on the agent I ended up altering the `Wait` function as follows:

```
func Wait(client dockerclient.Client, name string) (*dockerclient.ContainerInfo, error) {
	<-client.Wait(name)
	return client.InspectContainer(info.Id)
```

When I originally wrote this `Wait` function the Docker client didn't expose the `client.Wait` endpoint so I had to simulate by reading the logs and writing to `ioutil.Discard`. So this might simplify things even further!

macb:
That would definitely be simpler than the discard logic and might even be more reliable. I can switch this around to use that and maintain the retry logic if that seems reasonable?

macb:
Switched it around to use the `Wait` method. Left in the retries since it seemed possible to hit an intermittent failure due to a network hiccup https://github.com/samalba/dockerclient/blob/master/dockerclient.go#L453-L469.

bradrydzewski:
looks good

PR #1488, opened 2016-02-16
Add ability to hide Archived projects from Gitlab

athieriot:
I currently have a fair bit of Archived projects when using GitLab with Drone.

This PR provides a way to hide archive projects from the listing using GitLab API options. The default behaviour stay the same and it has to be voluntarily activated via the ```hide_archives``` connection string parameter.


tboerger:
LGTM

bradrydzewski:
LGTM

gtaylor:
@athieriot Can you add that to the list here? http://readme.drone.io/setup/remotes/gitlab/

athieriot:
Ah I was wondering if the two documentations where somehow updated together.
Guess not. Will do then.

athieriot:
https://github.com/drone/docs/pull/105

PR #1490, opened 2016-02-16
Removing legacy docs.

gtaylor:
People keep updating these thinking that this is where they reside. Time to clean house!

Left swagger.yml alone, since I wasn't sure if we were using it.


tboerger:
The go generate part for the docs is still part of the build process.

gtaylor:
This is going to need some additional checking, but the latest commit is my attempt to remove the last remaining pieces. Once we get to the point of being OK with the PR, I can rebase + squash this. For now we'll do this iteratively until I get it right.

tboerger:
Still missing something. go-bindata doesn't get executed anymore.

bradrydzewski:
I removed as part of some 0.5 changes that I backported :)

PR #1499, opened 2016-02-18
[WIP] Phabricator remote integration

tboerger:
This is currently work in progress. I have started to integrate phabricator as a remote into drone. The authentication already works, but now we need to find ways to add the other missing parts as well.

The client parts are taken from the Gitlab client.

### DIFF
--- a/remote/phabricator/client/phabricator.go
+++ b/remote/phabricator/client/phabricator.go
@@ -1,0 +1,114 @@
+package client
+
+import (
+	"bytes"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+type Client struct {
+	BaseUrl string
+	Token   string
+	Client  *http.Client
+}
+
+func New(baseUrl, token string, skipVerify bool) *Client {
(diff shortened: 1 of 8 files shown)


tboerger:
It still doesn;t really work is much more complicated. As nobody requested that I will close that for now.

tlwt:
is there any suggested way on working with phabricator? I really appreciate the drone as well as phabricator and think a combination of both would be great!

tboerger:
An phabricator remote driver must implement the interface defined at https://github.com/drone/drone/blob/master/remote/remote.go. 

I don't know how open brad is for new remote drivers or if he wants to enable an external service implementation instead of this.

tlwt:
If you mean by external services that you can create some random web hook and drone kicks off - that would be great as well, I suppose.

zicklag:
I would *really* like to be able to integrate Phabricator with Drone somehow.

zicklag:
Cool, I might look into that after 0.9 comes out, then. :slightly_smiling_face:

mavimo:
@bradrydzewski I suppose should be https://github.com/drone/drone-scm

bradrydzewski:
oops, it is https://github.com/drone/go-scm

PR #1501, opened 2016-02-19
Enable restrict_orgs

flyinprogrammer:
This allows you to standup an instance of drone and have it restrict the repos it sees to the orgs you specify.

It's naive.
There are no tests.
And we could argue about names for days.

If we want this I'm willing to work on this!

// Sidestory
I run drone against an internal github. My user has access to over 1800 repos. Without this feature my life is sad. I can't imagine I'm alone, but maybe I am.

### DIFF
--- a/remote/github/github.go
+++ b/remote/github/github.go
@@ -25,15 +25,16 @@ const (
 )
 
 type Github struct {
-	URL         string
-	API         string
-	Client      string
-	Secret      string
-	Orgs        []string
-	Open        bool
-	PrivateMode bool
-	SkipVerify  bool
-	GitSSH      bool
+	URL          string
+	API          string
+	Client       string
+	Secret       string
+	Orgs         []string
+	Open         bool
+	PrivateMode  bool
+	SkipVerify   bool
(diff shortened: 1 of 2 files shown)


adamkdean:
This would help me out immensely. Signal-to-noise ratio right now is just unbearable.

Thanks!

bradrydzewski:
GitHub gives you the ability to restrict organizations at the user level:

screen shot 2015-12-08 at 3 23 59 pm

And at the organization level:
https://help.github.com/articles/about-third-party-application-restrictions/

We are also planning UI improvements that remove inactive repositories from the dashboard. I believe these existing GitHub features, in addition to existing GitHub capabilities, should give you more control over what appears in Drone.

flyinprogrammer:
Sadly it appears Github Enterprise doesn't support the Third-Party organization restrictions?

tboerger:
I think https://github.com/drone/drone/pull/1536 should be better instead of restricting access at boot time

flyinprogrammer:
This is stale code.

PR #1504, opened 2016-02-22
Added support for new attributes for GitLab

Bugagazavr:


bradrydzewski:
let me know if this is good to merge and I'll do so today

Bugagazavr:
@bradrydzewski i think yes

PR #1507, opened 2016-02-27
Use the `nonword` tokenizer in Bloodhound search

tillarnold:
As we discussed on gitter I had a look at Bloodhound and I used the nonword tokenizer for the `datumTokenizer`. This makes the search a little more fuzzy and allows to search for only the name of repos as well as for the full name.

Note that typeahead.js caches the search in localStorage so to test this change you have to clear you localStorage.

### DIFF
--- a/static/scripts/search.js
+++ b/static/scripts/search.js
@@ -2,7 +2,7 @@ var repoExpr = /.+\/.+/;
 
 var remoteRepos = new Bloodhound({
     queryTokenizer: Bloodhound.tokenizers.whitespace,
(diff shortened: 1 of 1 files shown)


tillarnold:
@bradrydzewski What do you think? Is this ok or do I need to change something?

tboerger:
LGTM

bradrydzewski:
we've moved to the user interface to the [drone-ui](https://github.com/drone/drone-ui) and switched to a new technology stack and no longer using jquery / bloodhound and instead using react with custom search.

PR #1511, opened 2016-03-02
Limit github authz scopes requested by drone

crhym3:
I've heard complains drone is asking for too many permissions, including direct code access.

Would the following still suffice for all the features drone currently provides for github?
screen shot 2016-03-02 at 10 48 07 am

### DIFF
--- a/remote/github/github.go
+++ b/remote/github/github.go
@@ -21,7 +21,7 @@ import (
 const (
 	DefaultURL   = "https://github.com"
 	DefaultAPI   = "https://api.github.com"
-	DefaultScope = "repo,repo:status,user:email"
+	DefaultScope = "repo_deployment,repo:status,write:repo_hook,user:email"
 )
 
 type Github struct {


bradrydzewski:
@crhym3 without the `repo` scope we would be unable to use drone with private repositories. The `repo` scope is required to clone a private repository, check individual private repository permissions, and helps generate the repository list in the user interface.

If you are only using drone with public repositories, however, and don't need access to private repositories perhaps we can instead make the scope configurable? 

We could add the following configuration parameters to our GitHub configuration:
https://github.com/drone/drone/blob/master/remote/github/github.go#L55

```Go
github.Scopes = params.Get("scope")
if github.Scopes == "" {
    github.Scopes = DefaultScopes
}
```

In fact, this is what I did with lgtm.co
https://lgtm.co/docs/install/#configuration:ee40007639ba1d9861b5cc3e7ade0aef

crhym3:
I see. Too bad. Yeah, I'll change this to use `scope` param and fall back to defaults.
Thanks!

crhym3:
Done. @bradrydzewski PTAL.

PR #1514, opened 2016-03-04
Add basic exlusion to build matrix

jbakosi:
I took a crack at exclusions for the build matrix per [my conversation](https://gitter.im/drone/drone?at=56d5f11844ba0664026a332d) with @bradrydzewski and this is a first shot.

Please bear with me as I've never written code in go before.

I also added unit tests that test also the _contents_ of the build matrix instead of only testing the correct size.

Please let me know what you think.

### DIFF
--- a/yaml/matrix/matrix.go
+++ b/yaml/matrix/matrix.go
@@ -14,6 +14,9 @@ const (
 // Matrix represents the build matrix.
 type Matrix map[string][]string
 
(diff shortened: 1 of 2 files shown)


tboerger:
LGTM

vtolstov:
does it possible to update this pr?

tboerger:
@bradrydzewski maybe something we want to have?

PR #1520, opened 2016-03-11
Do not attempt to set memory.swappiness to zero

msteinert:
This is the same issue solved here: drone/drone-exec@94fd9ba10d707b8f959feb097e197f5a8fa0349e

This issue was introduced by this commit: b5dcfc4edd6ad405527cc29c197ddb240d3d3930

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -301,7 +301,8 @@ func (e *engine) runJob(c context.Context, r *Task, updater *updater, client doc
 		Cmd:        args,
 		Env:        e.envs,
 		HostConfig: dockerclient.HostConfig{
-			Binds: []string{"/var/run/docker.sock:/var/run/docker.sock"},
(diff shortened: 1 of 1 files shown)


msteinert:
Without this fix (on Ubuntu 14.04) all builds will fail to start and you'll see something like this in the logs:

```
time="2016-03-11T14:42:53Z" level=error msg="Error starting notification container drone_build_475_notify. 500 Internal Server Error: Cannot start container a8e56457a99c2b33f8135ae5deaf8e18ab9230f6e9d529242fa45111144cc93c: [9] System error: write /sys/fs/cgroup/memory/docker/a8e56457a99c2b33f8135ae5deaf8e18ab9230f6e9d529242fa45111144cc93c/memory.swappiness: invalid argument\n"
```

bradrydzewski:
thanks!

PR #1522, opened 2016-03-12
Add merge conflict handling for Github pull requests

### LINKED ISSUE #1420: Pull Request builder does not fail gracefully if the Pull Request has a merge conflict
The Pull Request builder fetches the merged ref of the Pull Request:
`git fetch --no-tags --depth=50 origin +refs/pull/:id/merge:`. This is a good thing, I agree. But if the PR has a merge conflict, the only error you see is:

```
$ git remote add origin https://github.com/myorg/my-project
$ git fetch --no-tags --depth=50 origin +refs/pull/143/merge:
fatal: Couldn't find remote ref refs/pull/143/merge
[info] Pulling image plugins/drone-cache:latest
[info] build failed (exit code 1)
```

This error really had me puzzled as I thought my network setup was somehow not working. So I think Drone should catch unmergable PRs. The (Go) Github API [gives you the ability to check if a PR is mergable](https://github.com/google/go-github/blob/master/github/pulls.go#L33).

Drone could print to the build log that the PR is unmergable because of merge conflicts and the developer should fix the conflicts first, otherwise the PR ref cannot be checked out.

tolgap:
Let me preface by saying this is my first time dabbling into Golang.

We introduced Drone into our company, but I still have some confused first-time users that get merge conflicts, and see the vague "remote ref not found". I tried to fix this using `github.PullRequest.Mergeable` and throw an error if it's not mergeable and have the `hook` controller take care of the resulting error.

Fixes #1420.


bradrydzewski:
I appreciate you taking the time to submit a pull request! Some thoughts below

The merge fields in the payload are not always populated. In the last PR payload that drone processed I see the following for one of my pull requests:

```
    "mergeable": null,
    "mergeable_state": "unknown",
```

This below of code would result in a panic when it tries to de-reference the nil pointer. It would need to check for a nil value first.

```
mergeable := *hook.PullRequest.Mergeable
```

This mergeable value, from my understanding, is populated by GitHub asynchronously and in some cases may be nil or unknown which means this approach may work most of the time, but not all of the time. I don't see this as an issue, by the way, but thought it was worth mentioning.

The bigger concern is that this will exit without starting a build and may be more confusing to the end user than a git clone failure. Instead of getting asked "why did the git clone fail" I can see being asked "why didn't the build execute for the pull request" ... "did drone receive the hook" ... we will be replacing a visible error message with a silent error message that will only be visible via the drone server logs.

We need to decide which provides a better user experience. Is it worse to silently ignore the build or to execute a build knowing it will result in a git merge error message. I'm going to open this up to the drone maintainers to weigh in on.

tolgap:
@bradrydzewski thanks for the explanation. I don't yet have enough knowledge about the entirety of drone, but this implementation is indeed broken if `mergeable` is nullable. I'll close this :)

Issue #1429, opened 2016-01-21
feature request: download full log from Drone CI web interface


fracting:
I use Drone CI to build large project, for example GCC:

http://52.8.143.18/fracting/gcc-dummy/1
(BTW, this is Win32 version of GCC compiled by Win32 version of GCC on Wine on Linux in Drone CI)

The log is very long, and it seems to be truncated in the web interface.
Could we implement a "download full log" feature?

Thanks!

bradrydzewski:
although we don't have a download button in the UI we do have an endpoint to download the full log `/api/repos/fracting/gcc-dummy/1/1?full=true`

fracting:
Thanks for the response, but sorry I still don't understand, I tried `http://52.8.143.18/api/repos/fracting/gcc-dummy/1/1?full=true` but it does not work, I should be missing something, could you please give me an example how to use it?

Thank you!

bradrydzewski:
sorry I gave you the wrong url pattern. use `/api/repos/{owner}/{name}/logs/{number}/1?full=true`

fracting:
Thanks, I tried http://52.8.143.18/api/repos/fracting/gcc-dummy/logs/1/1?full=true and it works this time!

BTW, the log seems to be truncated, (as well as the log in web interface), but it might be either my own problem or a separate bug, I'll keep monitor on that and report an issue if needed.

tboerger:
Maybe the database column is too small?

fracting:
I downloaded the truncated log, and the size is exactly 2000000 bytes.

I read GetBuildLogs()

```
        defer r.Close()
        if full {
                io.Copy(c.Writer, r)
        } else {
                io.Copy(c.Writer, io.LimitReader(r, 2000000))
        }
}
```

Is it reasonable to ask upstream Drone increasing the number, or should I only do it in my fork?

To me it seems like it is reasonable to have a full log large enough for compiling gcc using gcc.

Thanks!

bradrydzewski:
> Is it reasonable to ask upstream Drone increasing the number, or should I only do it in my fork?

We've observed that > 2MB of logs will crash the browser. This is why we limit the amount to 2MB. The reason we have the `full` flag is so that we can (eventually) add a button to the UI to download the full logs. This is how most other CI systems handle the issue.

tboerger:
Maybe we need to group the output somehow like travis is doing that

fracting:
[...]

PR #1524, opened 2016-03-16
controller: Properly parse "full" query in GetBuildLogs.

fracting:
Hi,

This is related to https://github.com/drone/drone/issues/1429#issuecomment-176020154

I guess most Drone users doesn't build large projects like gcc ;)
We use Drone to build many large projects like gcc, and it expose bugs like logs more than 2MB ;)

See https://github.com/gin-gonic/gin#querystring-parameters for correct usage to parse a query with gin.

Thanks for the great job on Drone!

### DIFF
--- a/controller/build.go
+++ b/controller/build.go
@@ -82,7 +82,7 @@ func GetBuildLogs(c *gin.Context) {
 
 	// the user may specify to stream the full logs,
 	// or partial logs, capped at 2MB.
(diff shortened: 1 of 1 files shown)


fracting:
Test case here: https://tea-ci.org/api/repos/fracting/longtest/logs/2/1?full=true

Before this patch, full log doesn't really work for me.

bradrydzewski:
good catch, thanks!

PR #1532, opened 2016-03-22
Add tuneable build file parameter

frapposelli:
This let administrators override the build config file (to be something else than `.drone.yml`) by adding a `BUILD_CONFIG_FILE` environment variable to `dronerc`.

There are no changes to how secrets work, they will be sourced from `.drone.sec` and merged as usual.

I'm working on the use case where a repo is hooked to two drone instances and they have to share a single secrets file. This is currently not working.

### DIFF
--- a/remote/bitbucket/bitbucket.go
+++ b/remote/bitbucket/bitbucket.go
@@ -237,9 +237,8 @@ func (bb *Bitbucket) Perm(u *model.User, owner, name string) (*model.Perm, error
 	return perms, nil
 }
 
-// Script fetches the build script (.drone.yml) from the remote
-// repository and returns in string format.
-func (bb *Bitbucket) Script(u *model.User, r *model.Repo, b *model.Build) ([]byte, []byte, error) {
+// File fetches a file from the remote repository and returns in string format.
+func (bb *Bitbucket) File(u *model.User, r *model.Repo, b *model.Build, f string) ([]byte, error) {
 	client := NewClientToken(
 		bb.Client,
 		bb.Secret,
@@ -249,19 +248,12 @@ func (bb *Bitbucket) Script(u *model.User, r *model.Repo, b *model.Build) ([]byt
 		},
 	)
 
-	// fetches the .drone.yml for the specified revision. This file
-	// is required, and will error if not found
-	config, err := client.FindSource(r.Owner, r.Name, b.Commit, ".drone.yml")
+	config, err := client.FindSource(r.Owner, r.Name, b.Commit, f)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	// fetches the .drone.sec for the specified revision. This file
-	// is completely optional, therefore we will not return a not
-	// found error
-	sec, _ := client.FindSource(r.Owner, r.Name, b.Commit, ".drone.sec")
-
-	return []byte(config.Data), []byte(sec.Data), err
+	return []byte(config.Data), err
 }
 
 // Status sends the commit status to the remote system.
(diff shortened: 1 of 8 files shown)


frapposelli:
Just came to the conclusion that there is no sensible way to sync two drone instances to use the same secrets file, I'm amending the PR with a change that aligns secrets filename to the config file

frapposelli:
Changed the behavior to let two or more instances to cohexist but not share any `.sec` file (they can obviously share the originating `secrets.yml` file). 

If the administrator uses `BUILD_CONFIG_FILE` then the secrets file is going to be a file with the same name but with extension `.sec`.

Example:

```
REMOTE_DRIVER=gogs
REMOTE_CONFIG=http://172.16.94.133:10080?open=true
BUILD_CONFIG_FILE=.foobar.yml
```

Then it will look for `.foobar.yml` and `.foobar.sec` in the root of the git repo.

remote/bitbucket/bitbucket.go
```diff
@@ -251,15 +254,23 @@ func (bb *Bitbucket) Script(u *model.User, r *model.Repo, b *model.Build) ([]byt
 
 	// fetches the .drone.yml for the specified revision. This file
 	// is required, and will error if not found
-	config, err := client.FindSource(r.Owner, r.Name, b.Commit, ".drone.yml")
+	BUILD_CONFIG_FILE := os.Getenv("BUILD_CONFIG_FILE")
```
tboerger: Please use envconfig instead of one environment functions

bradrydzewski:
thanks, some ideas to slightly change the implementation

I'm thinking we rename `Script` to `File` inside the remote package:
https://github.com/drone/drone/blob/master/remote/remote.go

```diff
-Script(u *model.User, r *model.Repo, b *model.Build) ([]byte, []byte, error)
+File(u *model.User, r *model.Repo, b *model.Build, path string) ([]byte, error)
```

We would need to split into two calls and then pass the file path into the method
https://github.com/drone/drone/blob/master/controller/hook.go#L127

```Go
// fetch the .drone.yml file from the database
raw, err := remote_.Script(user, repo, build, ".drone.yml")
if err != nil {
	log.Errorf("failure to get .drone.yml for %s. %s", repo.FullName, err)
	c.AbortWithError(404, err)
	return
}
sec, err := remote_.File(user, repo, build, ".drone.sec")
if err != nil {
	log.Debuf("cannot find .drone.sec for %s. %s", repo.FullName, err)
	c.AbortWithError(404, err)
	return
}
```

And finally we could set the yaml and secrets files at runtime like this:

```Go
var (
  droneYml = os.Getenv("BUILD_CONFIG_FILE")
  droneSec = os.Getenv("BUILD_SECRET_FILE")
)

func init() {
  if droneYml == "" {
    droneYml = ".drone.yml"
  }
  if droneSec == "" {
    droneSec = ".drone.sec"
  }
}
```

So a slightly different implementation strategy if you are up for the challenge :smile: 

Thanks!

frapposelli:
I like your implementation idea, one thing I would keep from my current implementation though, is the automatic discovery of the secrets file.

I'll take a stab at the implementation today.

frapposelli:
Refactored as requested, I'm setting the yaml file in `hook.go`, also maintained the automatic naming of the secrets file.

bradrydzewski:
awesome, looks really good. I'm going to increment the docker image tag to 0.4.1 tomorrow in master. Once incremented, if you can grab the updated yaml and sec file and squash the commit we should be all set. Thanks for this!

frapposelli:
Sure, will do

bradrydzewski:
ok, just bumped the docker tag to 0.4.1

frapposelli:
alright, squashed and ready for prime time :smile:

PR #1540, opened 2016-03-26
Prevent github repo duplicates

Bugagazavr:
## Problem:

`GetUserRepos(client)` - returns not only repos owned by user, but and repos when user registred as developer, then drone try to get list of organizations, and list of repos belongs to organization.

And we got a duplicates of repositories in drone cache, because we have the same repos in **user_repos** and **organization_repos**

## Solution:

Use map with key `repo.FullName`, to prevent duplicates, and then convert this to array with unique elements

### DIFF
--- a/remote/github/helper.go
+++ b/remote/github/helper.go
@@ -75,6 +75,9 @@ func GetRepo(client *github.Client, owner, repo string) (*github.Repository, err
 // GetAllRepos is a helper function that returns an aggregated list
 // of all user and organization repositories.
 func GetAllRepos(client *github.Client) ([]github.Repository, error) {
+	var filteredRepos []github.Repository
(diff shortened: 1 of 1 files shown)


bradrydzewski:
we should be able to fix with just a one line change:

```diff
// Repos fetches a list of repos from the remote system.
func (g *Github) Repos(u *model.User) ([]*model.RepoLite, error) {
	client := NewClient(g.API, u.Token, g.SkipVerify)

-	all, err := GetAllRepos(client)
-	all, err := GetUserRepos(client)
	if err != nil {
		return nil, err
	}
```

GitHub changed the `/user/repos` endpoint to include organization repositories, which means we no longer need to fetch organization repos separately

PR #1543, opened 2016-03-27
Use gofmt to fmt & simplify

MetalMatze:
Also the credit goes to: https://goreportcard.com/report/github.com/drone/drone

If you want you can add this to the Makefile:
```makefile
fmt:
	gofmt -s -w .
```

I have not added it because it also formats the files in `vendor/`.
Currently I'm not able to exclude those.

### DIFF
--- a/engine/worker.go
+++ b/engine/worker.go
@@ -54,7 +54,7 @@ func (w *worker) Build(name string, stdin []byte, pr bool) (_ int, err error) {
 			Binds: []string{"/var/run/docker.sock:/var/run/docker.sock"},
 		},
 		Volumes: map[string]struct{}{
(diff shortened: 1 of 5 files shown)


engine/worker.go
```diff
@@ -54,7 +54,7 @@ func (w *worker) Build(name string, stdin []byte, pr bool) (_ int, err error) {
 			Binds: []string{"/var/run/docker.sock:/var/run/docker.sock"},
 		},
 		Volumes: map[string]struct{}{
-			"/var/run/docker.sock": struct{}{},
+			"/var/run/docker.sock": {},
```
bradrydzewski: very cool, didn't realize we could do this now :)

Issue #470, opened 2014-09-19
Deploy Pipelines


gregory90:
I'm using github-flow branching model. My ideal CD workflow looks like this :
1. Merge some feature branch to master,
2. Drone picks up change, build starts,
3. If build succeeds Drone is creating artifacts and sends them somewhere (it could be docker image or some .tgz file - that part is pretty much covered by publish plugins),
4. Deploy plugin starts - application is deployed to staging environment,
5. When everything looks good on staging - I'd trigger manually deployment to production.

 So basically what I need in the most simple form is something like this:
```yaml
deploy: 
    ssh: 
        target: user@staging_host:/ 22 
        cmd: /redeploy.sh
    ssh: 
        target: user@production_host:/ 22 
        cmd: /redeploy.sh
        manual: true
```
(Actually I cannot use two ssh plugins in deploy part right now, Drone is picking up only the last one, I'm using latest exp branch).
After successful build and review on staging I'd trigger deployment to production from Drone UI.

I don't have experience with golang, but I'd try to implement this if it's meeting project goals.

ianneub:
@bradrydzewski Is this still "in the pipeline"? I think adding this feature is what I'd need to use Drone as well. At first I thought about writing my own CI system, but maybe I could help with this?

I'm not sure how this fits in though with the changes you are working on for 0.4. Maybe it should wait until 0.4 is stable?

bradrydzewski:
@ianneub yes, I definitely want to tackle pipelines but ideally once 0.4 is more stable. There is just so much changing to support plugins + matrix right now

ltackmann:
+1

zakjan:
Support for Continuous Delivery (GitHub, TBD flow) model is the last missing feature that keeps me from using Drone. Really. I don't want to clutter my git repo with branches for each environment, as seen in Continuous Deployment (Gitflow) model.

Obviously it needs some kind of Deployment Pipelines with manual triggers. As a good example, see Go.CD https://www.go.cd/documentation/user/current/resources/images/PipelineActivity.png

tboerger:
@zakjan if you need a real pipeline or sounds like drone is simply wrong for you. You need different tool like concourse.ci

zakjan:
[...]

PR #1544, opened 2016-03-27
#470 initial pipeline support

jeffastorey:
Changes to drone core to support the pipelines https://github.com/drone/drone/issues/470. Full support requires changes to drone-exec, which I'll committing shortly.

### DIFF
--- a/controller/build.go
+++ b/controller/build.go
@@ -236,8 +236,8 @@ func PostBuild(c *gin.Context) {
 	build.Started = 0
 	build.Finished = 0
 	build.Enqueued = time.Now().UTC().Unix()
-	build.Event := c.DefaultQuery("event", build.Event)deploy_to
-	build.Deploy := c.DefaultQuery("deploy_to", build.Deploy)	
+	build.Event = c.DefaultQuery("event", build.Event)
+	build.Deploy = c.DefaultQuery("deploy_to", build.Deploy)	
 	for _, job := range jobs {
 		job.Status = model.StatusPending
 		job.Started = 0
--- a/controller/build.go
+++ b/controller/build.go
@@ -236,6 +236,8 @@ func PostBuild(c *gin.Context) {
 	build.Started = 0
 	build.Finished = 0
 	build.Enqueued = time.Now().UTC().Unix()
+	build.Event := c.DefaultQuery("event", build.Event)deploy_to
+	build.Deploy := c.DefaultQuery("deploy_to", build.Deploy)	
 	for _, job := range jobs {
 		job.Status = model.StatusPending
 		job.Started = 0


controller/build.go
```diff
@@ -236,6 +236,8 @@ func PostBuild(c *gin.Context) {
 	build.Started = 0
 	build.Finished = 0
 	build.Enqueued = time.Now().UTC().Unix()
+	build.Event = c.DefaultQuery("event", build.Event)
+	build.Deploy = c.DefaultQuery("deploy_to", build.Deploy)	
 	for _, job := range jobs {
```
bradrydzewski: looks good!

we should probably just make sure that `build.Event` is equal to one of these values
https://github.com/drone/drone/blob/master/model/const.go#L3

bradrydzewski: we should probably also only enable overriding `build.Event` and `build.Deploy` if `fork` is true. The reason being when `fork` is false it just re-runs the existing build number and does not create a new build number. We probably want to treat the event end deploy values as immutable unless being forked.

bradrydzewski:
Looks great. I added a few minor notes for validation. I'm sorry I didn't bring them up previously, however, they hadn't crossed my mind until I started the actual code review.

jeffastorey:
Those make sense. In what situations is a build actually forked? Does the restart build button do it or only through API calls?

Related to this, is it possible to access/use the information that's passed as part of the deploy hook? For example I just wanted a single drone deployment configuration whose branch is based on the name passed in the deployment event, is that possible?

bradrydzewski:
> Those make sense. In what situations is a build actually forked? Does the restart build button do it or only through API calls?

The restart button restarts the build and does not fork (assign new build number). The api and command line utility can be used to fork and re-execute the build:

```
drone build fork octocat/hello-world 55
```

The primary reason for this behavior is the restart button in the user interface is intended for quick restarts due to some external failure (ie npm was down while executing the build, so I need to re-run). The fork functionality was created to enable nighty builds from cron jobs, using the cli, where a new build number is desirable.

> Related to this, is it possible to access/use the information that's passed as part of the deploy hook? For example I just wanted a single drone deployment configuration whose branch is based on the name passed in the deployment event, is that possible?

not sure I'm 100% clear. Can you provide some examples or mockup yaml? Which data and how you'd like to use?

jeffastorey:
Sample YAML might be something like this. The idea being that I can use a bucket name that is the same as the environment in this case, rather than creating separate deploy configurations for each of my environments.

```
deploy:
    terraform:
        plan: false
        remote:
            backend: S3
            config:
                bucket: <environment-from-deploy>
                key: tfstate
        when:
            event: deployment
            environment: mytestenvironment
```

jeffastorey:
Also I'm going to close this PR and reopen where I'm not trying to merge in my master branch to avoid the situation where I tried to open 2 PRs on master.

PR #1546, opened 2016-03-28
#470 Pipeline support

jeffastorey:
Reopening the pipeline support PR from https://github.com/drone/drone/pull/1544. I had mistakenly opened 2 PRs off my master branch. @bradrydzewski apologies for the inconvenience on this.

### DIFF
--- a/controller/build.go
+++ b/controller/build.go
@@ -228,11 +228,13 @@ func PostBuild(c *gin.Context) {
 			c.String(500, err.Error())
 			return
 		}
-		if build.Event == model.EventPush ||
-				build.Event == model.EventPull ||
-				build.Event == model.EventTag ||
-				build.Event == model.EventDeploy {
-			build.Event = c.DefaultQuery("event", build.Event)
+
+		event := c.DefaultQuery("event", build.Event)
+		if event == model.EventPush ||
+		    event == model.EventPull ||
+		    event == model.EventTag ||
+		    event == model.EventDeploy {
+		    build.Event = event
 		}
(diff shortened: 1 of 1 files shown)


controller/build.go
```diff
@@ -228,6 +228,13 @@ func PostBuild(c *gin.Context) {
...
 			return
 		}
+		if build.Event == model.EventPush ||
+				build.Event == model.EventPull ||
+				build.Event == model.EventTag ||
+				build.Event == model.EventDeploy {
+			build.Event = c.DefaultQuery("event", build.Event)
+		}
+		build.Deploy = c.DefaultQuery("deploy_to", build.Deploy)
 	}
```
bradrydzewski: I think we want to validate the incoming event (from the query parameter) instead of the existing value. So it would look something like this:

```
var event = c.DefaultQuery("event", build.Event)
if event == model.EventPush ||
    event == model.EventPull ||
    event == model.EventTag ||
    event == model.EventDeploy {
    build.Event = event
}
```

bradrydzewski:
thanks for updating based on feedback! just a minor tweak and we should be all set to merge

controller/build.go
jeffastorey: Doh, you're right. Will fix.

bradrydzewski:
thanks!  I submitted a patch to add a deploy command to the command line utility:

```
drone deploy octocat/hello-world 5 production
```

see  https://github.com/drone/drone-cli/commit/6a10d47f29449401fbac90b04084141b0b8c6478

PR #1548, opened 2016-03-28
Remove install '-t' option

zaro0508:
The script to setup sassc does not work on OSX because the '-t' option
isn't available.  Remove this option so that it's more generic.

### DIFF
--- a/contrib/setup-sassc.sh
+++ b/contrib/setup-sassc.sh
@@ -18,4 +18,4 @@ cd sassc
 make
 
 # isntall the sassc binary
-install -t /usr/local/bin bin/sassc
+install bin/sassc /usr/local/bin


tboerger:
IMHO you don't want to use it on OSX at all because you should prefer homebrew.

gtaylor:
Homebrew is not used or installed on all Macs. We should fix this rather
than make prescriptions on which package manager our users use.

LGTM. We had to yank -t at work, too.

tboerger:
> Homebrew is not used or installed on all Macs. We should fix this rather
> than make prescriptions on which package manager our users use.
> 
> LGTM. We had to yank -t at work, too.

Our general users are not building it on a mac. If the user wants to build it on a mac we can make some prescriptions for sure because those users are the exception.

tboerger:
Beside that officially supported is anyway only the docker version as far as I know.

zaro0508:
I was just following the documentation.  If you don't want to change then maybe you should think about updating your docs to tell people to use a package manager instead? I'm refering to this,  https://github.com/drone/drone#from-source

gtaylor:
@zaro0508 Since the section says "from source" and not "from package manager", I 100% agree. I haven't tried this without -t on Linux, but absolute worst case the shell script could be modified to not use the -t form on Mac OS (assuming yanking the -t causes issues on Linux, which may not be the case at all).

bradrydzewski:
`-t` is required for linux installs and will otherwise error. I'm not opposed to a fix, but at the same time we are trying to remove libsass as a dependency so the useful life of such a script change will be short lived. Feel free to re-open a PR with a posix-compatible fix

contrib/setup-sassc.sh
```diff
@@ -18,4 +18,4 @@ cd sassc
 make
 
 # isntall the sassc binary
```
xiaods: typo: isntall

bradrydzewski:
closing because 0.5 no longer uses sassc and switches to a separate JS front-end using less. You can see the updated UI and repository at github.com/drone/drone-ui

Issue #1550, opened 2016-03-29
send deployment status updates


jeffastorey:
The github remote should support sending status updates for deployments. Discussed a plan with @bradrydzewski on how to implement.

PR #1551, opened 2016-03-29
#1550 deployment status hooks for github

jeffastorey:


remote/github/github.go
```diff
@@ -249,6 +256,21 @@ func (g *Github) Status(u *model.User, r *model.Repo, b *model.Build, link strin
...
+func deploymentStatus(client *github.Client, r *model.Repo, b *model.Build, link string) error {
+	// the deployment ID is only available in the the link to the build as the last element in the URL
+	parts := strings.Split(b.Link,"/")
+	id, _ := strconv.Atoi(parts[len(parts)-1])
+	status := getStatus(b.Status)
+	desc := getDesc(b.Status)
+	data := github.DeploymentStatusRequest{
+		State:			github.String(status),
+		Description: 	github.String(desc),
+		TargetURL:		github.String(link),
```
bradrydzewski: run gofmt :)

remote/github/github.go
```diff
@@ -249,6 +256,21 @@ func (g *Github) Status(u *model.User, r *model.Repo, b *model.Build, link strin
 	return err
 }
 
+func deploymentStatus(client *github.Client, r *model.Repo, b *model.Build, link string) error {
+	// the deployment ID is only available in the the link to the build as the last element in the URL
+	parts := strings.Split(b.Link,"/")
+	id, _ := strconv.Atoi(parts[len(parts)-1])
+	status := getStatus(b.Status)
```
bradrydzewski: we should probably verify the deployment matches `deployment/{number}` since it is possible to trigger a deployment manually without using the GitHub deployment api. When this happens we can just exit gracefully with no error

remote/github/github.go
jeffastorey: Gah, sorry. Tabs/spaces got messed up there. I'm still new to the go ecosystem - forgot to run that.

remote/github/github.go
jeffastorey: Not sure I follow - what would we be matching against `deployment/{number}`?

ericanderson: Reading through the GH API pages, the workflow is what matters.

1. Create GH "Deployment"... Which triggers...
2. A deployment build... Which...
3. Update GH with the status.

The GH deployment has an id thats independent of Drone. When you submit back a status, you need to have this match the deployment.

In the case of a manual trigger, there won't be a GH deployment ID, thus this will fail.

In general I'm guessing this will fail for that reason.

ericanderson: See: https://developer.github.com/guides/delivering-deployments/

bradrydzewski: the deployment event can be triggered by GitHub, however, a deployment event can also be triggered directly using the Drone API without involving the GitHub API.

when a deployment is triggered using the Drone API a deployment status should not be triggered. We can detect this by looking at the Link which won't match the expected pattern.

jeffastorey: In the event that it doesn't match, is it appropriate to just do nothing?

bradrydzewski: yes, when the deployment link does not match we should exit without failure. This will indicate a deployment was triggered by something other than GitHub in which case no deployment status should be sent to GitHub.

```diff
+var re = regexp.MustCompile(".+/deployment/(\\d)")

func deploymentStatus(client *github.Client, r *model.Repo, b *model.Build, link string) error {
+	matches := re.FindStringSubmatch(b.Link)
+	if len(matches) != 2 {
+		return nil
+	}
+	id, _ := strconv.Atoi(matches[1])

```

remote/github/github.go
bradrydzewski: no worries :)

remote/github/github.go
jeffastorey: Yeah that makes sense. It's too bad it looks like the only way to get the deployment ID is from the link, but I don't see a cleaner way to do this right now.

jeffastorey:
updated with the checks. changed the proposed regex to /deployments not /deployment to match the incoming github url. and used gofmt

remote/github/github.go
```diff
@@ -249,6 +259,26 @@ func (g *Github) Status(u *model.User, r *model.Repo, b *model.Build, link strin
 	return err
 }
 
+func deploymentStatus(client *github.Client, r *model.Repo, b *model.Build, link string) error {
+	// the deployment ID is only available in the the link to the build as the last element in the URL
+	parts := strings.Split(b.Link, "/")
+	matches := githubDeployRegex.FindStringSubmatch(b.Link)
```
bradrydzewski: no need for the string split here, the number will be in `matches[1]`

jeffastorey:
fixed

bradrydzewski:
thanks! LGTM

MetalMatze:
I think #1550 can be closed, right?

PR #1556, opened 2016-03-30
relocate api handlers, add swagger annotations

bradrydzewski:
this change starts to add swagger annotations for /api handlers. Eventually we'll be able to auto-generate the swagger file instead of hand writing. woohoo!


tboerger:
I'm excited to see it generated

PR #1560, opened 2016-04-02
Match user url against default api url

iamjarvo:
Some GHE urls can have the github url prefix.
I.E https://github.compaq.com, this caused the GHE code not to run.
User.url is the api url for that that user
GHE urls always have /api/version at the end. Probability of a GHE url
starting with api.github is low


bradrydzewski:
awesome job ... I never thought of `git.compaq.com` would match `github.com`!

PR #1562, opened 2016-04-04
Google Cloud SQL support

### LINKED ISSUE #1373: Unable to run Drone on MySQL using Google Cloud SQL
Google Cloud SQL second generation database (MySQL), does not work with Drone 0.4.

When starting up, you get an error like this (using 'docker logs -f drone'):
time="2015-12-12T08:14:56Z" level=info msg="using database driver mysql" time="2015-12-12T08:14:56Z" level=info msg="using database config drone:REDACTED@tcp(REDACTED:3306)/drone?parseTime=true" time="2015-12-12T08:14:56Z" level=error msg="Error 1071: Specified key was too long; max key length is 767 bytes" time="2015-12-12T08:14:56Z" level=fatal msg="migration failed" 

I believe the reason is that drone has very long VARCHAR entries in the config, specified here:
https://github.com/drone/drone/blob/master/store/migration/mysql/1_init.sql
Some of them are longer than 767.

The Google Cloud SQL database uses the following settings:
| innodb_file_format                       | Barracuda              |
| innodb_large_prefix                      | OFF                    |
It is not possible to get SUPER privileges on Google Cloud SQL, so I can't change those settings (ON for innodb_large_prefix.

Would it be possible to use TEXT or do something else for those long columns?

Some more detail about this problem can be found here:
http://mechanics.flite.com/blog/2014/07/29/using-innodb-large-prefix-to-avoid-error-1071/

Information about Google Cloud SQL (2nd generation) can be found here:
https://cloud.google.com/sql/docs/getting-started

kop:
Fixes #1373, adds support for `innodb_large_prefix=off`.

### DIFF
--- a/store/datastore/ddl/mysql/3.sql
+++ b/store/datastore/ddl/mysql/3.sql
@@ -3,7 +3,7 @@
 CREATE TABLE secrets (
  secret_id        INTEGER PRIMARY KEY AUTO_INCREMENT
 ,secret_repo_id   INTEGER
-,secret_name      VARCHAR(500)
+,secret_name      VARCHAR(255)
 ,secret_value     MEDIUMBLOB
 ,secret_images    VARCHAR(2000)
 ,secret_events    VARCHAR(2000)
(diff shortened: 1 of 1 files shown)


bradrydzewski:
Thanks!

PR #1571, opened 2016-04-13
backport 0.5 queue packages

bradrydzewski:
This pull requests adds `stream`, `bus` and `queue` packages. These are not currently being used in 0.4 but are being added in preparation for 0.5

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,4 +21,9 @@ ENV DATABASE_CONFIG=/var/lib/drone/drone.sqlite
 
 ADD drone_static /drone_static
 
+# Alpine Linux doesn't use pam, which means that there is no /etc/nsswitch.conf,
+# but Go and CGO rely on /etc/nsswitch.conf to check the order of DNS resolving.
+# To fix this we just create /etc/nsswitch.conf and add the following line:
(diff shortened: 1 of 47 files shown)


Dockerfile
```diff
@@ -21,4 +21,9 @@ ENV DATABASE_CONFIG=/var/lib/drone/drone.sqlite
 
 ADD drone_static /drone_static
 
+# Alpine Linux doesn't use pam, which means that there is no /etc/nsswitch.conf,
```
tboerger: But I never had those issues with my alpine base image for my drone image.

bradrydzewski: me neither, but under certain host configurations the Go program will be unable to resolve domain names without this file. This is because we are using cgo which statically links glibc which looks for this file. Alpine doesn't use glibc (it uses musl) and doesn't include this file by default.

https://github.com/gliderlabs/docker-alpine/issues/11
https://github.com/drone/drone/issues/1269

thomasf: Go's net package also parses nsswitch.conf file by itself so it might help even if pam isn't installed.

IIRC the situation that did not work was that the hosts-file wasnt looked at all which docker used to link containers togeher earlier. Since Docker has moved to using an DNS server for the networks the nsswitch.conf might not be required anymore. Idk.

PR #1578, opened 2016-04-17
Handle Bitbucket mercurial hooks

Jyrno42:
https://confluence.atlassian.com/bitbucket/event-payloads-740262817.html#EventPayloads-RepositoryEvents states that only difference between a git and hg `repo:push` is the change type field.

Changed the checks to also work for hg


Jyrno42:
Along with the patch from https://github.com/drone-plugins/drone-hg/pull/8) this worked splendidly.

tboerger:
LGTM

PR #1587, opened 2016-04-21
enables cloning private repos in 0.5

bradrydzewski:


bradrydzewski:
This pull request enables cloning private repositories in 0.5 which was previously unavailable due to lack of credentials. The change sends the netrc credentials to the plugin when the repository is private. This re-uses the new secret implementation which provides added security.

Issue #1386, opened 2015-12-26
Pull queue, agent implementation


bradrydzewski:
### update

This issue will be closed once the communication protocol is finalized. We are currently looking to implement the STOMP protocol with features based on Apache Apollo (http://activemq.apache.org/apollo/documentation/stomp-manual.html). This will provide a standard communication protocol with the potential to easily choose between the embedded drone queue and a compatible STOMP mq server for added scaleability.

* [x] implement stomp protocol for standard communication
* [x] update documentation to reflect protocol change and configuration
* [x] remove existing custom protocol
* [x] remove existing agent token / secret for custom protocol

---

This is a proposal to switch from a push queue in 0.4 to a pull queue in 0.5. The current version of Drone manages incoming build requests, and pushes builds to docker daemons when build slots are available. This doesn't really allow for very fine-grained scheduling, and using the Docker remote API as the transport protocol makes supporting non-Docker builds (like iOS app) much more difficult.

The pull queue would have us place an agent on a server that will poll the main Drone server for builds. The agent would be started on each server by running the Drone server in agent mode:

```
docker run drone/drone --agent --host=http://mydrone.com
```

The Drone server would need an endpoint for polling jobs:

```
POST /api/work/pull
```

And an endpoint for posting job results, including intermediate updates like pending, running, finished, as well as build logs:

```
POST /api/work
```

Below discusses some pros and cons to the current approach vs. proposed new approach.

### Durability

The Drone server can be taken down for periodic upgrades without worrying about losing the in-memory build queue (see https://github.com/drone/drone/issues/1195). Agents can use backoff to retry contacting Drone server if unavailable when posting results or requesting jobs.

### Multi-architecture

This makes it easier to support multiple operating system and architectures, especially those that don't support Docker (I'm looking at you, iOS). A build system that doesn't support iOS applications can never be a true alternative to Jenkins.

### Resource Usage

This allows (in a future release) the agent to regulate how many builds can execute based on resource usage, instead of just a fixed number (ie 2 builds on this server). This is similar to how Docker swarm has different profiles for allocating resources, and we could (eventually) provide something similar.

### Elasticity

The current design makes increasing or decreasing capacity relatively simple, however, I think agents make this even easier. You could simply snapshot an image or AMI and start up a bunch of servers in batch, which would immediately start processing jobs when they come online. This could still be used in conjunction with Docker machine to bring up the server and run an agent. It would also work better with something like Rancher and Mesos, which have their own GUIs for launching servers.

[...]

PR #1590, opened 2016-04-21
Add health check to drone-workers

ogavrisevs:
This PR adds health check to test can `drone-workers` be reached (is there any point to send build job). 
This is useful in distributed setup when `drone-master` / `drone-workers  are on different host. 
If `drone-worker` is terminated / removed from auto scaling group due to low CPU load you dont need to notify `drone-master` directly (with drone-cli / drone rest api) because `drone-master` will remove `drone-worker` from node list after three connection retries.


bradrydzewski:
@ogavrisevs we are in the process of changing the approach to an agent-based pull queue that would obsolete this pull request https://github.com/drone/drone/issues/1386

I really appreciate the pull request, however, this isn't something we would be able to merge since we are taking things in a different direction. I recommend using the [gitter](https://gitter.im/drone/drone) in the future to discuss changes prior to implementation so we can all get on the same page. Cheers!

PR #1591, opened 2016-04-22
use new 0.5 .drone.sig signature file

bradrydzewski:


bradrydzewski:
this change enables the following 0.5 functionality:

* new signature file to replace drone.sec
* secrets injected based on signature verification
* agent endpoints now require authorization

PR #1592, opened 2016-04-22
re-attempt pushing updates from agent on failure for 0.5

bradrydzewski:


bradrydzewski:
this change uses exponential backoff when the agent attempts to send updates to the build server. If the build server is unavilable waits and retries. This should hopefully make the agents more resilient to any sort of networking issues, drone server upgrades while builds are running, etc

PR #1593, opened 2016-04-23
implemented 0.5 in-memory stream multiplexer

bradrydzewski:


bradrydzewski:
this implements code for an in-memory stream multiplexer. It was a bit tedious to write and with all the `sync` code has potential for race conditions if I screwed up, so it needs to be heavily tested with the [go race detector](https://golang.org/doc/articles/race_detector.html).

tboerger:
Cool

PR #1595, opened 2016-04-25
Add user validation to the API

metalmatze:
Right now a user is able to
- delete itself
- create users with empty login name

This PR fixes those cases and returns a `422` status code.
See https://httpstatuses.com/422 to why `422`.

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -22,7 +22,7 @@ import (
 func GetUsers(c *gin.Context) {
 	users, err := store.GetUserList(c)
 	if err != nil {
-		c.String(http.StatusInternalServerError, fmt.Errorf("Error getting user list. %s", err))
+		c.String(http.StatusInternalServerError, fmt.Sprintf("Error getting user list. %s", err))
 		return
 	}
 
@@ -81,7 +81,7 @@ func PostUser(c *gin.Context) {
 	in := &model.User{}
 	err := c.Bind(in)
 	if err != nil {
-		c.String(http.StatusBadRequest, err)
+		c.String(http.StatusBadRequest, err.Error())
 		return
 	}
 
@@ -99,7 +99,7 @@ func PostUser(c *gin.Context) {
 
 	err = store.CreateUser(c, user)
 	if err != nil {
-		c.String(http.StatusInternalServerError, err)
+		c.String(http.StatusInternalServerError, err.Error())
 		return
 	}
 


api/users.go
```diff
@@ -55,6 +59,13 @@ func PatchUser(c *gin.Context) {
 		c.AbortWithStatus(http.StatusNotFound)
 		return
 	}
+
+	// A user can't deactivate or remove it's admin rights by itself.
+	if session.User(c).ID == user.ID {
+		c.AbortWithStatus(422)
```
tboerger: I would also use `AbortWithError` like you did it below including a message.

api/users.go
```diff
@@ -75,6 +86,10 @@ func PostUser(c *gin.Context) {
 		return
 	}
 
+	if in.Login == "" {
+		c.AbortWithError(422, errors.New("User's login can't be empty"))
```
tboerger: Generally I'm mostly using `fmt.Errorf` instead of `errors.New`, but that is more like a personal taste.

api/users.go
```diff
@@ -102,12 +117,20 @@ func PostUser(c *gin.Context) {
...
 	user, err := store.GetUserLogin(c, c.Param("login"))
 	if err != nil {
-		c.String(404, "Cannot find user. %s", err)
+		c.String(http.StatusNotFound, "Cannot find user. %s", err)
+		return
+	}
+
+	// User can't delete itself.
+	if user.ID == session.User(c).ID {
+		c.AbortWithStatus(422)
```
tboerger: Here I would also suggest `AbortWithError`

api/users.go
MetalMatze: Ok, I didn't knew there is a fmt that returns an error.

api/users.go
```diff
@@ -36,10 +39,11 @@ func GetUsers(c *gin.Context) {
 func GetUser(c *gin.Context) {
 	user, err := store.GetUserLogin(c, c.Param("login"))
 	if err != nil {
-		c.String(404, "Cannot find user. %s", err)
-	} else {
-		c.JSON(200, user)
+		c.AbortWithError(http.StatusNotFound, fmt.Errorf("Cannot find user. %s", err))
```
bradrydzewski: I'm actually using `c.String(404, "Cannot find user. %s", err)` for new code

tboerger: Should we maybe always return proper JSON responses?

MetalMatze: @tboerger That would be better. I propose something like:
```json
{
  "errors": [{
    "status": "400",
    "detail": "JSON parse error - Expecting property name at line 1 column 2 (char 1)."
  }]
}
```
An array of errors makes sense for validation.

tboerger: And that should be doable with a custom recovery middleware

bradrydzewski: I'm not sure I agree. A structured JSON response that only has the status code and detail is effectively the same as `c.String(404, "Cannot find user. %s", err)`

I am open to further discussion on the matter, however, from a practical perspective we cannot make this sort of change without modifying the CLI and API as well which currently expect text messages.

tboerger: I personally expect a json response from a json api and not a text reponse, even when it's an error :)

api/users.go
```diff
@@ -55,6 +59,12 @@ func PatchUser(c *gin.Context) {
 		c.AbortWithStatus(http.StatusNotFound)
 		return
 	}
+
+	if session.User(c).ID == user.ID {
+		c.String(422, "Users can't deactivate or change admin rights themselfs")
+		return
```
bradrydzewski: with 0.5 user administrators are declared via environment variable so this no longer applies

api/users.go
```diff
@@ -75,6 +85,10 @@ func PostUser(c *gin.Context) {
 		return
 	}
 
+	if in.Login == "" {
+		c.String(422, "User's login can't be empty")
```
bradrydzewski: this needs a `return`

PR #1598, opened 2016-04-27
Replace configuraiton by configuration in router/middleware/remote.go

bbenoist:


bradrydzewski:
thanks! :smile:

bbenoist:
Looks like I have no way to cla :cry: Even when opening a new PR...

PR #1599, opened 2016-04-27
add matrix include capability

bradrydzewski:
cc @sivel

### DIFF
--- a/drone/agent/exec.go
+++ b/drone/agent/exec.go
@@ -83,6 +83,11 @@ func (r *pipeline) run() error {
 		})
 	}
 
+	var lastStatus string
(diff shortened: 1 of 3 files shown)


sivel:
That was unexpectedly fast!  I'll try to test in the next few days.

bradrydzewski:
yep, after you asked I realized it was simple to implement, and figured I might as well just add it since I was already thinking about it :smile:

jbakosi:
Thanks guys, @sivel, @bradrydzewski. I was thinking about the same but have not had the time. Do you guys think it will work together with exclusions in #1514?

PR #1601, opened 2016-04-29
When running with CANARY, avoid 'string too long', print full agreement

sivel:
It seems that golang will print `[string too long]` when using `panic`/`println`/etc... if the string is really long.

When running with `CANARY=true`, and without accepting the agreement, we instead of getting the agreement just see `[string too long]`.  Based on https://github.com/golang/go/issues/8339 this is fixed in some places, but apparently not all, however the easy solution is to use `fmt`.

### DIFF
--- a/drone/server/server.go
+++ b/drone/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"fmt"
 	"net/http"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


bradrydzewski:
I ended up resolving this locally but thanks for letting me know!

PR #1602, opened 2016-04-29
Override ansi-to-hml default background color

cdaguerre:
Simple patch for nicer output in console.
Obviously you could configure the ANSI-to-HTML package [see here](https://github.com/rburns/ansi-to-html/blob/master/src/ansi_to_html.coffee#L72) to do this.

capture d ecran 2016-04-29 a 19 24 43


MetalMatze:
I think this patch could be directly forwarded to [drone/drone-ui](https://github.com/drone/drone-ui).

cdaguerre:
This patch sucks anyway, I'll try to come up with something better ;)

Issue #1610, opened 2016-05-03
Travis-CI like line numbering and folding


cdaguerre:
Hi, my builds being several thousand lines long, I needed travis-ci like line folding so I made some changes to the web ui in my fork.
I would have to clean it up a little to make a PR. Is this something you would like to get into core?
Screencast attached for reference.
I removed the current ansi-to-html script and used [ansi_up](https://github.com/drudru/ansi_up) instead, allowing terminal output colors to be customized by css classes rather than inlined styles.
![folding](https://cloud.githubusercontent.com/assets/4642448/14987568/60952f0a-1150-11e6-8eb2-c303f8ee1a5b.gif)

MetalMatze:
Sorry to say this to you. But in the 0.5 release there has been a significant change in the output of logs. It now returns json, not just plain text for each line.
I'm currently building a react.js based frontend which will eventually be merged to this repo.
If you don't mind, I'm very much interested on how you implemented this. :blush:

bradrydzewski:
This is awesome and I do want something like this in drone :smile: 

As @MetalMatze mentioned we are in the process of working on the 0.5 release which how now the master branch. The logging code in master now sends structured logging output in JSON format to the web browser.

This is the JSON format:

```Go
type Line struct {
   Proc string // process name
   Time int // line was output N seconds after the container started
   Pos int // line number for the logs. order is not guaranteed
   Out string // line text
}
```

Note that commands are also currently being wrapped in xml tags although we can potentially alter if necessary. The benefit is that some commands are multiple lines which is why the previously used dollar sign was not idea.

```
<command>go build</command>
```

This is a sample JSON output for a full build:

```json
{"proc":"clone","pos":0,"out":"\u003ccommand\u003egit config --global http.sslVerify false\u003c/command\u003e"}
{"proc":"clone","pos":1,"out":"\u003ccommand\u003egit init\u003c/command\u003e"}
{"proc":"clone","pos":2,"out":"Initialized empty Git repository in /go/src/github.com/garyburd/redigo/.git/"}
{"proc":"clone","pos":3,"out":"\u003ccommand\u003egit remote add origin http://gogs.drone.io/bradrydzewski/redigo.git\u003c/command\u003e"}
{"proc":"clone","pos":4,"out":"\u003ccommand\u003egit fetch --no-tags origin +refs/heads/master:\u003c/command\u003e"}
{"proc":"clone","pos":5,"out":"From http://gogs.drone.io/bradrydzewski/redigo"}
[...]

PR #1611, opened 2016-05-03
[WIP][POC] Travis-CI like line numbering and folding

cdaguerre:
Please refer #1610.


cdaguerre:
The build fails because I'm using my own drone instance and secrets, but I'd fix this if you have interest in this PR.

tboerger:
You should write SASS instead of CSS and integrate it into the single generated CSS.

bradrydzewski:
wondering if you confirmed `ansi_up` is able to stream output? the reason I ask is because to properly display terminal output spanning multiple lines some state needs to be maintained, since the begenning and end of an ansi termination sequence may not appear in the same chunk

bradrydzewski:
We've moved the user interface code to a separate project. It is also now a simple-page application instead of server-side rendered + jquery. If you are interested in re-working / re-submitting to the official ui repository we can get it merged. https://github.com/drone/drone-ui

PR #1615, opened 2016-05-04
Migrate yaml to 0.5

bradrydzewski:


.drone.yml
```diff
@@ -1,24 +1,36 @@
...
+  base: /drone
+  path: src/github.com/drone/drone
 
-build:
+script:
   test:
     image: drone/golang:1.5
     environment:
       - GO15VENDOREXPERIMENT=1
+      - GOPATH=/drone
```
tboerger: Do you want to keep that to be forced to set the GOPATH and PATH manually? :(

bradrydzewski: this is just a temporary workaround until we can use the official Go image (by getting rid of sassc). The official Go image already has set values for GOPATH and GOBIN. With 0.5 we can instruct drone to use `/go` as the base directory instead of `/drone`:

```
workspace:
  base: /go

script:
  test:
    image: golang:1.6
```

PR #1621, opened 2016-05-06
migrate almost all CLI commands to core

bradrydzewski:


bradrydzewski:
this pull request moves all CLI commands to the core binary. The only exception at the moment are `drone compile` (new) and `drone exec` which require a bit of refactoring so we can ideally share the same code with the agents running builds.

tboerger:
:dancers:

PR #1622, opened 2016-05-08
fix(typo): sqite3 -> sqlite3

compressed:
I believe this is a typo... Trying to build from master and use sqlite3 db.


tboerger:
LGTM

bradrydzewski:
great catch!!

PR #1625, opened 2016-05-11
refactored build runner, enables drone exec command

bradrydzewski:
This pull request enables `drone exec` with (as far as I can tell) the ability to fully replicate a build locally from the command line. This required a lot of refactoring for the agent and build runner to share the same code.

This also adds a new field to capture and display error messages. This is useful when a build errors and is not executed (and has empty logs) such as when Yaml validation fails.


gtaylor:
That's a lot of deletions.

bradrydzewski:
@gtaylor ha, yes it is!  I realized I __completely__ over-engineered the initial build runner, parser and yaml compiler. I was able to eliminate probably 80% of the original code.

PR #1630, opened 2016-05-11
quick hack to group log output by step

bradrydzewski:


bradrydzewski:
This is a quick hack to group log output by step. It only works once the build is completed. Streaming logs still render as JSON

PR #1633, opened 2016-05-11
added scopes to the oauth2 config

jdamick:
Now the request after redirect to github now contains scope:
https://github.<enterprise location>/login/oauth/authorize?client_id=...etc...&scope=%5Brepo+repo%3Astatus+user%3Aemail+read%3Aorg+gist+repo_deployment%5D&state=drone

I see there aren't tests yet for this, so i'm not sure you all plan to validate these going forward...?

### DIFF
--- a/remote/github/github.go
+++ b/remote/github/github.go
@@ -53,7 +53,7 @@ func New(opts Opts) (remote.Remote, error) {
 		URL:         defaultURL,
 		Client:      opts.Client,
 		Secret:      opts.Secret,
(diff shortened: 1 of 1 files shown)


tboerger:
And why do you want to add the scopes?

bradrydzewski:
@jdamick thanks, good catch!

@tboerger this is a regression from when I upgraded to use the official go oauth2 provider and forgot to include the scopes. This is required to give drone access to private repos, ability to update statuses, etc.

PR #1634, opened 2016-05-11
ability to get team list from Gogs

bradrydzewski:


bradrydzewski:
this allows us to use `DRONE_TEAMS` in 0.5 with Gogs (in addition to GitHub, Bitbucket)

PR #1635, opened 2016-05-11
ability for agent to ping central drone server

bradrydzewski:


bradrydzewski:
this commit gives agents the ability to ping the server every 5 minutes (configurable) to let the server know the agent is healthy. The list of agents and last ping time can be queried at the `/api/agents` endpoint. This is a WIP with more to come

PR #1638, opened 2016-05-12
Fix Makefile build on OS X

makhmutov:
The ld `-static` linking in not supported on OS X. And it fails with error:
```
/usr/local/Cellar/go/1.6.1/libexec/pkg/tool/darwin_amd64/link: running clang failed: exit status 1
ld: library not found for -lcrt0.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,12 @@
-.PHONY: vendor docs
+.PHONY: vendor docs build
 
 PACKAGES = $(shell go list ./... | grep -v /vendor/)
+
+ifneq ($(shell uname), Darwin)
+	EXTLDFLAGS = -extldflags "-static" #
+else
+	EXTLDFLAGS =
+endif
 
 all: gen build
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,10 +30,8 @@ gen_template:
 gen_migrations:
 	go generate github.com/drone/drone/store/datastore/ddl
 
-build: build_static
-
-build_static:
-	cd drone && go build --ldflags '-extldflags "-static" -X github.com/drone/drone/version.VersionDev=$(DRONE_BUILD_NUMBER)' -o drone
+build:
+	cd drone && go build --ldflags '${EXTLDFLAGS}-X github.com/drone/drone/version.VersionDev=$(DRONE_BUILD_NUMBER)' -o drone
```
tboerger: Missing a space?

Makefile
```diff
@@ -1,7 +1,13 @@
-.PHONY: vendor docs
+.PHONY: vendor docs build
 
 PACKAGES = $(shell go list ./... | grep -v /vendor/)
 
+ifneq ($(shell uname), Darwin)
+	EXTLDFLAGS = -extldflags "-static" #
```
tboerger: Why the # at the end?

Makefile
makhmutov: ${EXTLDFLAGS} contains whitespace in the end

Makefile
makhmutov: Some IDE, editors and tools may remove whitespace in the end. It is one of the solution how to keep spaces.

tboerger:
I would prefer a simple whitespace instead of the null workaround. Even with the whitespace at the beginning it should still work as expected

bradrydzewski:
awesome, thanks!

PR #1646, opened 2016-05-16
fix a typo on Makefile.

xiaods:


tboerger:
That's not a typo ;)

xiaods:
@tboerger could you give me some hints? not idea on the convention.

bradrydzewski:
closing since this functionality was removed from the makefile

but for reference starting the line with `-dpkg` instead of `dpkg` prevented the makefile from failing when a non-zero exit code was returned. For some reason dpkg always returned a non-zero exit code, even though it was working file, which kept causing the build to fail

PR #1654, opened 2016-05-23
agents stream logs using websockets

bradrydzewski:


bradrydzewski:
this pull request makes two important changes. First, the agent now streams the logs using websockets instead of a long-lived http request. Second, the agent posts the full logs at the end of the build in case the websocket connection is broken.

PR #1660, opened 2016-05-27
[github] Add modifiable context for Github status API

frapposelli:
Adds a flag (`DRONE_GITHUB_CONTEXT`) to modify the context used when reporting status check to Github, default is still `continuous-integration/drone`.


bradrydzewski:
LGTM

PR #1661, opened 2016-05-30
Fix undesirable behaviour in drone cli

MirrorZ:
Using drone cli to the view info for or delete a repository that does not exist leads to undesirable behaviour.
Instead of just a simple error message (404 Not found), the contents of the page are dumped in response.
This patch tries to fix the issue for the cli by inspecting the response for the error code and print its corresponding http status text. 
Additionaly this patch changes Cli invocation for adding the repo RepoRemove to RepoAdd.


drone/repo_info.go
```diff
@@ -39,7 +43,14 @@ func repoInfo(c *cli.Context) error {
...
 	if err != nil {
-		return err
+		re := regexp.MustCompile("client error \\d{3}")
+		err_msg := re.FindAllString(err.Error(), -1)
+		if err_msg != nil {
+			r := regexp.MustCompile("\\d{3}")
+			status_code, _ := strconv.Atoi((r.FindAllString(err_msg[0], -1))[0])
+			status_text := http.StatusText(status_code)
+			return fmt.Errorf("%s: %s", err_msg[0], status_text)
+		}
```
bradrydzewski: This can be removed. We are migrating to a single page application and therefore will no longer be returning html error pages, and will instead return a 403 error with an error message.

drone/repo_rm.go
```diff
@@ -30,8 +33,16 @@ func repoRemove(c *cli.Context) error {
...
-		return err
+		re := regexp.MustCompile("client error \\d{3}")
+		err_msg := re.FindAllString(err.Error(), -1)
+		if err_msg != nil {
+			r := regexp.MustCompile("\\d{3}")
+			status_code, _ := strconv.Atoi((r.FindAllString(err_msg[0], -1))[0])
+			status_text := http.StatusText(status_code)
+			return fmt.Errorf("%s: %s", err_msg[0], status_text)
+		}
 	}
```
bradrydzewski: This can be removed. We are migrating to a single page application and therefore will no longer be returning html error pages, and will instead return a 403 error with an error message.

MirrorZ:
Thanks, opened #1666 with just the first change.

PR #1668, opened 2016-06-10
Add `--secrets-file` option to exec command

frapposelli:
This adds a flag to `drone exec` to pass a yaml file with plaintext secrets, file follows this convention:

```yaml
secrets:
  SLACK_URL: https://hooks.slack.com/services/whatever/floats/yourboat
  AWS_ACCESS_KEY_ID: loremipsum
  AWS_SECRET_ACCESS_KEY: blahblahblahblah
  AWS_DEFAULT_REGION: us-east-1
```

I'm introducing a new type directly in the exec file, not sure if it warrants its own type in the drone/yaml package

### DIFF
--- a/drone/exec.go
+++ b/drone/exec.go
@@ -15,6 +15,7 @@ import (
 	"github.com/drone/drone/queue"
 
 	"github.com/codegangsta/cli"
+	"github.com/joho/godotenv"
 )
 
 var execCmd = cli.Command{
@@ -40,6 +41,11 @@ var execCmd = cli.Command{
 			Name:   "secret",
 			Usage:  "build secrets in KEY=VALUE format",
 			EnvVar: "DRONE_SECRET",
+		},
+		cli.StringFlag{
+			Name:   "secrets-file",
+			Usage:  "build secrets file in KEY=VALUE format",
+			EnvVar: "DRONE_SECRETS_FILE",
 		},
 		cli.StringSliceFlag{
 			Name:   "matrix",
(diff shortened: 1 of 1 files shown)


bradrydzewski:
It definitely makes sense to give the ability to load from file.

I'm using the following library in all the plugins to load environment variables:
https://github.com/joho/godotenv

Which allows the following:

```
envs, err := godotenv.Read(filename)
for key, val := range envs {

}
```

I don't have strong opinions on an environment file vs yaml file, but I figured I throw it out there as an option @frapposelli do you have a strong opinion either way? cc @benschumacher since I know you were loading secrets from files in 0.4. Any strong preference?

frapposelli:
@bradrydzewski I've just tried to port the same experience I had with the 0.4 client, in my workflow I was loading secrets directly from the YAML plaintext file used to build `.drone.sec` with `drone exec -E <filename>`.

I wouldn't mind using an environment file, but I like the fact that I don't have to update all the YAML files I've used on 0.4.

gtaylor:
FWIW, `docker run` has a `--env-file` param that we refer to in the install docs. I could see copying that over and continuing the pattern as being consistent, but this is not something I feel strongly about. Figured I'd mention the `--env-file` thing in passing.

bradrydzewski:
> I wouldn't mind using an environment file, but I like the fact that I don't have to update all the YAML files I've used on 0.4.

I definitely can sympathize here since it is a pain updating stuff, but at the same time, I'm not sure we should carry the legacy 0.4 yaml format into 0.5 if it can be avoided. The more I think about it, the more I'm in favor of using the environment file.

I should point out that the CLI will automatically look for a `.env` file in the working directory. So you can already do something like this (just make sure to add `.env` to your `.gitignore`)

```
$ cat .env

DRONE_SECRET=foo=bar,baz=boo
```

For sourcing secrets from a central file I would therefore recommend we alter the `--secret-file`  to use the environment file format instead:

```
envs, err := godotenv.Read(filename)
for key, val := range envs {

}
```

frapposelli:
@bradrydzewski sounds good, I'll update the PR to reflect that.

frapposelli:
Just updated the PR and realized that `godotenv` actually support YAML-style environment file, I've tested the code change with this file format:

```yaml
SLACK_URL: https://hooks.slack.com/services/whatever/floats/yourboat
AWS_ACCESS_KEY_ID: loremipsum
AWS_SECRET_ACCESS_KEY: blahblahblahblah
AWS_DEFAULT_REGION: us-east-1
```

and worked flawlessly.

gtaylor:
> I wouldn't mind using an environment file, but I like the fact that I don't have to update all the YAML files I've used on 0.4.

Yeah, I'm just not sure that we should be making design decisions based on not having to update old files.

Also, is there any need to separate environment vars from secrets during the build process? `--env-file` would match existing Docker conventions and would set any environment variables, secrets or not.

frapposelli:
@gtaylor I would keep the secrets separate from generic environment variables, the `--secrets-file` option loads the file content into a `model.Secret` struct, which mimics exactly how the build is run on drone server.

bradrydzewski:
thanks!

frapposelli:
🎉 🎉 🎉 🎉 🎉 🎉

PR #1670, opened 2016-06-11
update to initialize the Environment map

josmo:
@bradrydzewski I'm not sure if this is the right place to initialize but the agent crashes on builds without it being initialized.  Let me know if you think there's a better place for it.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -210,9 +210,6 @@ func (a *Agent) exec(spec *yaml.Config, payload *queue.Work, cancel <-chan bool)
 			}
 			// updates the build status passed into each container. I realize this is
 			// a bit out of place and will work to resolve.
-		        if pipeline.Head().Environment == nil {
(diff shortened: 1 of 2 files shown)


bradrydzewski:
I'm wondering if you can provide more information under which circumstances the nil environment map is present. We have a transform function that automatically sets the environment variable field if nil:
https://github.com/drone/drone/blob/master/yaml/transform/environ.go#L20

This transform function is invoked by the agent prior to running the build:
https://github.com/drone/drone/blob/master/agent/agent.go#L141

So I'm wondering if this is related to the service container? Does this panic if a service container doesn't have an environment section? If so, I think we should correct this in the transform function.

josmo:
@bradrydzewski yet you were totally right.  I made the change and the transformation now works correctly

bradrydzewski:
excellent, great catch :)

PR #1671, opened 2016-06-12
Netrc stash fix

josmo:
Looks like I didn't clean up the clone link to remove the user id that activated the repo.
I also tried to take care of a few comments from before while I was at it.

### DIFF
--- a/remote/bitbucketserver/bitbucketserver.go
+++ b/remote/bitbucketserver/bitbucketserver.go
@@ -154,9 +154,9 @@ func (c *client) Repo(u *model.User, owner, name string) (*model.Repo, error) {
 
 	client := NewClientWithToken(&c.Consumer, u.Token)
 
-	url := fmt.Sprintf("%s/rest/api/1.0/projects/%s/repos/%s", c.URL, owner, name)
+	urlString := fmt.Sprintf("%s/rest/api/1.0/projects/%s/repos/%s", c.URL, owner, name)
 
-	response, err := client.Get(url)
+	response, err := client.Get(urlString)
 	if err != nil {
 		log.Error(err)
 	}
@@ -178,13 +178,12 @@ func (c *client) Repo(u *model.User, owner, name string) (*model.Repo, error) {
 
 	for _, item := range bsRepo.Links.Clone {
 		if item.Name == "http" {
-			//TODO sdhould find a clean way to do this
-			//We are removing the username out of the link to allow for Netrc to work
-			splitUrl := strings.SplitAfterN(item.Href,"@",2)
-			splitProtocal := strings.SplitAfterN(splitUrl[0],"//",2)
-			cleanUrl := fmt.Sprintf("%s%s",splitProtocal[0], splitUrl[1])
-
-			repo.Clone = cleanUrl
+			uri, err := url.Parse(item.Href)
+			if err != nil {
+				return err
+			}
+			uri.User = nil
+			repo.Clone = uri.String()
 		}
 	}
 	for _, item := range bsRepo.Links.Self {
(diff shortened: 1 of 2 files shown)


remote/bitbucketserver/bitbucketserver.go
```diff
@@ -240,12 +263,16 @@ func (*client) Status(*model.User, *model.Repo, *model.Build, string) error {
 }
 
 func (c *client) Netrc(user *model.User, r *model.Repo) (*model.Netrc, error) {
-	u, err := url.Parse(c.URL) // TODO strip port from url
+	u, err := url.Parse(c.URL)
+	//remove the port
```
bradrydzewski: if there is an error here, calling `u.Host` below will result in a nil pointer and panic

remote/bitbucketserver/bitbucketserver.go
```diff
@@ -152,20 +161,28 @@ func (c *client) Repo(u *model.User, owner, name string) (*model.Repo, error) {
...
+		IsPrivate: true, // TODO(josmo) possibly set this as a setting - must always be private to use netrc
 		FullName:  fmt.Sprintf("%s/%s", bsRepo.Project.Key, bsRepo.Slug),
 	}
 
 	for _, item := range bsRepo.Links.Clone {
 		if item.Name == "http" {
-			repo.Clone = item.Href
+			//TODO sdhould find a clean way to do this
+			//We are removing the username out of the link to allow for Netrc to work
+			splitUrl := strings.SplitAfterN(item.Href,"@",2)
```
bradrydzewski: what about doing a url.Parse, setting the username to `""`, and then converting back to a string?

remote/bitbucketserver/bitbucketserver.go
```diff
@@ -115,16 +116,24 @@ func (c *client) Login(res http.ResponseWriter, req *http.Request) (*model.User,
 
 	// TODO errors should never be ignored like this
 	response1, err := client.Get(fmt.Sprintf("%s/rest/api/1.0/users/%s", c.URL, login))
+	if err != nil {
+		return nil, err
+	}
 	contents, err := ioutil.ReadAll(response1.Body)
+	if err !=nil {
+		return nil, err
```
bradrydzewski: if we get an error here, the response won't be closed and will result in open file descriptors. We should add the defer statement right after the response block, before attempting to read

bradrydzewski:
It would be great to clean up this package a bit. Would you consider taking some time and looking at the structure and coding style of the Gogs and Bitbucket packages and apply to this package? Including tests? :)

remote/bitbucketserver/bitbucketserver.go
josmo: Cool will fix

remote/bitbucketserver/bitbucketserver.go
josmo: That was my first thought.  I just haven't used the golang libs enough but I'll take a look tomorrow at switching it that way.  It would be cleaner.

remote/bitbucketserver/bitbucketserver.go
josmo: Thanks fix in the next push

josmo:
Totally agree on cleaning it up.  Sadly to say I'm still learning golang so I'll keep at it slowly :)

bradrydzewski:
> Totally agree on cleaning it up. Sadly to say I'm still learning golang (this is still the only time I've had to use it) so I'll keep at it slowly :)

Yes, completely understand and appreciate all the effort!

remote/bitbucketserver/bitbucketserver.go
bradrydzewski: you should be able to do something like this:

```
uri, err := url.Parse(item.Href)
if err != nil {
    return err
}
uri.User = nil
repo.Clone = uri.String()
```

bradrydzewski:
LGTM!

PR #1674, opened 2016-06-14
Stash security cleanup

josmo:
First take at implementing Permissions from stash.
Started to pull out some of the functions to resemble a bit more Bitbucket Cloud

@bradrydzewski I've tested a few scenarios but I'll keep trying to see if I can find any issues.  Next stop I'll start matching the Bitbucket Cloud remote as well as being more golangish.  I'll also start looking in how tests work and add those as well.


bradrydzewski:
LGTM

PR #1677, opened 2016-06-14
ability to flush user repo cache

bradrydzewski:


bradrydzewski:
This pull request gives us the ability to flush the user repository cache. This will finally resolve the annoying issue that arises when you create a new repository in GitHub, but then it takes 30 minutes for it to appear in Drone.

PR #1679, opened 2016-06-15
endpoint to get repository list + latest build

bradrydzewski:


bradrydzewski:
this pull request modifies the `/api/user/feed` to include `/api/user/feed?latest=true` to include only the latest build for each repository.

gtaylor:
This is great. We'll put this to good use.

PR #1680, opened 2016-06-16
Slightly better formatted text during the build. 

josmo:
Should look at the function that's used on refresh to use formatting with group


josmo:
I'm closing this PR since #1682 has a better fix till the new UI comes around

PR #1682, opened 2016-06-18
Fix log output

vespakoen:
This also fixes the log output when looking at an old build.


vespakoen:
related: https://github.com/drone/drone/pull/1680

This pull adds some other fixes that the related pull doesn't have.

tboerger:
Can we merge this to get the log output of beta.drone.io fixed? :)

ZombieHippie:
I'm using this on my server https://drone.dryclean.io/ZombieHippie/angular-2-webpack-starter/3

Logging still isn't perfect but at least it shows up.

bradrydzewski:
we have a completely new UI and won't merge this. I've kept it open so that individuals running 0.5 can patch their local versions while waiting for the new UI

bradrydzewski:
closing now that we've officially switched to drone-ui repository

PR #1685, opened 2016-06-20
Fix missing space in authorization failure message

bbayles:
I noticed that there is a missing space in some authorization errors due to line continuation:
missing-space

This PR adds the spaces.


bbayles:
Seems that this file is no longer in master, closing issue.

tboerger:
Yeah this have not been merged because of the new UI at drone/drone-ui.

PR #1693, opened 2016-06-27
Integrated initial command to list secrets

tboerger:
Since we are not able to list the already set secrets I have added the
required API andpoint and the required sub command to list them.

### DIFF
--- a/drone/secret_list.go
+++ b/drone/secret_list.go
@@ -1,1 +1,87 @@
 package main
+
+import (
+	"log"
+	"os"
+	"strings"
+	"text/template"
+
+	"github.com/codegangsta/cli"
+)
+
+var secretListCmd = cli.Command{
+	Name:  "ls",
+	Usage: "list all secrets",
+	Action: func(c *cli.Context) {
+		if err := secretList(c); err != nil {
+			log.Fatalln(err)
+		}
+	},
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "format",
+			Usage: "format output",
+			Value: tmplSecretList,
+		},
+		cli.StringFlag{
+			Name:  "image",
+			Usage: "filter by image",
+		},
+		cli.StringFlag{
+			Name:  "event",
+			Usage: "filter by event",
+		},
+	},
+}
+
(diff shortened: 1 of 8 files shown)


drone/secret_list.go
```diff
@@ -1 +1,77 @@
...
+		}
+
+		tmpl.Execute(os.Stdout, secret)
+	}
+
+	return nil
+}
+
+// template for secret list items
+var tmplSecretList = `{{ .Name }}`
```
bradrydzewski: maybe also display the `--image` filters and `--event` list?

tboerger: How do you want to format it? Suggestions are welcome :)

server/secret.go
```diff
@@ -1,32 +1,46 @@
...
 	"github.com/drone/drone/router/middleware/session"
 	"github.com/drone/drone/store"
 
 	"github.com/gin-gonic/gin"
 )
 
+func GetSecrets(c *gin.Context) {
+	repo := session.Repo(c)
+	secrets, err := store.GetSecretList(c, repo)
+
```
bradrydzewski: let's convert secrets to the below type before we write to the response. This way we can prevent the plaintext password from being returned in the response.

```
type secretResp struct {
	ID int64 `json:"id"`
	Name string `json:"name"`
	Images []string `json:"image,omitempty"`
	Events []string `json:"event,omitempty"`
}
```

tboerger: Do we need the ID? How should I build that? Iterate through the secrets and convert them?

bradrydzewski: yes let's include ID just in case. I might use it in the UI

maybe we do something like this:

```
func (s *Secret) Clone() *Secret {
	return &Secret {
		ID: s.ID,
		Name: s.Name,
		Images: s.Images,
		Events: s.Events,
	}
}
```

and then in the handler we can do something like this:


```
var results []*Secret
for _, s := range secrets {
   results = append(results, s.Clone())
}
```

tboerger: Done

bradrydzewski:
great, LGTM

PR #1699, opened 2016-07-05
Support backslash-escaped slashes in find-and-replace substitution.

vshih:


bradrydzewski:
thanks for the PR and unit tests!

PR #1700, opened 2016-07-05
Add missing space between "your" and "system administrator"

strk:
Otherwise the error on wrong login is very confusing...

### DIFF
--- a/contrib/setup-sassc.sh
+++ b/contrib/setup-sassc.sh
@@ -17,5 +17,5 @@ export SASS_LIBSASS_PATH=/tmp/libsass
 cd sassc
 make
 
-# isntall the sassc binary
+# install the sassc binary
 install -t /usr/local/bin bin/sassc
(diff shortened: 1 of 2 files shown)


tboerger:
The ui will be replaced soon, so I don't think changes make any sense

bradrydzewski:
as @tboerger mentioned we are soon replacing the UI w/ code my fork at [bradrydzewski/drone](https://github.com/bradrydzewski/drone) which no longer uses amber templates. Thanks though!

strk:
@bradrydzewski should I build from your fork then ? will you accept PR there ?
(I'm also interested in running drone from a /subpath/

tboerger:
> @bradrydzewski should I build from your fork then ? will you accept PR there ?
> (I'm also interested in running drone from a /subpath/

The subpath is a bigger efford that needs to be tested. I already started this stuff in the past but it have not been integrated yet.

PR #1702, opened 2016-07-06
Fix PostgreSQL error on empty repository list

### LINKED ISSUE #1649: Error with Postgresql and Gogs
On trying to use Drone with Gogs and Postgresql, authentication is successful but throws the error pq: syntax error at or near ")" on logging in. This is with both latest and Drone 0.4. No error seen in logs.

Gogs with Sqlite works perfectly fine. So I suspect something to do with the postgres DB driver issue

strk:
Closes #1649

### DIFF
--- a/store/datastore/repos.go
+++ b/store/datastore/repos.go
@@ -24,6 +24,7 @@ func (db *datastore) GetRepoListOf(listof []*model.RepoLite) ([]*model.Repo, err
 		repos []*model.Repo
 		args  []interface{}
 		stmt  string
+		err   error
 	)
 	switch meddler.Default {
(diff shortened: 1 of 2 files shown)


tboerger:
:-1: this will break mysql and sqlite

strk:
So another idea could be to always append the empty string
as an element: ` IN (%s,'') `

Unless there's a db-specific sql that could be used ...

strk:
Force-pushed a better approach, just skipping run of a query if the list of repos is known to be empty

tboerger:
LGTM

bradrydzewski:
LGTM thanks for taking the time to fix this!

PR #1706, opened 2016-07-09
Update version in debian package

strk:
found it still pointing at 0.4


bradrydzewski:
0.4 discontinued the debian distribution and then we removed all together in 0.5

PR #1710, opened 2016-07-12
Make badgeNone usable in HTML img tag (returning 200 instead of 404)

strk:
Also adds a print in the log about the kind of failure

### DIFF
--- a/server/badge.go
+++ b/server/badge.go
@@ -43,7 +43,7 @@ func GetBadge(c *gin.Context) {
 
 	build, err := store.GetBuildLast(c, repo, branch)
 	if err != nil {
-		log.Error(err)
+		log.Warning(err)
 		c.String(200, badgeNone)
 		return
 	}
--- a/server/badge.go
+++ b/server/badge.go
@@ -2,7 +2,7 @@ package server
 
 import (
 	"fmt"
-	"log"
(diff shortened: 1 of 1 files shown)


tboerger:
You should use `logrus` like it's done at https://github.com/drone/drone/blob/master/server/build.go#L8 and https://github.com/drone/drone/blob/master/server/build.go#L156 otherwise it's :-1: for me.

server/badge.go
```diff
@@ -2,6 +2,7 @@ package server
 
 import (
 	"fmt"
+	log "github.com/Sirupsen/logrus"
```
tboerger: Just a tiny nitpick, put that one line below with a blank line separated from the core imports.

tboerger:
LGTM

server/badge.go
```diff
@@ -42,7 +43,8 @@ func GetBadge(c *gin.Context) {
 
 	build, err := store.GetBuildLast(c, repo, branch)
 	if err != nil {
-		c.String(404, badgeNone)
+		log.Warning(err)
```
bradrydzewski: actually I'll just adjust this later. I need to go through an audit logging anyway and I don't want to delay the PR :)

PR #1722, opened 2016-07-19
fix dns resolution issues due to static linking and cgo

jdamick:
Running into issues on ubuntu 14.04 due to name resolution:

```dial tcp: lookup xx.redacted.xxx: no such host```

This was resolved by using netgo, this has been seen in:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=63731 and https://github.com/docker/docker/issues/9449

gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,10 @@ PACKAGES = $(shell go list ./... | grep -v /vendor/)
 
 ifneq ($(shell uname), Darwin)
 	EXTLDFLAGS = -extldflags "-static" $(null)
+	EXTFLAGS = -tags netgo
 else
 	EXTLDFLAGS =
+	EXTFLAGS =
 endif
 
 all: gen build_static
@@ -44,7 +46,7 @@ test_postgres:
 build: build_static build_cross build_tar build_sha
 
 build_static:
-	go build --ldflags '${EXTLDFLAGS}-X github.com/drone/drone/version.VersionDev=$(DRONE_BUILD_NUMBER)' -o release/drone github.com/drone/drone/drone
+	go build --ldflags '${EXTLDFLAGS}-X github.com/drone/drone/version.VersionDev=$(DRONE_BUILD_NUMBER)' $(EXTFLAGS) -o release/drone github.com/drone/drone/drone
 
 # TODO this is getting moved to a shell script, do not alter
 build_cross:

### CODE REVIEW

**Comment by bradrydzewski** [comment]
@jdamick if you look at the Dockerfile I set `GODEBUG=netdns=go` to force the pure Go resolver. Is this not working as expected?

see https://golang.org/pkg/net/#hdr-Name_Resolution

```
export GODEBUG=netdns=go    # force pure Go resolver
export GODEBUG=netdns=cgo   # force cgo resolver
```

**Comment by jdamick** [comment]
yeah that does work - missed it in the dockerfile, was just running it from cmd line..

**Comment by stevenjlho** [comment]
I have the same problem. But when i run a docker image after i build docker image from source , the error message show below:

    docker: Error response from daemon: oci runtime error: exec: "/drone": permission denied.

**Comment by bradrydzewski** [comment]
> I have the same problem

the fix to this problem is to start the docker container with  `-e "GODEBUG=netdns=go"`

> docker: Error response from daemon: oci runtime error: exec: "/drone": permission denied.

this error can happens when the binary doesn't exist or wasn't correctly compiled. You first need to make sure you compile drone using the proper commands as defined in the `.drone.yml` to create the binary.

then you need to make sure you are building the image on a linux machine or inside a linux virtual machine. Drone uses CGO which means it needs to be compiled on its target architecture (linux) when creating a linux docker image. If you try to compile drone on osx and then build the Docker image you will see errors.

note that for local testing purposes you can run drone directly on osx or windows. You don't need to run drone itself in a container.

**Comment by stevenjlho** [comment]
> the fix to this problem is to start the docker container with -e "GODEBUG=netdns=go"
Can i add this environment variable `GODEBUG=netdns=go` to dronerc file?

**Comment by bradrydzewski** [comment]
@stevenjlho the dronerc is really just a docker environment variable file and provides a way to load environment variables from a file instead of with `-e`

```
  -e, --env value                   Set environment variables (default [])
      --env-file value              Read in a file of environment variables (default [])
```

so yes, any environment variables you add to the file loaded by `--env-file` will get added to the docker container when it is created.

stevenjlho:
And I am  compile drone [from source](https://github.com/drone/drone#from-source) in Ubuntu, and use command `docker build --rm -t drone/drone .` to build docker image, then run a docker container according to the [document](http://readme.drone.io/setup/overview/#create-and-run-a-drone-container:68f80267fa3a50980dbb745a782b8dca). 

    REMOTE_DRIVER=gogs
    REMOTE_CONFIG=http://gogs.test.amwares.com?open=false
    GODEBUG=netdns=go

This is my dronerc content.

bradrydzewski:
@stevenjlho please use the support channel for general usage questions
https://gitter.im/drone/drone

PR #1724, opened 2016-07-21
ensure client uses ProxyFromEnvironment

bradrydzewski:


rbuckland:
Patch backported to v0.4.2 here https://github.com/rbuckland/drone-cli/tree/backport1724

PR #1725, opened 2016-07-21
Run Matrix by default when running `drone exec`

frapposelli:
This PR adds a new `--matrix-all` flag to `drone exec` that will run all the matrix permutations as specified in the `.drone.yml` file.

I'm considering it a WIP as I'd like to discuss the following points before considering it for merging:

- Failure behavior (right now a failure in a job is fatal to the entire `drone exec` run).
- Default behavior for matrix (right now there is no change to default). 
- Announcing the job run to the user.

As for the last point, right now the job is announced with:

```go
fmt.Printf("Running Matrix job #%d\n", job.Number)
```

Which may not be ideal, and definitely not standing out from the build log.

### DIFF
--- a/drone/exec.go
+++ b/drone/exec.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
@@ -13,6 +14,7 @@ import (
 	"github.com/drone/drone/build/docker"
 	"github.com/drone/drone/model"
 	"github.com/drone/drone/queue"
+	"github.com/drone/drone/yaml"
 
 	"github.com/codegangsta/cli"
 	"github.com/joho/godotenv"
@@ -338,7 +340,7 @@ func exec(c *cli.Context) error {
 		Pull:      c.Bool("pull"),
 	}
 
-	payload := queue.Work{
+	payload := &queue.Work{
 		Yaml:     string(file),
 		Verified: c.BoolT("yaml.verified"),
 		Signed:   c.BoolT("yaml.signed"),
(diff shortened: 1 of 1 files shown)


bradrydzewski:
LGTM

I'm ok with merging as-is or improving if you have ideas. Really up to you. We can always to follow-up pull requests to incrementally improve. Either way, thanks for the PR!

> Failure behavior (right now a failure in a job is fatal to the entire drone exec run).

I don't really have a preference here, so whatever you think is the best approach works for me.

> Default behavior for matrix (right now there is no change to default).

I am fine with changing to run the matrix by default. It probably makes more sense. We could still keep the `--matrix` field to give the option to filter which jobs are executed.

> Announcing the job run to the user.

current implementation looks good

frapposelli:
> I am fine with changing to run the matrix by default. It probably makes more sense. We could still keep the --matrix field to give the option to filter which jobs are executed.

I'm going to amend the PR and introduce this change, I think it makes more sense for people with a matrix in their file to have a good OOTB experience and be able to run the file right away instead of getting an error.

frapposelli:
Amended, now the matrix is run automatically, while you can still override the run using `--matrix`.

frapposelli:
ping @bradrydzewski 😉

drone/exec.go
```diff
@@ -382,12 +384,46 @@ func exec(c *cli.Context) error {
...
+			Environment: axis,
+		})
+		count++
+	}
+
+	for _, job := range jobs {
+		fmt.Printf("Running Matrix job #%d\n", job.Number)
+		p := *payload
+		p.Job = job
+		p.Build.Number = p.Build.Number + job.Number
```
bradrydzewski: this line should get removed. The build number is set using `DRONE_BUILD_NUMBER`  and this would override. All build meta-data is exposed via environment variables

frapposelli: Ha!, that's a leftover from testing 😄

frapposelli: Fixed in frapposelli/drone@46dba43

PR #1728, opened 2016-07-21
Disallow to deactivate repo without admin permissions

Bugagazavr:


bradrydzewski:
this is handled through middleware:
https://github.com/drone/drone/blob/master/router/router.go#L87

right now it uses `session.MustPush` but could be changed to `session.MustAdmin`. Also note the perms are stored in the context, so it is un-necessary to fetch them again.

Bugagazavr:
i try to use `repo.DELETE("", session.MustAdmin(), server.DeleteRepo)`, but when i try to delete drone from my own repo i got 413 ( typo? )

https://github.com/drone/drone/blob/master/router/middleware/session/user.go#L80

And so, user should be repository admin, but not a system admin, this middleware check is user a drone admin

bradrydzewski:
@Bugagazavr `session.MustRepoAdmin` would need to be created. Forgot that `session.MustAdmin` exists and is used for admin roles.

Bugagazavr:
@bradrydzewski done

Bugagazavr:
maybe add MustRepoAdmin to chown?

bradrydzewski:
> add MustRepoAdmin to chown

sure!

Bugagazavr:
done

bradrydzewski:
thanks!

PR #1729, opened 2016-07-21
Fix cli args typo

vaijab:
This was preventing from using --github-secret while env var worked.


bradrydzewski:
great catch, thanks!

PR #1735, opened 2016-07-27
Add custom build parameters to build re-start / deploy

### LINKED ISSUE #894: Allow environment variables to be set on build basis + ondemand builds
I would like to propose a feature and also to offer my help to implement/develop it if it is accepted. The feature is important for organisations with many repositories and many teams. It addresses the security of environment variables.
Feature:
  Allow the client to run builds on arbitrary branches/commits with specific environment variables so that it doesn't need to expose secrets in config files or web ui.

Example:
``curl -X POST https://drone.io/api/repos/github.com/foo/bar/run?branch=develop&env[0]=[secret_name=secret_value]``
  A CLI could be used to automate this kind of request(e.g. run it from the current branch and it picks up the branch and repository) so you may need just one more command after push. 
The command above would run a new build using the latest changes of the develop branch with a secret set as environment variable(secret_name=secret_value). I find this transparent and more secure than defining the secrets in config files or the web UI. 
Use case:
   Many projects require external dependencies during the build process. An example is a go project which imports many other repositories/packages. The external dependencies have specific access restrictions (e.g. only some developers have access to them). The develop has no problem to download these dependencies locally because it's supposed to already have access to them (though github teams ACL or whatever). However on the build server various secrets/tokens are required(tokens, password etc). 
The current practice is to simply define the secrets/tokens as environment variables. Usually the token is global because the tokens management becomes laborious as the number of dependencies and repositories increases. This presents a great security risk because a rough team member(or someone who hacked the account)  with access to a small repository within the organisation could simply commit code to print the secrets/tokens in the build log and use them to access restricted dependencies. This may even go unnoticed because it may commit the code in an unmonitored branch so the change/build is not reviewed.
  Another issue is that even if you would set a different secret/token for each repository(unfeasible)  there is still the risk for someone who leaves the team to use the secrets later when its access was restricted. It's not feasible to change all the tokens whenever someone leaves a specific team.

vaijab:
This change also adds build parameters to the deploy endpoint. However, if
people disagree about build params to deploy, I can remove it.

Fixes #894

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -76,21 +76,21 @@ type Client interface {
 	BuildQueue() ([]*model.Feed, error)
 
 	// BuildStart re-starts a stopped build.
-	BuildStart(string, string, int) (*model.Build, error)
+	BuildStart(string, string, int, map[string]string) (*model.Build, error)
 
 	// BuildStop stops the specified running job for given build.
 	BuildStop(string, string, int, int) error
 
 	// BuildFork re-starts a stopped build with a new build number, preserving
 	// the prior history.
-	BuildFork(string, string, int) (*model.Build, error)
+	BuildFork(string, string, int, map[string]string) (*model.Build, error)
 
 	// BuildLogs returns the build logs for the specified job.
 	BuildLogs(string, string, int, int) (io.ReadCloser, error)
 
 	// Deploy triggers a deployment for an existing build using the specified
 	// target environment.
-	Deploy(string, string, int, string) (*model.Build, error)
+	Deploy(string, string, int, string, map[string]string) (*model.Build, error)
 
 	// AgentList returns a list of build agents.
 	AgentList() ([]*model.Agent, error)
--- a/server/build.go
+++ b/server/build.go
@@ -248,6 +248,18 @@ func PostBuild(c *gin.Context) {
 		build.Deploy = c.DefaultQuery("deploy_to", build.Deploy)
 	}
 
+	// Read query string parameters into buildParams, exclude reserved params
+	var buildParams = map[string]string{}
+	for key, val := range c.Request.URL.Query() {
+		switch key {
+		case "fork", "event", "deply_to":
+		default:
+			// We only accept string literals, because build parameters will be
+			// injected as environment variables
+			buildParams[key] = val[0]
+		}
+	}
+
 	// todo move this to database tier
 	// and wrap inside a transaction
 	build.Status = model.StatusPending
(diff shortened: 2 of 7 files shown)


vaijab:
I will write documentation as well, just want to make sure that the changes are good.

client/client.go
```diff
@@ -76,21 +76,21 @@ type Client interface {
 	BuildQueue() ([]*model.Feed, error)
 
 	// BuildStart re-starts a stopped build.
-	BuildStart(string, string, int) (*model.Build, error)
+	BuildStart(string, string, int, []string) (*model.Build, error)
```
bradrydzewski: maybe instead do `BuildStart(string, string, int, ...string)` so that it can remain an optional input parameter?

bradrydzewski: edit: maybe instead this should be `map[string]string` and the caller should be in charge of converting to the `map[string]string` format.

the reason is that this client package can be imported and used by programs that want to interact with drone, and using `map[string]string` seems more declarative that using `[]string` and having to document `key=value` format for each item in the slice

vaijab: > instead this should be map[string]string

Yeah, that makes sense. I will update it.

server/build.go
```diff
@@ -248,13 +248,30 @@ func PostBuild(c *gin.Context) {
...
+		// We only accept string literals, because build parameters will be
+		// injected as environment variables
+		buildParams[key] = val[0]
+		for _, r := range reservedParams {
+			if r == key {
+				delete(buildParams, r)
+			}
+		}
+	}
+
```
bradrydzewski: we could probably slightly alter the code to avoid the extra allocation of the map with a global variable, and avoid extra operations adding and removing items from the map as well with lookup:

```
	var buildParams = map[string]string{}
	for key, val := range c.Request.URL.Query() {
		if _, ok := reservedParams[key]; !ok {
			buildParams[key]=val
		}
	}
```

or a simple switch statement:

```
	var buildParams = map[string]string{}
	for key, val := range c.Request.URL.Query() {
		switch {
                case "fork", "event", "deply_to":
		default:
			buildParams[key]=val
		}
	}
```

bradrydzewski:
added some minor notes / thoughts but otherwise looks good

vaijab:
@bradrydzewski great. I will address the comments.

vaijab:
@bradrydzewski updated.

bradrydzewski:
great thanks so much!

PR #1737, opened 2016-07-31
Organization/Team secrets

tboerger:
I just want to open this PR even if i'm still not done with it to show the progress since there are various people interested in that feature.


store/store.go
```diff
@@ -59,16 +59,31 @@ type Store interface {
...
+	GetTeamSecret(string, string) (*model.TeamSecret, error)
+
+	// SetTeamSecret sets the named team secret.
+	SetTeamSecret(*model.TeamSecret) error
+
+	// DeleteTeamSecret deletes the named team secret.
+	DeleteTeamSecret(*model.TeamSecret) error
+
+	// GetMergedSecretList gets a list of repo and team secrets
+	GetMergedSecretList(*model.Repo) ([]*model.Secret, error)
```
bradrydzewski: we can actually merge the two lists in `store` instead of `datastore`. That way, if we have multiple implementations of datastores, we don't need to duplicate the merge logic

store/datastore/secret.go
```diff
@@ -2,52 +2,32 @@ package datastore
...
 	"github.com/drone/drone/model"
-	"github.com/russross/meddler"
 )
 
-func (db *datastore) GetSecretList(repo *model.Repo) ([]*model.Secret, error) {
-	var secrets = []*model.Secret{}
-	var err = meddler.QueryAll(db, &secrets, rebind(secretListQuery), repo.ID)
-	return secrets, err
-}
+func (db *datastore) GetMergedSecretList(repo *model.Repo) ([]*model.Secret, error) {
```
bradrydzewski: mentioned this below, but let's move this up a level. We can merge the two lists in `store` instead of `datastore`. That way, if we have multiple implementations of datastores, we don't need to duplicate the merge logic

bradrydzewski:
awesome work! added a few minor comments related to the merge function and moving that up a level. But otherwise looks awesome and almost ready to merge!

tboerger:
> awesome work! added a few minor comments related to the merge function and moving that up a level. But otherwise looks awesome and almost ready to merge!

Thanks, I already realized that I need to write a middleware to check org admin rights. I will address the other comments.

donny-dont:
@tboerger should that be a requirement? Or just Drone admin?

tboerger:
@donny-dont an org admin should be able to add org secrets :). Global secrets should be done only by drone admins.

tboerger:
@bradrydzewski IMHO this should work now.

bradrydzewski:
really sorry for the delay but I sort of have a good excuse :)

merging now!

PR #1738, opened 2016-08-01
Expose GitHub machine user options.

johnwchadwick:
I noticed there was support for a so-called GitHub "machine user." This is a useful thing to have for when you have private repositories that depend on other private repositories.

Unfortunately, this support is just barely not publicly exposed anywhere it seems, as the option is missing from the CLI flags. This patch adds those flags. I'm currently running this in production and it seems to be doing fine. (At the very least, it is absolutely writing the username/password to netrc.) I believe it is also possible to use an OAuth token here by putting the username as a token and the password as `x-oauth-basic`, but I did not actually test this. (I can only assume that this is how it is internally handled already, since OAuth is used to begin with.)

### DIFF
--- a/drone/server.go
+++ b/drone/server.go
@@ -124,6 +124,16 @@ var serverCmd = cli.Command{
 				"read:org",
 			},
 		},
+		cli.StringFlag{
+			EnvVar: "DRONE_GITHUB_GIT_USERNAME",
+			Name:   "github-git-username",
+			Usage:  "github machine user username",
+		},
+		cli.StringFlag{
+			EnvVar: "DRONE_GITHUB_GIT_PASSWORD",
+			Name:   "github-git-password",
+			Usage:  "github machine user password",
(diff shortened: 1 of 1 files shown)


bradrydzewski:
yep, this is a new feature that we added in 0.5 and this PR looks like it properly enables setting these variables. Thanks!

> This is a useful thing to have for when you have private repositories that depend on other private repositories.

as a minor clarification this is generally still possible without using a machine account. The default netrc that gets generated is based on the account that activated the repository in drone. As long as this account has access to private dependencies, they are accessible from your build environment.

the biggest benefit to the machine account is that it prevents user oauth tokens from being injected directly into (private) build environments, injecting the machine account instead. This provides a bit more individual security.

again, just a minor clarification because someone could end up at this thread via google search and think that machine accounts are required for private dependencies, which isn't necessarily the case :)

Issue #1746, opened 2016-08-08
Drone 0.5 fails on getting builds for some repos


onbjerg:
Hi.

I tried using Drone 0.5 today on k8s. I have a previously populated Postgres database with builds from 0.4.

A lot of the repo build lists are working, but some of them aren't. When I navigate to some repos to see the build history, the API returns 500.

I checked the logs, but nothing of note is in there. It might be because when ``store.GetBuildList`` fails, the API just returns HTTP 500 without logging the actual error ([ref](https://github.com/drone/drone/blob/master/server/build.go#L22)).

Could we log the error so I can debug further? As I said, it only happens for some repos of mine, but there is no pattern so far, so I don't really know what's going on.

bradrydzewski:
the 0.4 log format is not compatible with 0.5 data so this is expected. There is not a 100% clean path to upgrade to 0.5. Note that 0.5 is also not released yet.

bradrydzewski:
sorry misread this. I thought build history was referring to the logs. I see you have a PR to log the error message, which I've commented on and recommended returning the error message to surface in the UI

onbjerg:
The build history is not showing for some repos. Is this expected?

onbjerg:
Or? 😟

qrpike:
I'm having this issue as well, possibly a DB corruption issue as it used to work fine.

bradrydzewski:
This issue centers around the fact that when you upgrade legacy versions of drone, the newer version of drone cannot read the old logs (they were stored in different formats). Nothing to do with database corruption.

qrpike:
@bradrydzewski I've been using v 0.5 for what seems like almost a year now. I have the mysql volume as a ceph rbd. I switched it to run on a different server, now everything but the actual 'builds' list is working. mysqlcheck says everything is fine :/

PR #1747, opened 2016-08-08
Log errors in `server.GetBuilds`

onbjerg:
Getting the build history for some repos on my Drone 0.5 installation fails (for more details see #1746), but the error is not logged. This PR will log the error.

### DIFF
--- a/server/build.go
+++ b/server/build.go
@@ -23,7 +23,7 @@ func GetBuilds(c *gin.Context) {
 	repo := session.Repo(c)
 	builds, err := store.GetBuildList(c, repo)
 	if err != nil {
-		c.AbortWithStatus(http.StatusInternalServerError)
+		c.AbortWithError(http.StatusInternalServerError, err)
 		return
 	}
 	c.JSON(http.StatusOK, builds)


tboerger:
Maybe we should do that on more places

onbjerg:
Probably. Should I make this a general purpose "logging improvement" PR?

donny-dont:
@bradrydzewski any particular reason to not show the err as well across the board?

bradrydzewski:
I started doing this in 0.5
https://github.com/drone/drone/blob/master/server/user.go#L54

The issue with abort is that it logs the error but does not return the error string in the response. I think we want the reverse. Surface errors in the UI instead of the logs

onbjerg:
@bradrydzewski The two messages should not necessarily be the same. Some information should be in logs for easier debugging, while a more user-friendly (read: non-sensitive) error message should be displayed in the UI.

bradrydzewski:
> The two messages should not necessarily be the same. Some information should be in logs for easier debugging, while a more user-friendly (read: less technical) error message should be displayed in the UI.

yes and no. Drone is a technical product for developers so we can get away with more technical error message. The pattern we are currently following in newer sections of the code is to prepend the error message with something more descriptive. For example:

```
	if err != nil {
		c.String(500, "Error fetching repository list. %s", err)
		return
	}
```

For server side logging we use `logrus` since it gives us more control over logging levels than the gin defaults. I personally try to limit server-side logging to areas where people consistently experience issues (logins and hooks and agent connectivity). People constantly copy/paste irrevelant debug logs into gitter because they think that means there is a problem, which results in more of my time spent on unnecessary support. I'm therefore more of a logging minimalist.

onbjerg:
Alright. Either way, I would still very much appreciate some better error messages and/or logs so I could debug the issue I am having in #1746

onbjerg:
Anything? It clearly must be a bug, unless the advice is to start from scratch (i.e. delete all data from 0.4).

tboerger:
There is no update path yet for the data. Just the db structure gets migrated. Maybe we can do something before the 0.5 release

onbjerg:
I don't know if that could be it. The majority of 0.5 worked with my 0.4 data.

onbjerg:
@bradrydzewski Should we just surface the internal error in the UI, then? If so, I guess this PR can be closed, and I might open a new one instead.

I still haven't figured out why Drone 0.5 didn't work some of the time in my case. Usually I would use the logs to get an indicator of what might be wrong, but seeing as there wasn't any and there are no plans on adding them, I would really appreciate if my issue could be resolved otherwise.

bradrydzewski:
I'm sorry if my prior comment was unclear. We should push errors to the user interface where they can be displayed. This is what we are doing for most of the endpoints, and the same pattern should be followed here.

```
    if err != nil {
        c.String(500, "Error fetching repository list. %s", err)
        return
    }
```

Feel free to open a new PR with the suggested change.

PR #1752, opened 2016-08-11
Integrated gogs webhook deletion

### LINKED ISSUE #1603: Gogs delete hooks
Gogs does not currently expose an API to remove hooks. This means when you de-activate a Gogs repository in Drone the previously created post-commit hooks are not removed. If you re-add the repository you will receive duplicate hooks and launch multiple builds.

We should explore adding the appropriate endpoints to the Gogs API client.

tboerger:
Fixes #1603

### DIFF
--- a/remote/gogs/gogs.go
+++ b/remote/gogs/gogs.go
@@ -214,6 +214,18 @@ func (c *client) Activate(u *model.User, r *model.Repo, link string) error {
 
 // Deactivate is not supported by the Gogs driver.
 func (c *client) Deactivate(u *model.User, r *model.Repo, link string) error {
+	client := c.newClientToken(u.Token)
+	hooks, err := client.ListRepoHooks(r.Owner, r.Name)
+	if err != nil {
+		return err
+	}
+	match := matchingHooks(hooks, link)
+	if match == nil {
+		return nil
+	}
+	if match != nil {
+		return client.DeleteRepoHook(r.Owner, r.Name, match.ID)
+	}
 	return nil
 }
 
@@ -254,4 +266,26 @@ func (c *client) newClientToken(token string) *gogs.Client {
 		client.SetHTTPClient(httpClient)
 	}
 	return client
+}
+
+// helper function to return matching hook.
+func matchingHooks(hooks []*gogs.Hook, rawurl string) *gogs.Hook {
+	link, err := url.Parse(rawurl)
+	if err != nil {
+		return nil
+	}
+	for _, hook := range hooks {
+		if hook.ID == 0 {
+			continue
+		}
+		v, ok := hook.Config["url"]
+		if !ok {
+			continue
+		}
+		hookurl, err := url.Parse(v)
+		if err == nil && hookurl.Host == link.Host {
+			return hook
+		}
+	}
+	return nil
 }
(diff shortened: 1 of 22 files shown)


bradrydzewski:
awesome, thanks for working on this one! I think we might need to mirror this after the GitHub hook removal as well. Notice we have logic to find the matching hook:
https://github.com/drone/drone/blob/master/remote/github/github.go#L238

the matching hooks function parses the URL to remove the token from the url query string. Otherwise it won't properly match the link that is passed in
https://github.com/drone/drone/blob/master/remote/github/github.go#L314

EDIT: I'm looking at your implementation and it does look like the `HasPrefix` might work fine. Perhaps for consistency we can break into a helper function similar to the GitHub implementation?

tboerger:
I will do it the same like GitHub. And than I have to fix the tests, looks like there have been something broken because of the client update.

tboerger:
@bradrydzewski I don't get the tests running correctly but at least the deletion of webhooks works now. Maybe you can do a small addition afterwards to get the tests correct since you are deeper in it?

bradrydzewski:
what issues are you seeing w/ the unit tests? maybe I can help

remote/gogs/gogs.go
```diff
@@ -255,3 +270,25 @@ func (c *client) newClientToken(token string) *gogs.Client {
 	}
 	return client
 }
+
+// helper function to return matching hook.
+func matchingHooks(hooks []*gogs.Hook, rawurl string) *gogs.Hook {
```
bradrydzewski: can you add some unit tests for this function similar to
https://github.com/drone/drone/blob/master/remote/bitbucket/bitbucket_test.go#L233

remote/gogs/gogs.go
```diff
@@ -214,6 +214,21 @@ func (c *client) Activate(u *model.User, r *model.Repo, link string) error {
...
+	hooks, err := client.ListRepoHooks(r.Owner, r.Name)
+	if err != nil {
+		return err
+	}
+	match := matchingHooks(hooks, link)
+	if match == nil {
+		return nil
+	}
+	if match != nil {
+
```
bradrydzewski: I think we can remove this `match != nil` since we return if `match == nil`?

remote/gogs/gogs.go
```diff
@@ -214,6 +214,21 @@ func (c *client) Activate(u *model.User, r *model.Repo, link string) error {
...
+		err := client.DeleteRepoHook(r.Owner, r.Name, match.ID)
+		fmt.Println(err)
```
bradrydzewski: remove println

remote/gogs/gogs.go
bradrydzewski: can probably clean up something like this:

```
	hook := matchingHooks(hooks, link)
	if hook != nil {
		return client.DeleteRepoHook(r.Owner, r.Name, match.ID)
	}
	return nil
```

MorphBonehunter:
@tboerger could you please take a look at this again?

tboerger:
> @tboerger could you please take a look at this again?

A little bit busy at the moment, but I try to come back ASAP.

vtolstov:
any progress in this pr?

tboerger:
Sorry, I forgot about this PR.

vtolstov:
does this needs to be merged before gitea driver?

tboerger:
Closing this in favor of go-scm

PR #1753, opened 2016-08-11
Add support for `drone build start foo/bar last`

devth:
This starts the last build on the default branch.

Alternatively: should it default to the last build when the arg is explicitly omitted?

### DIFF
--- a/drone/build_start.go
+++ b/drone/build_start.go
@@ -35,14 +35,26 @@ func buildStart(c *cli.Context) (err error) {
 	if err != nil {
 		return err
 	}
-	number, err := strconv.Atoi(c.Args().Get(1))
-	if err != nil {
-		return err
-	}
 
 	client, err := newClient(c)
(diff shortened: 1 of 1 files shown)


drone/build_start.go
```diff
@@ -35,16 +35,28 @@ func buildStart(c *cli.Context) (err error) {
...
+	var number int
+	if buildArg == "last" {
+		// Fetch the build number from the last build
+		build, err := client.BuildLast(owner, name, "")
+		if err != nil {
+			return err
+		}
+		number = build.Number
+	} else {
+		number, err = strconv.Atoi(c.Args().Get(1))
```
tboerger:
```
number, err = strconv.Atoi(buildArg)
```

devth: Thanks, fixed.

bradrydzewski:
LGTM thanks!

PR #1756, opened 2016-08-13
Remove the trailing / from DRONE_SERVER env in agent when it's there

josmo:
Seems like enough people are running into this that just removing it if it exists would be worth it.
@bradrydzewski let me know if it's even worth it or if you'd rather have it somewhere else :)


bradrydzewski:
ha, absolutely. thanks!

PR #1760, opened 2016-08-13
Adding the ability to label containers launched by drone

josmo:
Adding the ability to label containers that are launched using  
labels:
    something: value
    something2: value2

@bradrydzewski couple things here. I'm not sure if the label transformation is needed but I was trying to keep things consistent.

Also note on the use case.  When plugin containers for aws or anything that's using the metadata service right now the container gets the ec2s role. This change should get us started on being able to assign roles per container that drone launches instead of the ec2 instance that the agent is running on by using a metadata proxy type of solution.


bradrydzewski:
Transform can be removed. Transforms are used to modify a yaml before compilation to sanitize values, enforce security, defaults, etc. In this case it is not required

josmo:
Thanks! that's what I figured so I'll take it out and update

josmo:
Done! Let me know if you want anything else different :)

PR #1763, opened 2016-08-14
Fix issue with connection being closed by load balancer aws elb

kushtrimjunuzi:
This fix is configurable by enabling/disabling through environment variable at drone server:
- DRONE_CLOSE_NOTIFY_TIMEOUT=60s

### DIFF
--- a/drone/server.go
+++ b/drone/server.go
@@ -73,12 +73,7 @@ var serverCmd = cli.Command{
 			EnvVar: "DRONE_CLOSE_NOTIFY_TIMEOUT",
 			Name:   "close-notify-timeout",
 			Usage:  "close notify timeout",
-			Value:  time.Second * 50,
-		},
-		cli.BoolFlag{
-			EnvVar: "DRONE_ENABLE_CLOSE_NOTIFY_TIMEOUT",
-			Name:   "enable-close-notify-timeout",
-			Usage:  "enable close notify timeout",
+			Value:  0,
 		},
 		cli.StringFlag{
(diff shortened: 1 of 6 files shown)


tboerger:
I think the enable flag should be dropped and it should be enabled with a value >0 and disabled with the value 0.

tboerger:
Not an expert related to that part but LGTM

kushtrimjunuzi:
This is working quite well and all you have to do is at AWS load balancer, set idle timeout same as DRONE_CLOSE_NOTIFY_TIMEOUT and for jobs that takes time you have to increase aws idle timeout and set flag to that value, if job processing time is grater than idle timeout, you don't see the feeds/logs on the UI and you need to restart drone server in order to receive those logs, but the job will be completed at agent side.
The whole problem with idle timeout is that AWS load balancer has a limit of keeping connection open and unfortunately drone server does not send heartbeat/ping-pong to keep this TCP connection live for pulling api other long pulling apis (except websocket apis) and AWS ELB detects that there was not data going through this connection and it will close connection, but the drone server is not aware of that close connection, so we need to handle that case and the fix is on this commit.
Read more about that here http://felixge.de/2014/08/26/tcp-keepalive-with-golang.html
and here http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-idle-timeout.html
Thanks.

bradrydzewski:
closing because the master branch no longer uses long polling, and instead uses the [drone/mq](https://github.com/drone/mq) package with a websocket transport. This should therefore resolve this particular problem. How to configure ELB for websockets is a separate issue though :)

PR #1766, opened 2016-08-15
fixing static linking glibc problems: https://github.com/drone/drone/…

sokoow:
fixing …issues/1765

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 PACKAGES = $(shell go list ./... | grep -v /vendor/)
 
 ifneq ($(shell uname), Darwin)
-	EXTLDFLAGS = -extldflags "-static" $(null)
+	EXTLDFLAGS = -extldflags "-static-libgcc -static-libstdc++" $(null)
 else
 	EXTLDFLAGS =
 endif


Makefile
```diff
@@ -3,7 +3,7 @@
 PACKAGES = $(shell go list ./... | grep -v /vendor/)
 
 ifneq ($(shell uname), Darwin)
-	EXTLDFLAGS = -extldflags "-static" $(null)
+	EXTLDFLAGS = -extldflags "-static-libgcc -static-libstdc++" $(null)
 else
```
bradrydzewski: lets add another `if` block that only uses these new settings for arm. I don't want to change the amd64 compiler flags since they have been working quite well and I don't want to introduce potential instability to the project.

bradrydzewski: for example I would imagine something like this:


```sh
ifneq ($(shell uname), Darwin)
	EXTLDFLAGS = -extldflags "-static" $(null)
else
	EXTLDFLAGS =
endif

if ($(shell uname -m), armv7l)
	EXTLDFLAGS = -extldflags "-static-libgcc -static-libstdc++" $(null)
endif
```

bradrydzewski: or maybe try building on a different host environment than ubuntu? for local developement you can just run `go build`. There is no need to use the make file and statically compile. For building the distribution we can choose whatever build environment we want via dockerfile

tboerger: With your example you will also break alpine arm builds?

bradrydzewski: @tboerger yea, I think it is best to try to build with just `-static` on a platform that doesn't have known issues. For example instead of building on ubuntu one can build on alpine or another base image. This is an example 0.4 yaml that builds on arm:
https://github.com/armdrone/drone/blob/master/.drone.yml

bradrydzewski:
I'm going to close this because as @tboerger mentions it is tied to libgcc and environments like alpine won't have libgcc. I think instead of changing the build command, we should change the host machine that drone is compiled on, as done by the 0.4 arm part:
https://github.com/armdrone/drone/blob/master/.drone.yml

bradrydzewski:
@firecyberice any thoughts on how we can get an official arm build for drone 0.5?

firecyberice:
@bradrydzewski 
I just built a docker image that way:
https://gist.github.com/firecyberice/f1528f1fc8ad98fb7232805e4e03d006
The binary has been cross-compiled on amd64 with `make deps gen build-cross`.
I suppose even the docker image can be built on amd64 as there is no RUN defined.

bradrydzewski:
@firecyberice replied at https://github.com/drone/drone/issues/1767#issuecomment-240795490

PR #1771, opened 2016-08-25
choose ambassador by platform

bradrydzewski:


tboerger:
Not being configurable? Especially for setups without hub access or with special requirements within companies

bradrydzewski:
> Not being configurable? Especially for setups without hub access or with special requirements within companies

@tboerger both 0.4 and 0.5 have hardcoded ambassador images. This pull request therefore does not attempt to solve the problem where people want to change the default ambassador image, since it would be a more complex change. This PR instead attempts to provide an arm-compatible ambassador.

tboerger:
Do I miss something or should it just be a single option within the agent that just resolves the default image for the architecture it's compiled for?

bradrydzewski:
@tboerger it would be 3 options. Image name, entrypoint (string slice) and command (string slice) since different images / operating systems might require different commands. This is not something I'm looking to solve in this PR. It could always be a separate follow-up PR

tboerger:
:+1:

bradrydzewski:
I'm actually hoping docker will implement pod networking so we can use `docker network create` along with the existing `docker volume create`, and we can ditch pod networking all together. I'm not sure how likely they are to implement such a feature, but one can dream :)

vaijab:
LGTM

tboerger:
After merging this I can create a followup to provide those options for the agent.

alkar:
LGTM, thanks @bradrydzewski

bradrydzewski:
> After merging this I can create a followup to provide those options for the agent.

@tboerger after discussing with the docker folks there may be a way to remove the ambassador container completely. Let's discuss this further in gitter before we expose any more of the ambassador behavior.

PR #1772, opened 2016-08-25
Correct escaping of commands in YAML

benschumacher:
Two changes on the line:

1. when replacing, make sure we're starting with previously escaped string.
2. use \$ to escape $ instead of $\ -- the latter works in *most cases*, but the former is more correct.

This code could be slightly refactored to make it more testable.

Any interest?


benschumacher:
Just to clarify, the issue was that commands such as this one would fail:

```yaml
commands:
  - curl -sSL https://get.docker.com/builds/Linux/i386/docker-latest.tgz | tar --strip-components=1 -C /usr/local/bin -zxvf - docker/docker
```

tboerger:
LGTM

PR #1773, opened 2016-08-25
OAuth2 GHE Always Register BustedURL

brgaulin:
When going through the oauth2 module in go, there is a list of "busted" URLs which don't implement OAuth2 properly.

Github is on that list, however Github Enterprise(GHE) uses different URLs for each instance. Instead of adding all of these to the global list in the oauth module, we can simply flag all of our tokens as broken until github updates their API.


brgaulin:
This may only be an issue with older GHE installations. I'm still using a version 11 github(very old)

PR #1778, opened 2016-08-31
Add sample gogs web hooks (unused yet)

strk:
I'm hoping for these to be useful for implementing support for Gogs PR checking


bradrydzewski:
thanks! could you move these samples inside the existing `hooks.go` file?
https://github.com/drone/drone/blob/master/remote/gogs/fixtures/hooks.go

strk:
I see the existing HookPush in there has a different format
than the one I sampled. It was added on May 1st, any idea about
which version of Gogs it came from ?

For now I've added a New suffix to the new ones...

remote/gogs/fixtures/hooks.go
```diff
@@ -44,3 +45,223 @@ var HookPush = `
...
+  },
+  "sender": {
+    "id": 1,
+    "username": "strk",
+    "full_name": "",
+    "email": "strk@kbt.io",
+    "avatar_url": "https://avatars.kbt.io/avatar/fe2a9e759730ee64c44bf8901bf4ccc3"
+  }
+}
+'
```
bradrydzewski: looks like using `'` instead of `

remote/gogs/fixtures/hooks.go
```diff
@@ -44,3 +45,223 @@ var HookPush = `
...
```
bradrydzewski: ' instead of `

bradrydzewski:
> I see the existing HookPush in there has a different format than the one I sampled. It was added on May 1st

the payloads actually look quite similar, note that I manually removed fields from the hook that drone doesn't care about so that it was more readable, and easier to audit which fields drone actually used. If you consider that I remove fields, the payloads may actually not be so different.

I'll merge anyway and can sort out at a later time

strk:
On September 1, 2016 10:51:49 PM CEST, Brad Rydzewski <notifications@github.com> wrote:
>> I see the existing HookPush in there has a different format than the
>one I sampled. It was added on May 1st
>
>the payloads actually look quite similar, note that I manually removed
>fields from the hook that drone doesn't care about so that it was more
>readable, and easier to audit which fields drone actually used. If you
>consider that I remove fields, the payloads may actually not be so
>different.
>
>I'll merge anyway and can sort out at a later time

Manual fields removal explains it. Then we don't need the duplicated push hook
--strk;

Sent from hand-held device, sorry for brevity

PR #1783, opened 2016-09-02
Fix repo query when user has access to > 999 remotes

mariusgrigoriu:
Limiting to 999 in the query creates a few issues. 
- If a repo is the 1000th or greater remote, it disappears form the left after a page reload.
- The repo appears to be not active after a page reload (from the account page). And activating it again will not work, making it appear stuck.
- The repo never appears in drone repo ls

Issue #1776 won't be resolved until this PR or something with a similar effect gets merged.

### DIFF
--- a/store/datastore/repos.go
+++ b/store/datastore/repos.go
@@ -28,7 +28,7 @@ func (db *datastore) GetRepoListOf(listof []*model.RepoLite) ([]*model.Repo, err
 	)
 	switch meddler.Default {
 	case meddler.PostgreSQL:
-		stmt, args = toListPosgres(listof)
+		stmt, args = toListPostgres(listof)
 	default:
 		stmt, args = toList(listof)
 	}
--- a/store/datastore/users.go
+++ b/store/datastore/users.go
@@ -35,7 +35,7 @@ func (db *datastore) GetUserFeed(listof []*model.RepoLite) ([]*model.Feed, error
 	)
 	switch meddler.Default {
 	case meddler.PostgreSQL:
-		stmt, args = toListPosgres(listof)
+		stmt, args = toListPostgres(listof)
 	default:
 		stmt, args = toList(listof)
 	}
@@ -55,7 +55,7 @@ func (db *datastore) GetUserFeedLatest(listof []*model.RepoLite) ([]*model.Feed,
 	)
 	switch meddler.Default {
 	case meddler.PostgreSQL:
-		stmt, args = toListPosgres(listof)
+		stmt, args = toListPostgres(listof)
 	default:
 		stmt, args = toList(listof)
 	}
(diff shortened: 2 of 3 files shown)


tboerger:
LGTM but need to sign cla

bradrydzewski:
the 999 limit is in place because sqlite does will not accept an IN statement with > 999 arguments

we can adjust postgres and mysql to accept > 999 but we should still need to put some limits in place to avoid sending a packet > 1mb which would cause a mysql error. I would therefore accept a PR that looks like this:

```
switch {
case meddler.Default == meddler. SQLite && size > 999:
  size = 999		
  listof = listof[:999]	
case size > 9999:
  size = 9999		
  listof = listof[:9999]
}
```

mariusgrigoriu:
Our bitbucket server has 9800 accessible repositories and growing, so a limit at 9999 won't be a viable solution for too long. :(

bradrydzewski:
@marcosnils we can implement a temporary fix of 9999 which should work for your bitbucket installation in the short term. A longer term solution can be to make this limit configurable, which can be added in a separate PR

thiago:
I am with this problem using mysql.
I started to implement a solution to get paginated by a limit set by the system, thus solved the problem in any datasource and do not need to have a ceiling of 9999.

What do you think?

bradrydzewski:
> I am with this problem using mysql. I started to implement a solution to get paginated by a limit set by the system, thus solved the problem in any datasource and do not need to have a ceiling of 9999.

can you explain what you mean by pagination? http pagination? database pagination? batching? This was discussed in the gitter channel and this was my response:

> __tboerger:__ we need a pagination? Or infinite scrolling?

> __me:__ Http pagination wouldn't help the 999 limit for SQLite because we don't know how many results will be returned by the database query. The IN statement finds the intersection of repos in github configured in drone. This could be 0 or 999, we just don't know

> __tboerger:__ so we need to execute multiple 999 batches and combine the results?

> __me:__ yes that might work

so http or database pagination (using `limit` and `offset`) is not a viable solution for the reasons mentioned above. Instead we need to execute multiple queries and combine the results.

In some cases we need to return the last 50 builds for all active repos (for example). If we executed this query in batch multiple times, and combined the results, it could return > 50 results. This means we would need to sort in memory and then return the top 50 from the combined list.

So my thought is that we can of course accept a fix that runs multiple queries and combines the results when > 999 repositories exist. This is going to be a bit more effort but will provide a more permanent fix to the problem. Or we can implement a more temporary fix that uses a 9999 limit for mysql and postgres.

I don't really care which as long as it works as advertised and doesn't break existing functionality.

thiago:
Ok @bradrydzewski, I understand your points! I think was going in right direction, running multiple queries in the database and combines (without http).

I'll finalize the idea and push to discussion.

mariusgrigoriu:
Sorry I was away a bit... 
Seems like multiple requests to the database is more complicated than necessary and would add a lot of code. What if instead we just loaded the whole list of remotes and added repos to memory and performed the comparison using a map. Even if we're talking 100k remotes it shouldn't use that much memory and take long -- and might even be less complex code than status quo.

bradrydzewski:
merged #1929

PR #1791, opened 2016-09-11
Add test case with a normal matrix and an include statement

MorrisJobke:
This is how I thought the include statement would work, but it doesn't. Is this a bug or simply not wanted? Often you have a matrix and just want to extend it for some version and add a spare row to the matrix with only one cell filled.


bradrydzewski:
> This is how I thought the include statement would work, but it doesn't. Is this a bug or simply not wanted?

the fact that drone does not implement a feature does not make it a bug. Nor does it mean the feature is not desired. In this case it just means the capability has never been discussed or requested.

I'm going to close the PR since we wouldn't merge a failing unit test

MorrisJobke:
> I'm going to close the PR since we wouldn't merge a failing unit test

Okay - Good to know. Thanks for the feedback

PR #1793, opened 2016-09-13
adding paging to get repositories from the database

### LINKED ISSUE #1776: Stash not display all repositories

thiago:
Resolve #1776.
Related PR: #1777 #1783

### DIFF
--- a/model/feed.go
+++ b/model/feed.go
@@ -8,7 +8,6 @@ type Feed struct {
 	Name     string `json:"name"          meddler:"repo_name"`
 	FullName string `json:"full_name"     meddler:"repo_full_name"`
 
-	Id       int    `json:"id,omitempty"            meddler:"build_id,zeroisnull"`
 	Number   int    `json:"number,omitempty"        meddler:"build_number,zeroisnull"`
 	Event    string `json:"event,omitempty"         meddler:"build_event,zeroisnull"`
 	Status   string `json:"status,omitempty"        meddler:"build_status,zeroisnull"`
@@ -25,18 +24,4 @@ type Feed struct {
 	Author   string `json:"author,omitempty"        meddler:"build_author,zeroisnull"`
 	Avatar   string `json:"author_avatar,omitempty" meddler:"build_avatar,zeroisnull"`
 	Email    string `json:"author_email,omitempty"  meddler:"build_email,zeroisnull"`
-}
-
-type Feeds []*Feed
-
-func (slice Feeds) Len() int {
-	return len(slice)
-}
-
-func (slice Feeds) Less(i, j int) bool {
-	return slice[i].Id < slice[j].Id;
-}
-
-func (slice Feeds) Swap(i, j int) {
-	slice[i], slice[j] = slice[j], slice[i]
 }
--- a/model/feed.go
+++ b/model/feed.go
@@ -8,6 +8,7 @@ type Feed struct {
 	Name     string `json:"name"          meddler:"repo_name"`
 	FullName string `json:"full_name"     meddler:"repo_full_name"`
 
+	Id       int    `json:"id,omitempty"            meddler:"build_id,zeroisnull"`
 	Number   int    `json:"number,omitempty"        meddler:"build_number,zeroisnull"`
 	Event    string `json:"event,omitempty"         meddler:"build_event,zeroisnull"`
 	Status   string `json:"status,omitempty"        meddler:"build_status,zeroisnull"`
@@ -24,4 +25,18 @@ type Feed struct {
 	Author   string `json:"author,omitempty"        meddler:"build_author,zeroisnull"`
 	Avatar   string `json:"author_avatar,omitempty" meddler:"build_avatar,zeroisnull"`
 	Email    string `json:"author_email,omitempty"  meddler:"build_email,zeroisnull"`
+}
+
+type Feeds []*Feed
+
+func (slice Feeds) Len() int {
+	return len(slice)
+}
+
+func (slice Feeds) Less(i, j int) bool {
+	return slice[i].Id < slice[j].Id;
+}
+
+func (slice Feeds) Swap(i, j int) {
+	slice[i], slice[j] = slice[j], slice[i]
 }
--- a/store/datastore/repos_test.go
+++ b/store/datastore/repos_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/drone/drone/model"
 	"github.com/franela/goblin"
+	"fmt"
 )
 
 func TestRepos(t *testing.T) {
@@ -136,6 +137,52 @@ func TestRepos(t *testing.T) {
 			count, err := s.GetRepoCount()
 			g.Assert(err == nil).IsTrue()
 			g.Assert(count).Equal(2)
+		})
+
+		g.It("Should Get a Repo List Paginated", func() {
+			_repos := []*model.RepoLite{
+				{FullName: "bradrydzewski/drone"},
+				{FullName: "drone/drone"},
+			}
+			_repo := []*model.Repo{
+				{
+					UserID:   1,
+					Owner:    "bradrydzewski",
+					Name:     "drone",
+					FullName: "bradrydzewski/drone",
+				},
+				{
+					UserID:   2,
+					Owner:    "drone",
+					Name:     "drone",
+					FullName: "drone/drone",
+				},
+			}
+			s.CreateRepo(_repo[0])
+			s.CreateRepo(_repo[1])
+
+			for i := 0; i < 3000; i++{
+				// find by 3000 repos
+				_repos = append(_repos, &model.RepoLite{FullName: "octocat/hello-world"+ fmt.Sprint(i)})
+				// but create only 2000 repos
+				if i >= 1999 && i < 2999{
+					continue
+				}
+				repo := &model.Repo{
+					UserID:   2,
+					Owner:    "octocat",
+					Name:     "hello-world" + fmt.Sprint(i),
+					FullName: "octocat/hello-world" + fmt.Sprint(i),
+				}
+				_repo = append(_repo, repo)
+				s.CreateRepo(repo)
+			}
+			repos, err := s.GetRepoListOf(_repos)
+			g.Assert(err == nil).IsTrue()
+			g.Assert(len(repos)).Equal(2002)
+			g.Assert(repos[0].ID).Equal(_repo[0].ID)
+			g.Assert(repos[1].ID).Equal(_repo[1].ID)
+			g.Assert(repos[2001].ID).Equal(_repo[2001].ID)
 		})
 
 		g.It("Should Delete a Repo", func() {
--- a/store/datastore/users.go
+++ b/store/datastore/users.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/drone/drone/model"
 	"github.com/russross/meddler"
+	"sort"
 )
 
 func (db *datastore) GetUser(id int64) (*model.User, error) {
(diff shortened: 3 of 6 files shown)


lu4nation:
Running it here this solution solves the 999 limit to me.

josmo:
Thanks @thiago LGTM and works for me as well.   I do think we should consider moving about from an "IN" clause and using the temp inner join like what's mentioned in http://www.xaprb.com/blog/2006/06/28/why-large-in-clauses-are-problematic/   it would probably speed things up on orgs that have large repos :)

model/feed.go
```diff
@@ -8,6 +8,7 @@ type Feed struct {
 	Name     string `json:"name"          meddler:"repo_name"`
 	FullName string `json:"full_name"     meddler:"repo_full_name"`
 
+	Id       int    `json:"id,omitempty"            meddler:"build_id,zeroisnull"`
```
bradrydzewski: perhaps we sort on `Created` instead since we already have the field available?

model/feed.go
```diff
@@ -25,3 +26,17 @@ type Feed struct {
 	Avatar   string `json:"author_avatar,omitempty" meddler:"build_avatar,zeroisnull"`
 	Email    string `json:"author_email,omitempty"  meddler:"build_email,zeroisnull"`
 }
+
+type Feeds []*Feed
+
```
bradrydzewski: can we keep this type in the `store/datastore` package since it probably only used internally in that package?

store/datastore/users.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/drone/drone/model"
 	"github.com/russross/meddler"
+	"sort"
 )
```
bradrydzewski: imports should be grouped and sorted based on Go style guide. See https://github.com/golang/go/wiki/CodeReviewComments#imports

store/datastore/users.go
```diff
@@ -28,39 +29,49 @@ func (db *datastore) GetUserList() ([]*model.User, error) {
...
+			err = meddler.QueryAll(db, &_feed, fmt.Sprintf(userFeedQuery, stmt), args...)
+			if err != nil{
+				break
+			}
+			feed = append(feed, _feed...)
+		}
 	}
-	if len(args) > 0 {
-		err = meddler.QueryAll(db, &feed, fmt.Sprintf(userFeedQuery, stmt), args...)
+	sort.Sort(sort.Reverse(feed))
```
bradrydzewski: we can probably return `feed` if `len(feed) <= 50`. The majority of drone users don't have > 999 repos so we can optimize for this case

store/datastore/utils.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/drone/drone/model"
 	"github.com/russross/meddler"
+	"math"
 )
```
bradrydzewski: import grouping / ordering

store/datastore/utils.go
```diff
@@ -38,34 +39,33 @@ func rebind(query string) string {
...
-	for i, repo := range listof {
-		qs[i] = "?"
-		in[i] = repo.FullName
-	}
-	return strings.Join(qs, ","), in
-}
+func toList(listof []*model.RepoLite) ([]string, [][]interface{}) {
+	var total = len(listof)
+	var limit = 999
+	var pagesCount = int(math.Ceil(float64(total) / float64(limit)))
```
bradrydzewski: can we avoid the float64 allocation and conversion here?

store/datastore/utils.go
```diff
@@ -38,34 +39,33 @@ func rebind(query string) string {
...
-	var qs = make([]string, size, size)
-	var in = make([]interface{}, size, size)
-	for i, repo := range listof {
-		qs[i] = "?"
-		in[i] = repo.FullName
-	}
-	return strings.Join(qs, ","), in
-}
+func toList(listof []*model.RepoLite) ([]string, [][]interface{}) {
+	var total = len(listof)
```
bradrydzewski: can we avoid returning a `[][]interface{}` and instead do something like:

```
const limit = 999

total := len(listof)
if total == 0 {
    return feed
}

pages := total / limit
if total % limit != 0 {
  pages++
}

for i:=0;i<pages;i++ {
    stmt, args := toList(listof, i*limit, limit)

    var tmpFeed []*model.Feed
    err = meddler.QueryAll(db, &tmpFeed, 
      fmt.Sprintf(userFeedQuery, stmt), args...)
    if err != nil {
      return nil, err
    }

    feed = append(feed, tmpFeed...)
}

if len(feed) <= 50 {
   return feed, nil
}

// sort

return feed[:50], nil
```

where `start` and `limit` indicate where to slice

```
func toList(listof []*model.RepoLite, start, limit int) ([]string, []interface{}) {
   if len(listof) != limit {
     listof = listof[start,limit]
   }
```

it could make sense to slice the `listof` before invoking `toList` as well ... I don't have big preference either way

bradrydzewski:
> I do think we should consider moving about from an "IN" clause and using the temp inner join like what's mentioned in

To be fair this is from 2006 which may not account for 10 years of mysql or postgres optimizations.

> it would probably speed things up on orgs that have large repos

Is there any indication that this is currently a performance bottleneck? From my testing with pretty large datasets on a relatively small database server, the IN statement still serves hundreds of feed queries per second. I can't think of any real world install that would execute hundreds of feed queries per second.

https://github.com/drone/drone/issues/1234#issuecomment-148874156

bradrydzewski:
also relevant from the comments section of that post. If you are having perf issues try running ANALYZE on the database in postres or sqlite (not sure about mysql)

> My DBA had a really hard time believing my solution in the first email in this thread should be faster. So after digging some, he ran an ANALYZE on the tables in question.
> 
> This helps the PostgreSQL planner out to create more optimal query plans. It turns out that after analyze the original query that uses an "IN" takes 1ms.
> 
> So we went from:
> IN Clause: ~4500ms
> Inner Join: ~300ms
> Analyze with IN Clause: ~1ms
> 
> The lesson is: use explain, use analyze, and consult your DBAs. They are your friends.

model/feed.go
thiago: Yes, i dont know why did it! :blush:

model/feed.go
thiago: Yes, `store/datastore/users.go` would be a right file? It only used by him...

store/datastore/users.go
thiago: Sorry, I'm newbie in go.

store/datastore/users.go
thiago: Yep, you're right!

store/datastore/utils.go
thiago: This was my first solution... It did not seem attractive for me because the limit variable and the core logic for pagination was replicated on the repos and users functions.

store/datastore/users.go
bradrydzewski: no worries :)

appreciate you taking the time to fix the issue and learn Go along the way!

store/datastore/utils.go
thiago: I'm not sure, because `10 / 3 = 3` but `10.0 / 3.0 = 3.3333`... with ceiling to 4 pages.

store/datastore/utils.go
bradrydzewski: > It did not seem attractive for me because the limit variable

this limit could be a global constant, for example:

```
const maxPlaceholder = 999
```

> core logic for pagination was replicated

I think this is OK

In some cases duplicate logic is not bad. For example, the reason that I split `toList` and `toListPostgres` into separate functions was that it was more efficient. The `toListPostgres` function has to modify the length of the query string resulting in extra allocations for string manipulation. So by combining into a single function (while well intentioned) it removed a performance optimization.

store/datastore/utils.go
```diff
@@ -38,34 +39,33 @@ func rebind(query string) string {
...
+	var limit = 999
+	var pagesCount = int(math.Ceil(float64(total) / float64(limit)))
 
-// helper function that converts a simple repository list
-// to a sql IN statement compatible with postgres.
-func toListPosgres(listof []*model.RepoLite) (string, []interface{}) {
```
bradrydzewski: we should not combine `toList` and `toListPostgres` into a single function. If you look at the two functions, `toList` is more efficient because it doesn't need to grow the size of the string. The postgres version of the function needs to grow the size of the query string to account for variable digit size in `"$" + strconv.Itoa(i+1)`

store/datastore/utils.go
bradrydzewski: > core logic for pagination was replicated

note that it could always get split out into a separate helper function:

```
func calculatePagination(total, limit int) (pages int) {
  pages = total / limit
  if total % limit != 0 {
    pages++
  }
  return
}
```

store/datastore/repos_test.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/drone/drone/model"
 	"github.com/franela/goblin"
+	"fmt"
```
bradrydzewski: import grouping and sorting :)

store/datastore/users.go
```diff
@@ -84,6 +96,20 @@ func (db *datastore) DeleteUser(user *model.User) error {
 	return err
 }
 
+type Feeds []*model.Feed
```
bradrydzewski: this can probably be a private type since it won't be exported

store/datastore/utils.go
```diff
@@ -38,34 +39,33 @@ func rebind(query string) string {
...
+		var start = page * limit
+		var end = (page * limit) + limit
+		if end > total{
+			end = total
+		}
+		var size = end - start
+		var _qs = make([]string, size, size)
+		var _in = make([]interface{}, size, size)
+		for i, repo := range listof[start:end] {
+			if meddler.Default == meddler.PostgreSQL {
```
bradrydzewski: this seems a bit inefficient since it could add thousands of extra evaluations depending on the size of the list. This was the benefit of having this split into two functions before. The prior implementation was more efficient at the expense of being dry (which was on purpose)

store/datastore/utils.go
```diff
@@ -38,34 +39,33 @@ func rebind(query string) string {
...
-	if size > 999 {
-		size = 999
-		listof = listof[:999]
-	}
-	var qs = make([]string, size, size)
-	var in = make([]interface{}, size, size)
-	for i, repo := range listof {
-		qs[i] = "$" + strconv.Itoa(i+1)
-		in[i] = repo.FullName
+	var qs = make([]string, pagesCount, pagesCount)
```
bradrydzewski: shouldn't we have `[][]string` to correspond to `[][]interface{}`? `[0][]string` may have a different length than `[1][]interface{}` right? The first page, index `[0]` could have 999 items while the second page, index `[1]` could have != 999 items and therefore needs a different placeholder lenght

bradrydzewski:
the multiple float64 allocations could be avoided with some slight re-factoring as mentioned in https://github.com/drone/drone/pull/1793#discussion_r78681342

this code correctly calculates the number of pages

```
const limit = 999

total := len(listof)
if total == 0 {
    return feed
}

pages := total / limit
if total % limit != 0 {
  pages++
}
```

to be honest I find the changes made to `toList` to be a bit complex and difficult to follow. At first glance it looks buggy, but I can't tell. I would prefer a more simple approach as noted above.

store/datastore/repos_test.go
```diff
@@ -138,6 +139,52 @@ func TestRepos(t *testing.T) {
 			g.Assert(count).Equal(2)
 		})
 
+		g.It("Should Get a Repo List Paginated", func() {
+			_repos := []*model.RepoLite{
+				{FullName: "bradrydzewski/drone"},
+				{FullName: "drone/drone"},
+			}
+			_repo := []*model.Repo{
```
bradrydzewski: perhaps `repoList` instead of `_repo` since it is an array. When I see `_repo` I think it is a single repo

store/datastore/repos_test.go
```diff
@@ -138,6 +139,52 @@ func TestRepos(t *testing.T) {
 			g.Assert(count).Equal(2)
 		})
 
+		g.It("Should Get a Repo List Paginated", func() {
+			_repos := []*model.RepoLite{
+				{FullName: "bradrydzewski/drone"},
```
bradrydzewski: the newer versions of the Go linter will complain when underscores are used in variable names, so we've tried to start removing these from our code. So perhaps `repoLiteList` instead?

thiago:
Thank you @bradrydzewski for the details in the comments and your patience. I think I made all considerations, let me know if not.

I decided to create a new [branch](https://github.com/thiago/drone/commits/bugfix/paginated_repo_queries) and perform a clear change in the master branch.

If you confirm, I'll send a new pull request and close this.

thiago:
I have not found another way to perform the tests in the paging [function GetRepoListOf without using loop](https://github.com/thiago/drone/commit/239c36434279d6241555e9f8079405f950a39fc4#diff-5a3ac9fca5d71f881e34e0cfa4930a62R164), since the limit is not configurable.

Perhaps it makes sense to make the limit a global setting as `DATABASE_DRIVER` and `DATABASE_CONFIG` but don't know exactly how do this, But I think it's better to be solved in another pull request

thiago:
Did I do something wrong? Sorry my anxiety..

bradrydzewski:
I appreciate you working through the feedback.

The current implementation still uses the following function signature:

```
func toList(listof []*model.RepoLite) ([]string, [][]interface{})
```

Which I would like to see changed per the comments I made in https://github.com/drone/drone/pull/1793#issuecomment-246909969

thiago:
Yes, after analyzing the previous reviews I thought it best to implement a new branch to make the changes clearer in the master branch. 

But I'd like you validate it before I submit a new PR, is it possible? Thanks! =D

Adron:
Just checking in on this PR. I was speaking with @josmo a few days ago about it's status and ways to test and verify the paging and everything is working. Is there anything I could prospectively do to help push the priority up on this PR? (It's one we'd get great use out of for some customers, since they're in the gazillion repo range) .  // @thiago @bradrydzewski 

Thanks all!  👍

bradrydzewski:
> Is there anything I could prospectively do to help push the priority up on this PR

you just did, thanks for the reminder ... just merged :)

bradrydzewski:
I am closing this issue with the __wont fix__ label because ultimately I don't think we are going to fully solve this problem. If someone has 50,000 github repositories Drone simply will not be able to mashup the data efficiently. I view the current 15,000 repository limit as good enough workaround for the vast majority of installations.

Thanks to @thiago and @josmo for their work on implementing the workaround.

Issue #1798, opened 2016-09-29
shm-size does not seem to be supported ?


olymk2:
tried adding shm-size: 1g to a drone yaml when the tests start the container complains about the shm size being to small, running docker or docker compose with this parameter works and the container runs, guessing this is not being passed through drone to docker compose correctly

build:
  image: alexeiled/docker-oracle-xe-11g
  pull: True
  shm-size: 1g

am i passing this into drone correctly, and should i work or is this not supported yet ?

bradrydzewski:
drone supports a subset of fields from the docker-compose file, where new fields are added on an as-needed basic. This is the list of currently supported fields:
https://github.com/drone/drone/blob/master/yaml/container.go#L58:L81

this is an example of a pull request that adds a new field:
https://github.com/drone/drone/commit/aaa702dab6858ac47a71e93c2ed6e9c54e0527be#diff-81eb6a8c464a5e832e044ddd009644ef

feel free to send a pr if it is something you would like added.

PR #1801, opened 2016-10-03
Add in docker shm-size option, mainly for oracle xe

olymk2:
Based on #1798 based on the label example i was linked to, unfortunately i don't know go and i could not get drone to build with go, from what i could tell i do not need the template ? 

Got stuck with this error before i even made changes so i could not test this.

    ~/g/s/g/d/drone> make gen
    go generate github.com/drone/drone/server/template
    bindata: Create output directory: mkdir : no such file or directory
    server/template/template.go:3: running "go-bindata": exit status 1
    Makefile:26: recipe for target 'gen_template' failed
    make: *** [gen_template] Error 1
 
as its a simple change i thought some one may be able to tell if its correct or try it, or if you know what the error above is about I would like to know as this is my first try at using the go language.

### DIFF
--- a/build/docker/util.go
+++ b/build/docker/util.go
@@ -22,6 +22,7 @@ func toContainerConfig(c *yaml.Container) *dockerclient.ContainerConfig {
 			Privileged:       c.Privileged,
 			NetworkMode:      c.Network,
 			Memory:           c.MemLimit,
+		  ShmSize:          c.ShmSize,
 			CpuShares:        c.CPUShares,
 			CpuQuota:         c.CPUQuota,
 			CpusetCpus:       c.CPUSet,
--- a/build/docker/util.go
+++ b/build/docker/util.go
@@ -22,7 +22,7 @@ func toContainerConfig(c *yaml.Container) *dockerclient.ContainerConfig {
 			Privileged:       c.Privileged,
 			NetworkMode:      c.Network,
 			Memory:           c.MemLimit,
-		  ShmSize:          c.ShmSize,
+			ShmSize:          c.ShmSize,
 			CpuShares:        c.CPUShares,
 			CpuQuota:         c.CPUQuota,
 			CpusetCpus:       c.CPUSet,
--- a/yaml/container.go
+++ b/yaml/container.go
@@ -40,6 +40,7 @@ type Container struct {
 	DNSSearch      []string
 	MemSwapLimit   int64
 	MemLimit       int64
+	ShmSize        int64
 	CPUQuota       int64
 	CPUShares      int64
 	CPUSet         string
@@ -75,6 +76,7 @@ type container struct {
 	DNSSearch      types.StringOrSlice `yaml:"dns_search"`
 	MemSwapLimit   int64               `yaml:"memswap_limit"`
 	MemLimit       int64               `yaml:"mem_limit"`
+	ShmSize        int64               `yaml:"shm_size"`
 	CPUQuota       int64               `yaml:"cpu_quota"`
 	CPUShares      int64               `yaml:"cpu_shares"`
 	CPUSet         string              `yaml:"cpuset"`
@@ -144,6 +146,7 @@ func (c *containerList) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			DNSSearch:      cc.DNSSearch.Slice(),
 			MemSwapLimit:   cc.MemSwapLimit,
 			MemLimit:       cc.MemLimit,
+			ShmSize:        cc.ShmSize,
 			CPUQuota:       cc.CPUQuota,
 			CPUShares:      cc.CPUShares,
 			CPUSet:         cc.CPUSet,
--- a/yaml/transform/validate.go
+++ b/yaml/transform/validate.go
@@ -51,6 +51,9 @@ func CheckTrusted(c *yaml.Container) error {
 	if c.Privileged {
 		return fmt.Errorf("Insufficient privileges to use privileged mode")
 	}
+	if c.ShmSize != 0 {
+		return fmt.Errorf("Insufficient privileges to override shm_size")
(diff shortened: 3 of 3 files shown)


olymk2:
might also be interested when i run these test locally they done run.


    ~/g/s/g/d/drone> drone exec 
    [DRONE] starting job #1
    [DRONE] finished job #1
    [DRONE] exit code 0
    [DRONE] job #1 passed
    [DRONE] build passed

I will try and figure out the error above and re push

tboerger:
Just look at the commands executed within the drone.yml than you will be able to build drone

olymk2:
drone.yaml looks to be doing basically the same thing, i tried it anyway and i get exactly the same error.

    ~/g/s/g/d/drone> export GO15VENDOREXPERIMENT=1
    ~/g/s/g/d/drone> make deps gen

the first step seems fine its just gen that's failing or am i missing something ?

````
go get -u golang.org/x/tools/cmd/cover
go get -u github.com/jteeuwen/go-bindata/...
go get -u github.com/elazarl/go-bindata-assetfs/...
go get -u github.com/drone/drone-ui/dist
go generate github.com/drone/drone/server/template
bindata: Create output directory: mkdir : no such file or directory
server/template/template.go:3: running "go-bindata": exit status 1
Makefile:26: recipe for target 'gen_template' failed
make: *** [gen_template] Error 1
````

not sure if the error is form go builtin the line that errors looked like it was creating a folder called files which already seems to exist.

**Comment by tboerger** [comment]
Looks like $GOPATH/bin is not part of your PATH

**Comment by olymk2** [comment]
unfortunately i don't think its that, I have these in my path
````
env | grep go
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/bin:/home/oly/go/bin
PWD=/home/oly/go/src/github.com/drone/drone
GOPATH=/home/oly/go
OLDPWD=/home/oly/go/src/github.com/drone/drone/drone

````

**Comment by tboerger** [comment]
Add it, than make gen should succeed

File: build/docker/util.go [hunk 1]
```diff
@@ -22,6 +22,7 @@ func toContainerConfig(c *yaml.Container) *dockerclient.ContainerConfig {
 			Privileged:       c.Privileged,
 			NetworkMode:      c.Network,
 			Memory:           c.MemLimit,
+		  ShmSize:          c.ShmSize,
```
bradrydzewski: just need to run `go fmt` on the code here

bradrydzewski:
one more minor addition ... can you please add check for ShmSize in this function?
https://github.com/drone/drone/blob/master/yaml/transform/validate.go#L50

```
	if c.ShmSize != 0 {
		return fmt.Errorf("Insufficient privileges to override shm_size")
	}
```

olymk2:
thanks for that I will do that as soon as I can i am away for a week so probably not till i am back.

olymk2:
okay got a few minutes so I made those changes.

tboerger:
I have not tested it but as far as I can say it looks good to me

olymk2:
Well i have now got it building, how ever i think something may have changed ?

when i run 

````/home/oly/go/src/github.com/drone/drone/release/drone exec````

it runs but does not seem to do anything at all the only output is 

````Running Matrix job #0````

in contrast if i run the old version i can see the job running and errors are reported because the database does not come up.

this is the config i am using has something changed ?
````
build:
  image: app
  extra_hosts:
    - oracle_xe:127.0.0.1
  commands:
    - cat /etc/hosts
    - ./vendor/bin/phpunit /drone/src/tests/phpunit

compose:
   oracle_xe:
      container_name: oracle_xe
      image: xe_data
      shmSize: 1g
      ports:
        - 1521:1521
      environment:
        - SERVER_ENVIRONMENT=drone
````

**Comment by bradrydzewski** [comment]
@olymk2 yes, the yaml configuration has significantly changed with 0.5. The 0.4 yaml file is not compatible with 0.5. I recommend reading the 0.5 documentation to learn more about the new format

**Comment by olymk2** [comment]
thanks for that, i have redone my config and all looks good except for my change it appears that the shm  size param is not making it to docker, will look into it a bit more when i get a chance any pointers ?

**Comment by fommil** [comment]
This looks like a really great feature and something I'm trying to mimick in my builds!

I've been manually mounting a larger `/dev/shm` and that doesn't appear to need any special permissions, so how come this feature requires the build to be trusted?

**Comment by fommil** [comment]
turns out my approach was silently failing

**Comment by fommil** [comment]
hmm, I added this setting to my build

```
```

and enabled trusted build. But if I do `df -h`, I don't see any change in `/dev/shm`

```
df -h
2s
25
Filesystem                                                                                       Size  Used Avail Use% Mounted on
2s
26
/dev/mapper/docker-8:1-4980740-abbb6856d511df404c2c29e60f68e374acccc631cce6196616c9826f93ee8f3a  9.8G  2.2G  7.1G  24% /
2s
27
tmpfs                                                                                             16G     0   16G   0% /dev
2s
28
tmpfs                                                                                             16G     0   16G   0% /sys/fs/cgroup
2s
29
/dev/sda1                                                                                        235G   15G  208G   7% /drone
2s
30
shm                                                                                               64M     0   64M   0% /dev/shm
2s
31
tmpfs                                                                                             16G     0   16G   0% /sys/firmware
```

**Comment by fommil** [comment]
if I use

```
    shmSize: 1g
```

the build just succeeds without doing anything.

olymk2:
I will have to try this myself i know passing 1g does not work and you need to specify with =

    shm_size: 1024000000

I will check next time i try to auto mate oracle testing.

fommil:
btw, I can use the privileged flag and get access to `mount`, but that seems... dangerous.

PR #1803, opened 2016-10-06
more logging in PostHook

mariusgrigoriu:
I had a user who spent a few days trying to track down why their repo wasn't built. They thought the push hook never made it to drone since there was no indication in the logs as to what went wrong. The problem was a malformed .drone.yml and the additional logging in this pull request made it visible.

We should think about how we surface events from the remote in the UI whether the .drone.yml parses or not or some of the many reasons PostHook could return early.

### DIFF
--- a/server/hook.go
+++ b/server/hook.go
@@ -30,6 +30,7 @@ func PostHook(c *gin.Context) {
 		return
 	}
 	if build == nil {
+		log.Warn("PostHook: build parsed to nil")
 		c.Writer.WriteHeader(200)
 		return
 	}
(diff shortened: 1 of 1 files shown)


server/hook.go
```diff
@@ -131,13 +132,15 @@ func PostHook(c *gin.Context) {
 	}
 	sec, err := remote_.File(user, repo, build, config.Shasum)
 	if err != nil {
-		log.Debugf("cannot find build secrets for %s. %s", repo.FullName, err)
+		log.Infof("cannot find build secrets for %s. %s", repo.FullName, err)
```
bradrydzewski: please revert to debug

bradrydzewski:
I'm not sure these changes are necessary. We should perhaps instead improve the documentation so that people know how to debug this. This is what we did for lgtm and it worked quite well for our users.

---
screen shot 2016-10-07 at 2 31 34 am

emmanuel:
I'm confused about this PR getting summarily closed. As far as I can tell, there is currently no way, short of MITM, to see the 500s that Drone is returning to the SCM server in these error cases. 

Our SCM server (Atlasssian Stash) does not report outbound web hook statuses through the UI, and users don't have self-service access to the underlying Stash logs, so rejecting the idea of Drone emitting some logs here is equivalent to advocating for silent failure behavior. 

I don't imagine that's desirable behavior for Drone, broadly, so is there some change that could make this additional logging palatable?

josmo:
@bradrydzewski Any chance we could reopen this to discuss a bit more. I totally get that in github you can see the response information for the web hook and having docs point to that is great and simple.  However for stash and other remotes there's no place to see the response and I know a few people that have gotten burnt on missing a space or the likes at least for the 500 errors.  I personally wouldn't mind having a malformed yml be a failed build itself since there is a problem with the source, but that's a bit more involved at this point since I think it would have to be passed to the agent.

derekperkins:
👍 to reopening

PR #1820, opened 2016-10-18
Update repo.go

u2takey:
typo

### DIFF
--- a/server/repo.go
+++ b/server/repo.go
@@ -100,8 +100,8 @@ func PatchRepo(c *gin.Context) {
 		Timeout     *int64 `json:"timeout,omitempty"`
 		AllowPull   *bool  `json:"allow_pr,omitempty"`
 		AllowPush   *bool  `json:"allow_push,omitempty"`
-		AllowDeploy *bool  `json:"allow_deploy,omitempty"`
-		AllowTag    *bool  `json:"allow_tag,omitempty"`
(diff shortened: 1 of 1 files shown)


tboerger:
First of all, the commit message doesn't reflect the change. The other thing is, why do you expect this is a typo?

metalmatze:
To me this seems to correct the way it is?! It's a single value and shouldn't be plural?!

bradrydzewski:
This is a breaking change and would impact the drone-ui and a number of existing integrations. I do not believe there is enough benefit to this change to justify breaking things.

u2takey:
in drone-go and  drone/client/client_impl.go  you use  model.Repo 
```
	AllowDeploy bool   `json:"allow_deploys"            meddler:"repo_allow_deploys"`
	AllowTag    bool   `json:"allow_tags"               meddler:"repo_allow_tags"`
```

but in drone/server/repo.go
```
	AllowDeploy *bool  `json:"allow_deploy,omitempty"`
	AllowTag    *bool  `json:"allow_tag,omitempty"`
```

the difference will cause bug : you cannot you cannot change AllowDeploy AllowTag config

bradrydzewski:
in this case the drone-go and CLI repository would need to change, as opposed to the server endpoint

u2takey:
ok, may change drone-go first

PR #1823, opened 2016-10-19
Add a reference to a parent build on forks / deploys

### LINKED ISSUE #1818: Expose previous build details in the model.Build{} (API)
Currently there is no way to know whether a build was a new build triggered by GitHub or it was a fork (build restart) of an existing build. The only place I managed to find the previous build number was in a forked build environment as `DRONE_PREV_BUILD_NUMBER` (https://github.com/drone/drone/blob/2d90155d5a0f1d2270a0f6ba4b112ed48b15248c/agent/agent.go#L289).

I would be great if it was possible to fetch information about the original build from the API. This information could then be surfaced in the UI as well.

Looking at the code, I see that `BuildLast` is there but only in `queue.Work{}`  (https://github.com/drone/drone/blob/2d90155d5a0f1d2270a0f6ba4b112ed48b15248c/queue/types.go#L14). Would it be possible to add some extra fields to `model.Build{}` about the original job?

vaijab:
Please note that `build.Parent` is a type of `int`, if not set, its default
value will be `0`. We don't want to have a `NULL` value in a database for
`build_parent` column. Another way of doing would to have `build.Parent` of
`interface{}` type, but that's a bit smelly.

Trigger a deployment of build number 2:
```bash
> drone-trigger --repo vaijab/test --number 2 --deploy-to some-env
Follow new build status at: http:/192.168.2.215:8000/vaijab/test/8
```

Read parent build number:
```bash
> curl -s -H "Authorization: Bearer ${DRONE_TOKEN}" ${DRONE_SERVER}/api/repos/vaijab/test/builds/8 | jq .parent
2
```

We will be making another PR to surface a href to parent build in the UI too.

Fixes #1818

### DIFF
--- a/model/build.go
+++ b/model/build.go
@@ -5,6 +5,7 @@ type Build struct {
 	ID        int64  `json:"id"            meddler:"build_id,pk"`
 	RepoID    int64  `json:"-"             meddler:"build_repo_id"`
 	Number    int    `json:"number"        meddler:"build_number"`
+	Parent    int    `json:"parent"        meddler:"build_parent"`
 	Event     string `json:"event"         meddler:"build_event"`
 	Status    string `json:"status"        meddler:"build_status"`
 	Enqueued  int64  `json:"enqueued_at"   meddler:"build_enqueued"`
--- a/server/build.go
+++ b/server/build.go
@@ -228,6 +228,7 @@ func PostBuild(c *gin.Context) {
 	if forkit, _ := strconv.ParseBool(fork); forkit {
 		build.ID = 0
 		build.Number = 0
+		build.Parent = num
 		for _, job := range jobs {
 			job.ID = 0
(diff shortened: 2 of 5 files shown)


tboerger:
If you change it to int64 the is a nullint64

vaijab:
> If you change it to int64 the is a nullint64

@tboerger yes, I could, but `sql.NullInt64` does not implement json marshal/unmarshal interfaces. I can certainly make that work, but I just I tried to keep it simple. Happy to change.

bradrydzewski:
> Please note that build.Parent is a type of int, if not set, its default value will be 0

this is the preferred behavior here. This is how we are handling all other columns, using the zero value of the field.

tboerger:
In my project I'm using gopkg.in/guregu/null.v3 :)

PR #1826, opened 2016-10-20
Implementing SkipVerify support for secrets

donny-dont:
Moves expansion of secrets to the YAML transform to properly support SkipVerify.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -94,13 +94,17 @@ func (a *Agent) prep(w *model.Work) (*yaml.Config, error) {
 	envs := toEnv(w)
 	w.Yaml = expander.ExpandString(w.Yaml, envs)
 
-	// inject the netrc file into the clone plugin if the repository is
-	// private and requires authentication.
+	// append secrets when verified or when a secret does not require
+	// verification
 	var secrets []*model.Secret
-	if w.Verified {
-		secrets = append(secrets, w.Secrets...)
+	for _, secret := range w.Secrets {
+		if w.Verified || secret.SkipVerify {
+			secrets = append(secrets, secret)
+		}
 	}
 
+	// inject the netrc file into the clone plugin if the repository is
+	// private and requires authentication.
 	if w.Repo.IsPrivate {
 		secrets = append(secrets, &model.Secret{
 			Name:   "DRONE_NETRC_USERNAME",


agent/agent.go
```diff
@@ -91,35 +91,36 @@ func (a *Agent) Run(payload *queue.Work, cancel <-chan bool) error {
...
 	w.Yaml = expander.ExpandString(w.Yaml, envs)
 
-	// inject the netrc file into the clone plugin if the repository is
-	// private and requires authentication.
+	// append all secrets. they will be checked by the ImageSecrets transform.
 	var secrets []*model.Secret
-	if w.Verified {
-		secrets = append(secrets, w.Secrets...)
-	}
+	secrets = append(secrets, w.Secrets...)
```
bradrydzewski: if we check skipverify here, we probably don't need to modify the transform function

```
	var secrets []*model.Secret

	for _, secret := range w.Secrets {
		if w.Verified || secret.SkipVerify {
				secrets = append(secrets, secret)
		}
	}
	
```

bradrydzewski: so before we added the secrets (team and repository) to the list as long as the yaml was verified:

```
var secrets []*model.Secret
if w.Verified { 
   secrets = append(secrets, w.Secrets...)
}
```

this change iterates through the secrets one by one and only adds them to the list if the yaml is verified OR the skip verify is true.

```
   var secrets []*model.Secret

    for _, secret := range w.Secrets {
        if w.Verified || secret.SkipVerify {
                secrets = append(secrets, secret)
        }
    }
```

I'm not sure we need different logic for global variables do we? If the yaml doesn't have an s3 step, and there is a global s3 secret with --image=plugins/s3, it will still be matched in the transform.

At least this is how I'm following the logic. I can take a deeper look tomorrow.

donny-dont: Ah yea that'd work too.

donny-dont:
@bradrydzewski fixed. Was apparently overthinking things 👍

agent/agent.go
tboerger: But if you define a global s3 with no verify, and a repo specific s3 with verify, and a build doesn't have a verified yaml, it will take the global one even if the credentials are not working for the project. So IMHO we should drop the following matching secrets if there is one that have been skipped.

bradrydzewski: > if you define a global s3 with no verify, and a repo specific s3 with verify, and a build doesn't have a verified yaml, it will take the global one even if the credentials are not working for the project

I think this is the desired behavior. If you setup a global secret with no verification, you should always expect it to be used without verification.

In general I would personally never use global secrets. They are a poor security choice, but I understand how they can be useful for private on-premise installations. I don't think we need to overcomplicate the implementation to provide added security to something that is always going to be insecure.

donny-dont: The behavior @tboerger is describing is also true for org secrets. If the org is skip and repo is not then the org one will be used.

If this is a problem then the secrets can be merged in the function that populates the list.

bradrydzewski:
just merged my message broker feature branch into master. looks like this has some conflicts that need to be resoled. Sorry about that :(

donny-dont:
@bradrydzewski fixed and ready to go.

PR #1827, opened 2016-10-21
Test real UTF8 in MySQL

MorrisJobke:
- change the tests to utf8mb4

cc @bradrydzewski 

I will send an PR for documentation to make setting up MySQL with utf8mb4 and drone easy.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -49,3 +49,4 @@ services:
     environment:
       - MYSQL_DATABASE=test
       - MYSQL_ALLOW_EMPTY_PASSWORD=yes
+    command: ["mysqld", "--character-set-server=utf8mb4", "--collation-server=utf8mb4_unicode_ci"]
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ gen_migrations:
 test:
 	go test -cover $(PACKAGES)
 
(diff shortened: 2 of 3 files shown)


MorrisJobke:
```
time="2016-10-21T10:00:51Z" level=error msg="Error 1071: Specified key was too long; max key length is 767 bytes" 
```

This wasn't the case here 😕 

Is this maybe because the signature is not valid and a different drone.yml is used?

bradrydzewski:
@MorrisJobke that is unrelated to the yaml signature. This error message comes from failing unit tests which were impacted by this change

MorrisJobke:
I have no idea why, but this now also happens for me. Good then I can check out what is going on

bradrydzewski:
the issue is that drone uses composite keys for unique index (ie repo_owner and repo_name) which cannot exceed a combined value of 767 bytes. When using utf8mb4 these column sizes need to be reduced to account for the extra bytes required by the encoding.

note this is for mysql only. The sqlite and postgres column sizes do not need to be adjusted

MorrisJobke:
> the issue is that drone uses composite keys for unique index (ie repo_owner and repo_name) which cannot exceed a combined value of 767 bytes. When using utf8mb4 these column sizes need to be reduced to account for the extra bytes required by the encoding.

Yes, I know - I just had no time yet to change this. I will come up with proper documentation and proper test setup here :)

PR #1841, opened 2016-11-02
Fixing yaml variable expansion

donny-dont:
YAML variables are not currently being expanded properly. Using YAML variables is especially useful for keeping notifications DRY.

This moves the custom YAML unmarshal routine to use a `map` rather than `yaml.MapSlice`. This is working around https://github.com/go-yaml/yaml/issues/184.


steve-gray:
Would this issue be preventing expansion of secrets into the 'script' commands of the drone-ssh plugin? I'm wondering because I'm unable to do anything other than hardcode them into the script or put them in the bash_profile for the remote user.

bradrydzewski:
@donny-dont confirmed this PR fixes the issue. Let me see if there is an ETA for getting it merged
https://github.com/go-yaml/yaml/pull/220

bradrydzewski:
@steve-gray this is unrelated. This is a workaround for a bug in the Go yaml parser that is preventing it from correctly implementing yaml inheritance:

```yaml
_slack: &SLACK
  image: plugins/slack

pipeline:
  notify_fail: *SLACK  # use slack variable
  notify_success:
    << : *SLACK # inherit slack variable attributes
    when:
      event: success
```

I don't want to hijack this thread, since it is unrelated to the ssh plugin, but there should be no reason to hardcode the SSH key in the Yaml when using the ssh plugin. Instead one should be using the secret store. See the docs at https://github.com/drone-plugins/drone-ssh/blob/master/DOCS.md

steve-gray:
@bradrydzewski: I'm not referring to the SSH Key. That's going in via drone secrets already. The issue I've got is that I can't get a secret value transposed into the script: property even though it's scoped to the plugins/ssh plugin. It appears this is just something that isn't possible on 0.5 as secrets never make it into the YAML sections themselves as a pre-execute pass.

PR #1843, opened 2016-11-03
Proxy values should not be made uppercase

donny-dont:
The value of `HTTP_PROXY` and `http_proxy` should be the same. URLs are actually case sensitive so this transformation should not happen. https://www.w3.org/TR/WD-html40-970708/htmlweb.html

This is actually breaking me in the docker plugin as alpine doesn't like the uppercase proxy name. Could be the case for other scenarios.


bradrydzewski:
yea, no clue why I was doing this

PR #1852, opened 2016-11-11
WIP: Check remote for org secrets permissions

jmccann:
Currently `org secrets` can only be set by a `DRONE_ADMIN`.  This PR attempts to lay the groundwork for validating permissions for a user against the remote.

Fully implemented for github remote.

Please let me know if approach looks acceptable, if I missed anything, etc ... feedback completely welcome!! 😃

### DIFF
--- a/router/middleware/session/team.go
+++ b/router/middleware/session/team.go
@@ -1,21 +1,73 @@
 package session
 
 import (
+	"github.com/drone/drone/cache"
+	"github.com/drone/drone/model"
+
+	log "github.com/Sirupsen/logrus"
 	"github.com/gin-gonic/gin"
 )
 
+func TeamPerm(c *gin.Context) *model.Perm {
+	user := User(c)
+	team := c.Param("team")
+	perm := &model.Perm{}
+
+	switch {
+		// if the user is not authenticated
+		case user == nil:
+			perm.Admin = false
+			perm.Pull  = false
+			perm.Push  = false
+
+		// if the user is a DRONE_ADMIN
+		case user.Admin:
+			perm.Admin = true
+			perm.Pull  = true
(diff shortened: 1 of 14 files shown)


tboerger:
I have not tested it but so far LGTM

jmccann:
I'm going to work on trying it out for reals next week.  If anyone else can do that too I'd appreciate it.

Thanks for all the eyeball reviews!

router/middleware/session/team.go
```diff
@@ -1,21 +1,93 @@
...
+		return nil
+	}
+	u, ok := v.(*model.Perm)
+	if !ok {
+		return nil
+	}
+	return u
+}
+
+func SetTeamPerm() gin.HandlerFunc {
```
bradrydzewski: I'm wondering if we need this function. The permissions are only ever viewed in a single place, MustTeamAdmin, so attaching more data to the context probably isn't necessary. But open to hear what everyone else thinks.

As a side note, I think if we were to use this middleware, it would need to be registered with the handler.

jmccann: I don't have any preference.  The only thing I did was copy/paste a lot of existing code and modify to suite my needs and make sure some unit testing passed so far.  😛 

Only thing I don't understand is what the following means and may not know how to do it ...

> As a side note, I think if we were to use this middleware, it would need to be registered with the handler.

jmccann:
I stopped:
> attaching more data to the context probably isn't necessary

😄 

I rebased from master also which included a caching fix I was running into.

I tested this in a dev environment with a github instance and it all seemed to work!

@bradrydzewski @tboerger @donny-dont Want to take another quick look?

bradrydzewski:
LGTM thanks for this!

PR #1853, opened 2016-11-12
fixes trailing slash on DRONE_SERVER env variable

gillesdemey:
Got stuck on this longer than I dare to admit.

Fixes accidental trailing slash in the `DRONE_SERVER` environment variable from throwing exceptions in the CLI.

If the trailing slash is present, it tries to parse the HTML of a 404 page resulting in a cryptic exception:

`invalid character '<' looking for beginning of value`

Related to https://github.com/drone/drone/pull/1756

### DIFF
--- a/drone/util.go
+++ b/drone/util.go
@@ -15,7 +15,7 @@ import (
 
 func newClient(c *cli.Context) (client.Client, error) {
 	var token = c.GlobalString("token")
-	var server = c.GlobalString("server")
+	var server = strings.TrimRight(c.GlobalString("server"), "/")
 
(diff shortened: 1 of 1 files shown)


donny-dont:
LGTM

tboerger:
We already fixed that for 0.4, have we forgot this for 0.5?

donny-dont:
@tboerger looks that way.

bradrydzewski:
@tboerger it was probably fixed in 0.4 while 0.5 development was already branched away. There was no clean merge between the two

PR #1856, opened 2016-11-16
Feature/secret concealer backend

donny-dont:
This adds `Conceal` to repo and team secret models. This will be used to conceal secrets within the build logs. Next step is to actually conceal the secrets according to this value.

@bradrydzewski @vaijab

### DIFF
--- a/drone/secret.go
+++ b/drone/secret.go
@@ -45,6 +45,10 @@ func secretAddFlags() []cli.Flag {
 			Name:  "skip-verify",
 			Usage: "skip verification for the secret",
 		},
+		cli.BoolFlag{
+			Name:  "conceal",
+			Usage: "conceal secret in build logs",
+		},
 	}
 }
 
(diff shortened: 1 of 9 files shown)


vaijab:
@donny-dont sounds like an interesting feature, but I am trying to understand when, as a user, I would want certain secrets to be concealed but not others? From my perspective, all secrets are secrets and should be concealed.

Wouldn't it make more sense to have a flag option per repo instead?

donny-dont:
@vaijab well for one there's precedence in Travis where you can either conceal or not conceal a secret.

But for me why this is really important is secrets can also be used for global configuration. Internally we used `PLUGIN_PARAMS` heavily in Drone 0.4 so global secrets allow me to emulate that behavior in 0.5. As an example I have `NPM_REGISTRY` as a secret for `plugins/npm` not because I want to hide it but because it allows me to simplify the configuration for the end user. They don't have to set the registry at all in their YAML.

vaijab:
> As an example I have NPM_REGISTRY as a secret

Yes, we use team/global secrets to set kubernetes API URLs, etc.

donny-dont:
What I'm saying is there's no reason to conceal the `NPM_REGISTRY` value. But the `NPM_PASSWORD` well that should be concealed.

patrik-gornicz:
Additionally, secrets frequently come in pairs, one a secret, one not so much. Example, USERNAME / PASSWORD, if USERNAME is drone it's concealed and that's awkward. ;-)

bradrydzewski:
LGTM

PR #1858, opened 2016-11-17
Feature/secret concealer

donny-dont:
Reverting previous implementation and moving to a separate file. Using interface to skip checks for whether a secret should be concealed.

@bradrydzewski @vaijab


vaijab:
@donny-dont can secrets be concealed globally by default still?

agent/secret.go
```diff
@@ -0,0 +1,50 @@
...
+	if len(r) > 0 {
+		replacer = &secretReplacer{
+			replacer: strings.NewReplacer(r...),
+		}
+	} else {
+		replacer = &noopReplacer{}
+	}
+
+	return replacer
+}
```
bradrydzewski: more of a styling note, but go style guidelines would say to drop the if / else and instead do the following:

```
if len(r) == 0 {
  return &noopReplacer{}
}
return replacer = &secretReplacer{
  replacer: strings.NewReplacer(r...),
}
```

donny-dont: My kingdom for `?` when writing Golang. On it now

donny-dont:
@vaijab by default secrets will not be concealed. The `--conceal` flag is used here within the Drone CLI. If you have a bunch you can probably just hit the database and apply it to all of them.

bradrydzewski:
LGTM  right now the `--concealer` flag is false by default but could always be changed to true by default using `c.BoolTrue`. I'm open to such a change in the future.

I think think we eventually want to manage secrets in the user interface, similar to Travis. Making the concealer toggle visible in the project settings should be pretty helpful as well.

PR #1859, opened 2016-11-17
Adds error message when signing the Yaml file fails

### LINKED ISSUE #1822: Drone sign feedback from CLI
Issue: Feedback from drone CLI could be improved when signing a .drone.yml

Currently: If the repo you are trying to sign isn't enabled in drone it won't give any feedback (just a blank line). Similarly it does't give any feedback when successful

Ideal: A sensible error message is presented if the repo you are trying to sign isn't enabled in drone. If it is successful it also tells you this.

Thank you for taking the time to use Drone and file an issue or feature request. Before filing an issue please ensure the following boxes are checked, if applicable:
- [x] I have searched for existing issues
- [ ] I have discussed the issue with the community at https://gitter.im/drone/drone
- [ ] I have provided a sample `.drone.yml` file to help the team reproduce
- [ ] I have provided details from the build logs
- [ ] I have provided details from the server logs by running `docker logs drone`
- [ ] I am not using the issue tracker to ask why my build failed

myth:
Fixes #1822 
* Adds error output for all HTTP response status codes when using the stream function (e.g when signing a Yaml file.), regardless of there being a response body or not.


bradrydzewski:
the sign endpoint returns valid error message in the body, so this change would in many cases replace detailed error messages with generic http error status descriptions
https://github.com/drone/drone/blob/master/server/sign.go#L12:L39

there are likely some middleware functions checking permissions and fetching repository meta-data from github that can fail, and do not write errors to the body. I think the fix to this issue would therefore be to audit these middlewares and ensure all paths write errors to the response.

myth:
Alright, I'll take a look at different middleware and see what is failing to return an informative message body.

bradrydzewski:
@myth great, thanks for looking into this!

myth:
I managed to figure out what was causing the problems. Signing with the client should produce some output on error now, even if the response body is empty. Also ensured that errors originating from the repo middleware provides some status text in the response body.

client/client_impl.go
```diff
@@ -429,7 +429,7 @@ func (c *client) open(rawurl, method string, in, out interface{}) (io.ReadCloser
 	if resp.StatusCode > http.StatusPartialContent {
 		defer resp.Body.Close()
 		out, _ := ioutil.ReadAll(resp.Body)
-		return nil, fmt.Errorf("client error %d: %s", resp.StatusCode, string(out))
+		return nil, fmt.Errorf("Client error %d: %s", resp.StatusCode, string(out))
```
tboerger: Since all messages start lowercase this have been on purpose :)

router/middleware/session/repo.go
```diff
@@ -157,11 +157,13 @@ func MustPull(c *gin.Context) {
 
 	// debugging
 	if user != nil {
-		c.AbortWithStatus(http.StatusNotFound)
+		c.String(http.StatusNotFound, http.StatusText(http.StatusNotFound))
```
tboerger: I don't think that this error message is really helpful. And supplies for the messages below

client/client_impl.go
myth: Lowercase it is then!

router/middleware/session/repo.go
myth: I don't think it's very informative either, but figured the 404 was returned to not reveal information about a private repository if it actually was missing read permissions instead of a true 404.

I can change it to reflect the same verbosity and content as the debug message below if you'd like.

tboerger: Let's see what brad prefers

myth: @bradrydzewski you got any input on this? Keep current 404 Not found scheme or use 403 with something along the lines of "you do not have permission to read/write to _repo_"?

bradrydzewski:
closing since signature was removed in 0.6+

PR #1864, opened 2016-11-19
populate github refspec

bradrydzewski:
this will populate the `refspec` field using the `<src>:<dest>` branches.

For example, lets say we have a pull request for feature branch `new-topic` that will get merged into `master`. The refspec would be stored as `new-topic:master`


bradrydzewski:
note that I think the refspec field can be used for cloning bitbucket pull requests, which do not support merge refs

```sh
# clone base
git clone https://github.com/drone-plugins/drone-slack.git
cd drone-slack

# clone head from remote repo, refspec patch-1:master
git pull https://github.com/ProLoser/drone-slack.git patch-1:master
```

PR #1865, opened 2016-11-20
add command line feature for chown api endpoint

patrickjahns:
This commit exposes the clients chown api via the command line

### DIFF
--- a/drone/repo_chown.go
+++ b/drone/repo_chown.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/codegangsta/cli"
+)
+
+var repoChownCmd = cli.Command{
+	Name:  "chown",
+	Usage: "assume ownership of a repository",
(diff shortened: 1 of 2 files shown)


drone/repo_chown.go
```diff
@@ -0,0 +1,37 @@
...
+
+	client, err := newClient(c)
+	if err != nil {
+		return err
+	}
+
+	if _, err := client.RepoChown(owner, name); err != nil {
+		return err
+	}
+	fmt.Printf("Successfully assumed ownership of repository %s/%s\n", owner, name)
```
tboerger: Do we print success messages?

bradrydzewski: I've gotten feedback that people want some sort of printed message, so I think this is good. We can eventually add `--silent` for people that want to use the CLI in scripting

bradrydzewski:
@patrickjahns thanks so much for the PR!

Issue #1367, opened 2015-12-08
Store GitHub pull request base branch


divercraig:
Add a new setting to the Repo Config page which allows users to provide a list of branches which should trigger a build on push.

By default an empty list should mean build on a push to any branch, but if the user providers a list of branches e.g. "master, dev" then builds should only be triggered when a push occurs to those branches.

This would allow developers to make multiple commits to development branches without triggering a build. Then a build would be triggered when the developer creates a pull request (if the repo was configured to run builds for pull requests). When the merge request was accepted it would be a push to master, which would trigger a build.

bradrydzewski:
@divercraig in the `.drone.yml` file you can configure a white-list list of branches that should be built. Any branch not in this list will be ignored. Hope that helps!

```
branches:
  - master
  - develop
```

divercraig:
@bradrydzewski this suggestion has the problem that it also prevents pull requests from being built.

Can we please re-open the issue?

bradrydzewski:
In this case, I think the fix is to use the `base` branch name instead of the `head` branch name for pull requests when evaluating the `branches` block in the yaml [1]

[1] https://developer.github.com/v3/activity/events/types/#webhook-payload-example-13

divercraig:
@bradrydzewski one issue with using the base branch instead of HEAD for merge requests is that this will mean evaluations like
```
when:
    branch: master
```
will be true for merge requests. How do you prevent steps like publish or deploy occurring for merge requests?

bradrydzewski:
@divercraig publish and deploy steps are always skipped for pull requests. We also support filtering based on build event type (hook, pull request, tag, deploy) that can be used to filter out pull requests:

```
when:
  event: pull_request
```

tvooo:
+1 that would be very useful

ChiperSoft:
We're also running into this problem. Every push to a feature branch triggers a build both for the branch and again for the PR. Like @divercraig I tried using the `branches` whitelist and it prevented PRs from building entirely.

bradrydzewski:
adding a quick note here since this came up recently in Gitter. Right now we are storing the source branch in the database for pull requests, which as far as I can tell, is only for UI / display purposes.

[...]

PR #1866, opened 2016-11-20
switch github pr branch from source to target

bradrydzewski:


bradrydzewski:
this fixed #1367

this PR changes the pull request branch to the target branch (github only). This is important because it allows teams to filter builds and build steps using the branch name more accurately.

this PR is now possible because we updated the UI to use refspec for pull requests, so that you can see the source and target branch. The reason we couldn't make this change sooner is because we needed a way to display the source branch in the UI, which refspec accomplishes.

donny-dont:
LGTM

vaijab:
LGTM

PR #1869, opened 2016-11-23
Add support for building inside a docker container.

STRML:
This simplifies the initial build, especially for users who want to clone & build a new binary off master or with a specific fix.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,13 @@
+FROM golang:1.6
+
+ENV GO15VENDOREXPERIMENT=1
+ENV GOPATH=/go
+ENV PATH=$PATH:$GOPATH/bin
+
+COPY . /go/src/github.com/drone/drone
+WORKDIR /go/src/github.com/drone/drone
+
+RUN make deps
+RUN make gen
+
+CMD ["make", "build"]
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,13 +1,16 @@
-FROM golang:1.6
+FROM golang:1.6-alpine
 
-ENV GO15VENDOREXPERIMENT=1
-ENV GOPATH=/go
-ENV PATH=$PATH:$GOPATH/bin
+RUN apk add --no-cache make git gcc musl-dev
 
-COPY . /go/src/github.com/drone/drone
 WORKDIR /go/src/github.com/drone/drone
 
+# Only copy what is needed to run deps & gen so these images can cache
+COPY Makefile /go/src/github.com/drone/drone/
+COPY ./server /go/src/github.com/drone/drone/server
+COPY ./store /go/src/github.com/drone/drone/store
 RUN make deps
 RUN make gen
+
+COPY . /go/src/github.com/drone/drone
 
 CMD ["make", "build"]
--- a/README.md
+++ b/README.md
@@ -32,14 +32,35 @@ Contributions, questions, and comments are welcomed and encouraged. Drone develo
 
 Please see our [installation guide](http://readme.drone.io/setup/overview) to install the official Docker image.
 
-### From Source
+#### From Source
 
-Build from source (artifacts will be in `./release`):
+Clone the repository to your Go workspace:
+
+```
+git clone git://github.com/drone/drone.git $GOPATH/src/github.com/drone/drone
+cd $GOPATH/src/github.com/drone/drone
+```
+
+Commands to build from source:
+
+```sh
+export GO15VENDOREXPERIMENT=1
+
+make deps    # Download required dependencies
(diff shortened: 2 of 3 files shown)


Dockerfile.build
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.6
+
+ENV GO15VENDOREXPERIMENT=1
```
tboerger: Drop that, it's the default on 1.6

Dockerfile.build
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.6
```
tboerger: Maybe use the alpine image?

Dockerfile.build
```diff
@@ -0,0 +1,13 @@
...
+
+ENV GO15VENDOREXPERIMENT=1
+ENV GOPATH=/go
```
tboerger: This variable is already part of the base image.

Dockerfile.build
```diff
@@ -0,0 +1,13 @@
...
+ENV PATH=$PATH:$GOPATH/bin
```
tboerger: This should also be already set

Dockerfile.build
```diff
@@ -0,0 +1,13 @@
...
+
+COPY . /go/src/github.com/drone/drone
+WORKDIR /go/src/github.com/drone/drone
```
tboerger: I would move this one line up since it's more fixed than the copy just to reuse this layer.

Dockerfile.build
```diff
@@ -0,0 +1,13 @@
...
+
+RUN make deps
```
tboerger: Since when does the go base image include make?

tboerger: The deps and gen task should also be part of the cmd.

README.md
```diff
@@ -34,21 +34,12 @@ Please see our [installation guide](http://readme.drone.io/setup/overview) to in
...
-Commands to build from source:
+Build from source (artifacts will be in `./release`):
 
 ```sh
-export GO15VENDOREXPERIMENT=1
-
-make deps    # Download required dependencies
-make gen     # Generate code
-make build   # Build the binary
+git clone git://github.com/drone/drone.git
```
tboerger: Not everybody wants to build it in docker, so extend the readme but don't drop the current guide.

Dockerfile.build
vaijab: golang image does have `make` actually.

STRML:
Thanks for the comments - will address.

Dockerfile.build
STRML: Point of separating them out was to generate intermediate images so you could re-run the build more easily (I'll have to separate the `ADD` statements to make this work properly, actually).

STRML:
Committed based on Alpine, updated README, set `.dockerignore` move to move only if it doesn't exist (so multiple failed runs will still restore the correct `.dockerignore`).

Dockerfile.build
appleboy: Why not use the latest go version 1.7.3?

bradrydzewski:
I'm wondering if a Dockerfile.build is necessary. I think one could just run the following command (or something like it) to build drone inside a container and have the artifacts output on the host machine

```
docker run --rm --volume=${pwd}:/go/src/github.com/drone/drone golang:1.7 make deps gen build
```

vaijab:
> and have the artifacts output on the host machine

That's the whole point. I also think that docs update is sufficient.

STRML:
Sharing a volume in that way won't work normally on osx/windows without
additional configuration. But the command could certainly be made shorter
if we're willing to stick with the regular go image over the alpine one.

bradrydzewski:
If you are using docker for mac (as opposed to older boot2docker) it mounts the volume from the host into the container. This command works for me on osx, compiling to `release/` on the host machine:

```
docker run \
    -w /go/src/github.com/drone/drone \
    -v $(pwd):/go/src/github.com/drone/drone \
    golang:1.6 make deps gen build
```

It is also possible to build drone using the drone command line utility. This will mount the working directory as a volume, and run the drone pipeline steps, compiling binaries to `release/`

```
drone exec
```

I definitely appreciate what this pull request is trying to do. I also see from the comments that everyone has an opinion about how this should be done. This leads me to believe that drone should provide one canonical way to build the project (Makefile) and let the individual developer modify or augment the build process with the tools and technologies of their choosing.

For these reasons I would prefer not to provide alternate methods as part of the official drone build process. I think this thread will remain useful to individuals that are looking for instructions to compile in Docker, and can use it to make informed decisions on the approach that best suits their workflow.

Issue #1863, opened 2016-11-18
Make lots of stuff pluggable


bradrydzewski:
The goal of this issue is to track things we might want to make pluggable in drone. We are getting to a point where configuration options are exploding, and people want to add new configuration options that may be more niche.

Instead of adding more and more configuration options, I'd like to explore making drone configurations scriptable. So instead of most options being provided as environment variables they would be provided as a script, or code, or something.

This could work by providing the server or agent with a directory to scripts or plugins. Drone would load all scripts in that directory, register them, and begin using them at runtime to alter or override the default behavior.

This issue will collect various use cases.

#### use Case 1

user wants to automatically add default steps to all Yaml files

#### use case 2

user wants to automatically mount a special directory only for a single plugin (cache plugin). This would allow a special cache plugin to safely work with a local directory on the host machine, without having to expose that volume to other pipeline steps, and without having to mark the build trusted

#### use case 3

user wants to automatically enable docker service containers to run in privileged mode without having to make the repository trusted. This could include adding/ dropping capabilities.

 #### use case 4

user wants a central or shared yaml file configuration that could apply to all, or a subset, of registered repositories.

#### use case 5

user wants to inject global or organization secrets using more advanced logic with data verification. For example, do not inject the global hipchat secret if the user has changed the hipchat url.

### Server Plugin, Authorizations & Scopes

admin wants to limit which repositories can be accessed in drone, and which users can activate which repositories.

## Example

This example demonstrates how one might use Docker compose to launch drone with a plugin to mutate the yaml prior to execution and add secrets (use case 5 above)

```
services:
  drone-agent:
    image: drone/drone:0.5
    command: agent
    environment:
      - DRONE_SERVER=ws://drone.company.com/ws/broker
      - DRONE_YAML_EXTENSION=http://drone-secret-injector-plugin

  drone-secret-injector-plugin:
    image: jmccann/drone-secret-injector
    environment:
      PLUGIN_CONFIG: |
        - match: plugins/docker|plugins/docker:(.+)
          environment:
            PLUGIN_USERNAME: foo
            PLUGIN_PASSWORD: bar
            PLUGIN_REGISTRY: registry.foo.com 
```

The above example assumes plugins are using some sort of RPC. The pro to using RPC is that it works with docker-compose and kubernetes without requiring a custom image.

[...]

PR #1872, opened 2016-11-25
fixes issues running services in detached mode

bradrydzewski:


bradrydzewski:
this PR enables optionally defining service containers inside the pipeline using detached mode for better control over execution order. This was previously possible, but resulted in a linter error when trying to override entrypoint or command, which this PR fixes.

```
pipeline:
  build:
    image: golang
    commands:
      - go build
  database:
    detach: true
    image: postgres
  ...
```

vaijab:
Sounds sensible. I still wish there was a way to run dind in privileged mode without trusting a repo. ;/

bradrydzewski:
@vaijab the issue is someone could `docker run` a container that mounts a volume to the parent (dind) container, and use that to escalate privilege to the host machine. Privileged containers have access to all devices on the host machine, and the ability to mount.

the good news is that running dind has definitely gotten easier over time, and I'm confident that one day we will be able to eventually run containers inside containers without escalated privileges. I believe it is already possible with LXC. So while we have these limitations today, hopefully they will be lifted in the future :)

vaijab:
> Privileged containers have access to all devices on the host machine, and the ability to mount.

Yes, for this to work, you would have to do the mounting, etc in dind, not in containers running on top of dind.

Technically it is still possible, because privileged containers have access to the underlying kernel, but if the repo is not trusted explicitly, then only the dind has access to the kernel because we're trusting dind service implicitly, which is why I think this is a much better way of doing than trusting every single repo and letting every single container to access the underlying kernel.

bradrydzewski:
@vaijab yes, this would be another good use case for pluggable yaml transformations. I've added the use case to https://github.com/drone/drone/issues/1863

Issue #1860, opened 2016-11-17
Commit author email missing in environment with bitbucket provider


patrickjahns:
The environment variable `DRONE_COMMIT_AUTHOR_EMAIL` is not provided when using bitbucket as resource provider

Simple .drone.yml to reproduce / verify
```
pipeline:
  build:
    image: golang
    commands:
      - env
```

I will try to implement this and provide a PR then

patrickjahns:
I did some investigation on this:
In the webhook payload, there is no email field present ([see here](https://confluence.atlassian.com/bitbucket/event-payloads-740262817.html)) 
However, there are two ideas for working on this:
1) fetch the details from the provided identity object
```
"actor": {
  "type": "user",
  "username": "emmap1",
  "display_name": "Emma",
  .....
}
```
more information:
https://confluence.atlassian.com/bitbucket/event-payloads-740262817.html#EventPayloads-entity_user
https://developer.atlassian.com/bitbucket/api/2/reference/resource/user/emails

2) parse the author from the raw commit
```
...
"author": {
   "raw": "Author <author@domain.tld>",
}
...

} 
```
However this would require some more parsing, to only fetch the author from the latest/correct commit
  
  
I'd prefer the first route, it seem more clean - however, would it be feasible to execute an api call/request when the webhook is triggered? 

Will think some more and see what I can do

bradrydzewski:
The second option is the best bet. The first option is not currently possible because making the extra API may require authorization credentials (for private repos) which are not available to the code parsing the hook.

patrickjahns:
I did some more investigation on this:
For `push` option 2) is possible since it exposes the author of the git commit (not the author (`actor` in bitbucket). 

When I create a `pull_request` only the `actor` objekt or the references to a `repository` (source/destination repository) are available, but no further information on the commits.

The option to create additional api calls to enrich the data from the webhook would probably the cleanest solution. I will check how fast I can get into go to see what might be possible 
Alternatively I was thinking of optionally allowing to parse the raw request body to the environment? This way any plugin could act upon the raw request body (and provide additional information to the environment?)

bradrydzewski:
> The option to create additional api calls to enrich the data from the webhook would probably the cleanest solution. I will check how fast I can get into go to see what might be possible 

the trouble is that the parser does not have a token to make additional api requests. It is a chicken and egg problem. We need to parse the hook in order to get the repository name, which would be required to lookup the token from the database.

[...]

PR #1875, opened 2016-11-25
extract email for bitbucket push requests from raw git commit

patrickjahns:
A proposed solution for the missing `DRONE_COMMIT_AUTHOR_EMAIL` as mentioned in #1860 .

This solution extracts the email from the actual git author fields since bitbucket does not expose the bitbucket account email within the webhook body. This solution only works for push requests.

### DIFF
--- a/remote/bitbucket/convert.go
+++ b/remote/bitbucket/convert.go
@@ -2,6 +2,7 @@ package bitbucket
 
 import (
 	"fmt"
+	"regexp"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


gtaylor:
@patrickjahns Looks like we're waiting on a signed CLA for you.

remote/bitbucket/convert.go
```diff
@@ -186,5 +187,18 @@ func convertPushHook(hook *internal.PushHook, change *internal.Change) *model.Bu
...
 	}
+	if len(change.New.Target.Author.Raw) != 0 {
+		build.Email = extractEmail(change.New.Target.Author.Raw)
+	}
 	return build
 }
+
+// extracts the email from a git commit author string
+func extractEmail(gitauthor string) (author string) {
+    re := regexp.MustCompile("<(.*)>")
```
bradrydzewski: this can be moved to a global variable so that we only have to parse the regular expression once. Otherwise PR looks good, thanks!

patrickjahns:
@gtaylor 
Will need to check how I can sign without adding my work email to my private account (somehow git didn't respect my author settings for this commit).

remote/bitbucket/convert.go
patrickjahns: Where should I move it then?

gtaylor:
Worst case, feel free to open a new PR if you figure out the right magical incantation on the CLI.

remote/bitbucket/convert.go
bradrydzewski: you can just put it right above the function. Please rename to `reEmail` so that we can differentiate from other global regexs, should we add them in the future. thanks :)

```diff
+var reEmail = regexp.MustCompile("<(.*)>")

// extracts the email from a git commit author string
func extractEmail(gitauthor string) (author string) {
-  re := regexp.MustCompile("<(.*)>")
    matches := reEmail.FindAllStringSubmatch(gitauthor,-1)
    if len(matches) == 1 {
        author = matches[0][1]
    }
    return
}
```

patrickjahns:
Closed in favor of #1908

PR #1885, opened 2016-12-14
added backoff to fetch yaml file

bradrydzewski:


bradrydzewski:
this is a temporary fix for GitHub returning a 404 error when attempting to retrieve the Yaml file. We believe this has something to do when hooks being sent before data is ready or propagated (maybe an eventual consistency issue?)

hopefully we'll get a permanent fix from github and can remove at a later time

PeterDaveHello:
Thank you soooooo much @bradrydzewski

PR #1887, opened 2016-12-19
Expose OAuth2 errors, avoid redirect loop.

### LINKED ISSUE #1775: expose errors in oauth2 authorization code grant
Right now when the oauth2 provider (ie Bitbucket) fails to grant an authorization code an `error` query parameter is returned, as defined in the oauth2 specification:
https://tools.ietf.org/html/rfc6749#section-4.1.2.1

Drone currently looks for the `code` and re-directs, without checking for the `error` query parameter. In error scenarios (misconfiguration for example) this can result in redirect loops, and by not catching and exposing the error it can be difficult to debug.

We should therefore start catching and returning the error. This is the block of code:
https://github.com/drone/drone/blob/master/remote/bitbucket/bitbucket.go#L46:L50

Which could be adjusted to something like this:

``` diff
package bitbucket

import (
+   "errors"
    "fmt"
    "net/http"
    "net/url"
```

``` diff
    code := r.FormValue("code")
+   if err := r.FormValue("error"); err != "" {
+       return nil, errors.New(err)
+   }
    if len(code) == 0 {
        http.Redirect(w, r, config.AuthCodeURL("drone"), http.StatusSeeOther)
        return nil, nil
    }
```

Note that this should be adjusted in both the GitHub and Bitbucket implementations.

AlekSi:
Closes #1775.

### DIFF
--- a/remote/bitbucket/bitbucket.go
+++ b/remote/bitbucket/bitbucket.go
@@ -1,7 +1,6 @@
 package bitbucket
 
 import (
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -46,15 +45,11 @@ func (c *config) Login(w http.ResponseWriter, req *http.Request) (*model.User, e
 
 	// get the OAuth errors
 	if err := req.FormValue("error"); err != "" {
-		description := req.FormValue("error_description")
-		if description != "" {
-			err += " " + description
+		return nil, &remote.AuthError{
+			Err:         err,
+			Description: req.FormValue("error_description"),
+			URI:         req.FormValue("error_uri"),
 		}
-		uri := req.FormValue("error_uri")
-		if uri != "" {
-			err += " " + uri
-		}
-		return nil, errors.New(err)
 	}
 
 	// get the OAuth code
--- a/remote/bitbucket/bitbucket.go
+++ b/remote/bitbucket/bitbucket.go
@@ -1,6 +1,7 @@
 package bitbucket
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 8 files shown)


remote/bitbucket/bitbucket.go
```diff
@@ -39,13 +40,27 @@ func New(client, secret string) remote.Remote {
...
+	if err := req.FormValue("error"); err != "" {
+		description := req.FormValue("error_description")
+		if description != "" {
+			err += " " + description
+		}
+		uri := req.FormValue("error_uri")
+		if uri != "" {
+			err += " " + uri
+		}
+		return nil, errors.New(err)
```
tboerger: Just use fmt.Error

remote/bitbucket/bitbucket.go
```diff
@@ -39,13 +40,27 @@ func New(client, secret string) remote.Remote {
 
 // Login authenticates an account with Bitbucket using the oauth2 protocol. The
 // Bitbucket account details are returned when the user is successfully authenticated.
-func (c *config) Login(w http.ResponseWriter, r *http.Request) (*model.User, error) {
-	redirect := httputil.GetURL(r)
+func (c *config) Login(w http.ResponseWriter, req *http.Request) (*model.User, error) {
```
tboerger: This change is not really related to the fix.

AlekSi: That's for style unification: other remotes use `req` for request and `r` for `*model.Repo`. Should I revert this change?

AlekSi:
> Generally I would prefer a simple error message and only log the details.

But that would introduce a dependency on logger. Is it ok?

bradrydzewski:
> Generally I would prefer a simple error message and only log the details.
> But that would introduce a dependency on logger. Is it ok?

agreed that we don't want to log in the `remote.github` package. In this case it would probably be better to create a `remote.AuthError` that holds the error and description. The logging could then be pushed to the handler.


```
if aerr, ok := err.(remote.AuthError); ok {
    log.Errorf(aerr.Desc)
}
```

but overall this is a great fix, and I really appreciate you taking the time to put this patch together. I'll get this merged this afternoon.

tboerger:
Looks good so far, but which error gets exposed to the user now?

bradrydzewski:
@tboerger right now drone does not capture an error at all, resulting in a redirect loop. See https://github.com/drone/drone/issues/1775

AlekSi:
Yeah, in my case that was due to `http` vs `https` in URL (absent `X-Forwarded-Proto` in nginx config). Wasn't easy to debug.

tboerger:
@bradrydzewski but with that pr the redirect loop should be avoided?

AlekSi:
Bump.

bradrydzewski:
wow, sorry, I thought I had merged this already. Thanks again!

PR #1895, opened 2017-01-03
Enable all webhook events by default for Gogs remote

mjwwit:
This PR enables all webhook events by default for Gogs and Gitea remotes. Since Drone now has support for all of these events, it makes sense to enable them in the webhook config.

This will save users the hassle of enabling these hooks every time they activate a repository in Drone.

### DIFF
--- a/remote/gogs/gogs.go
+++ b/remote/gogs/gogs.go
@@ -224,6 +224,7 @@ func (c *client) Activate(u *model.User, r *model.Repo, link string) error {
 	hook := gogs.CreateHookOption{
 		Type:   "gogs",
 		Config: config,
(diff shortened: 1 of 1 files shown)


bradrydzewski:
I believe `create` was explicitly excluded. In github this would trigger events for branch and tag creation, which would double-trigger for branch pushes and tag pushes. What is the behavior in Gogs?

mjwwit:
It will trigger for the creation of branches as well, but the Drone `create` hook parser will ignore everything that isn't a `tag` creation.

bradrydzewski:
ok awesome, thanks!

PR #1899, opened 2017-01-17
Socks support for drone cli 

veghead:
This patch allows the drone cli to use a SOCKS proxy as well as an HTTP proxy for issuing commands. This is useful in situations where there is no direct access to the drone server and all communication must go via an SSH tunnel on a bastion host.
simply setting the SOCKS_PROXY environment variable enables this behaviour.
e.g.
A tunnel is established thus:
`ssh bastion.host.example.com -D6969`
drone may then transparently connect to hosts on the other side of the tunnel by simply setting:
`SOCKS_PROXY=localhost:6969`
in the environment.

N.B. This patch works for us, but I'm new to drone and Go so please go easy :)


Makefile
```diff
@@ -17,6 +17,7 @@ deps_frontend:
 
 deps_backend:
 	go get -u golang.org/x/tools/cmd/cover
+	go get -u golang.org/x/net/proxy
```
tboerger: This should be added to the vendor directory

client/client_impl.go
```diff
@@ -76,9 +78,22 @@ func NewClientTokenTLS(uri, token string, c *tls.Config) Client {
...
 	if c != nil {
 		if trans, ok := auther.Transport.(*oauth2.Transport); ok {
-			trans.Base = &http.Transport{
-				TLSClientConfig: c,
-				Proxy:           http.ProxyFromEnvironment,
+			if os.Getenv("SOCKS_PROXY") != "" {
+				dialer, err := proxy.SOCKS5("tcp", os.Getenv("SOCKS_PROXY"), nil, proxy.Direct)
+				if err != nil {
+					fmt.Fprintln(os.Stderr, "can't connect to the proxy:", err)
+					os.Exit(1)
```
bradrydzewski: This should return the error. Calling `os.Exit` kills the entire program.

tboerger: So the func signature have to be changed.

veghead:
How do these changes look? As I mentioned - I'm new to Go so apologies if I've made some terrible errors here.

client/client_impl.go
veghead: Hopefully fixed in the latest push

client/client_impl.go
```diff
@@ -71,18 +73,31 @@ func NewClientToken(uri, token string) Client {
...
 	auther := config.Client(oauth2.NoContext, &oauth2.Token{AccessToken: token})
 	if c != nil {
 		if trans, ok := auther.Transport.(*oauth2.Transport); ok {
-			trans.Base = &http.Transport{
-				TLSClientConfig: c,
-				Proxy:           http.ProxyFromEnvironment,
+			if os.Getenv("SOCKS_PROXY") != "" {
+				dialer, err := proxy.SOCKS5("tcp", os.Getenv("SOCKS_PROXY"), nil, proxy.Direct)
+				if err != nil {
+					fmt.Fprintln(os.Stderr, "can't connect to the proxy:", err)
```
bradrydzewski: this printlin can be removed. The parent function that invokes and handles the error will be responsible for logging the error message

bradrydzewski:
looks great, 1 final comment and then good to merge. thanks!

client/client_impl.go
veghead: Good point. Removed.

bradrydzewski:
looks good :shipit:

PR #1908, opened 2017-01-20
Feature/bitbucket email author fixed

patrickjahns:
Copy for #1865 but with correct author field to be able to sign CLA

Original Text:

> A proposed solution for the missing DRONE_COMMIT_AUTHOR_EMAIL as mentioned in #1860 .  This solution extracts the email from the actual git author fields since bitbucket does not expose the bitbucket account email within the webhook body. This solution only works for push requests.


bradrydzewski:
thanks!

Issue #1906, opened 2017-01-20
RFC: Run all steps for all event types, by default


bradrydzewski:
The current behavior is that plugins do not execute, by default, for pull requests. In order to execute a plugin for a pull request it must be explicitly enabled:

```diff
pipeline:

  notify:
    image: plugins/slack
    channel: developers
    username: drone
   when:
     event: [ push, pull_request ]
```

This causes confusions for many drone users. There is a bit of magic under the covers where drone is automatically detecting what steps are plugins, and automatically disabling them for pull requests. The fact that this logic does not apply to build steps only adds to the confusion.

The goal, therefore, is to remove this inexplicit behavior and always execute all steps in the pipeline regardless of event type. This would eliminate the need to explicitly enable plugins for pull requests:

```diff
pipeline:

  notify:
    image: plugins/slack
    channel: developers
    username: drone
-  when:
-    event: [ push, pull_request ]
```

This means you would need to explicitly configure when you want plugins to execute, or not execute. For the record, I'm already doing this personally for more drone repositories ...

```diff
pipeline:

  notify:
    image: plugins/slack
    channel: developers
    username: drone
+  when:
+    event: push
```

The potential downside is the user could forget to configure a plugin to skip pull requests, which could have negative consequences if the plugin executes a deployment, for example.

I believe this problem is mitigated, however, by the fact that secrets are not injected for pull requests. So even though a plugin might accidentally execute for a pull request, it is unlikely it would successfully execute without secrets.

bradrydzewski:
cc @jmccann @tboerger @donny-dont @benschumacher

metalmatze:
> So even though a plugin might accidentally execute for a pull request, it is unlikely it would successfully execute without secrets.

Without that, it could be really bad. But like this I think it's ok. 👌

bradrydzewski:
I'm 50 / 50 on the change, so if there are strong feelings here we can keep the default implementation. Probably need to hear from folks using org or repo secrets

mjwwit:
Are you still making an exception for `deployment` type events, or will those be going through every pipeline step by default as well?

bradrydzewski:
@mjwwit no, this pull request removes that exception, and would require you add the following line in your yaml to skip steps locally:

```diff
pipeline:
  docker:
    image: plugins/docker
    when:
+     local: false
``` 

[...]

PR #1910, opened 2017-01-20
add when.local=false to disable plugin steps locally

bradrydzewski:


benschumacher:
Does this close #1906? Or is this just an explicit `when` parameter without a corresponding change to the default behavior?

PR #1922, opened 2017-02-01
Add support for customizing URLs for link and webhook generation.

Dirbaio:
This PR adds two new flags: --url and --hook-url. 

If --url is present, Drone will use it for generating links to itself (for webhook URLs, for links in github statuses...). If --hook-url is present, it is used instead of --url, only for hook links.

The use case for this is deploying Drone+GitHub behind a company firewall. Users access the UI on an internal-only domain. Webhooks from GitHub need to go to a public domain though, which reverse-proxies to the Drone `/hook` endpoint. Currently Drone has no way to do this, because it autodetects the private domain and uses it for setting up hooks. 

With this change, you set --url to the private domain, and --hook-url to the public domain, and everything works.

We're running our production CI with this change at @JoinVerse already, and it works fine.

### DIFF
--- a/drone/server.go
+++ b/drone/server.go
@@ -91,6 +91,18 @@ var serverCmd = cli.Command{
 			Usage:  "database driver configuration string",
 			Value:  "drone.sqlite",
 		},
+		cli.StringFlag{
+			EnvVar: "DRONE_URL",
+			Name:   "url",
+			Usage:  "url for generating links to the system. If empty, autodetect.",
+			Value:  "",
+		},
+		cli.StringFlag{
+			EnvVar: "DRONE_HOOK_URL",
+			Name:   "hook-url",
+			Usage:  "url to use for generating webhook links. If empty, use DRONE_URL or autodetect.",
+			Value:  "",
+		},
 		cli.BoolFlag{
 			EnvVar: "DRONE_GITHUB",
 			Name:   "github",
--- a/model/config.go
+++ b/model/config.go
@@ -2,12 +2,14 @@ package model
 
 // Config defines system configuration parameters.
 type Config struct {
-	Open   bool            // Enables open registration
-	Yaml   string          // Customize the Yaml configuration file name
-	Shasum string          // Customize the Yaml checksum file name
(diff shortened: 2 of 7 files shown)


bradrydzewski:
Thanks. I think this is good stuff. I do wonder if `--url` is necessary. If you access drone internally, it will use `httputil.GetURL` to determine its private url. If you are finding this doesn't work for you, perhaps you need to setup the `X-Forwarded-*` variables. See http://readme.drone.io/questions/how-to-install-behind-nginx/

Dirbaio:
It's still necessary because when receiving a hook, Drone generates a link to the UI for the github status, and GetURL would return the hook domain, not the internal one.

bradrydzewski:
>  Webhooks from GitHub need to go to a public domain though, which reverse-proxies to the Drone /hook endpoint

couldn't the proxy that receives the hook and forward to drone set the `X-Forward` headers? This would result in the status link using the correct url

sorenmat:
We have a need for this feature :)
In out setup we have 2 hostname `drone.company.com` and `drone-github.company.com`
`drone.company.com` is the main site where developers login and see their builds, this site is protected with mutual TLS.
`drone-github.company.com` is the site github should use, this is only TLS and limited to the github CIDR and the path needed for the webhook to work. At least we haven't figured out to do that with nginx or the like.

Daviey:
I had the same issue here.. I worked around it by using split horizon dns (ugh).

- From the internet, drone.company.com == Public IP address (only Github CIDR blocks has access)
- From internal resolver, drone.company.com == RFC1918 address

Dirbaio:
In our case, we recently upgraded some infra and moved to a identity-aware reverse proxy based off https://github.com/bitly/oauth2_proxy. 

We solved this issue by adding support for exempting a whitelist of paths from auth, to allow the github webhooks in. Everything lives on the same public IP and domain, so it just works and doesn't require changes from Drone's side.

Either way, this PR's code is now extremely out of date, so I'm closing it. Someone looking for this feature  would be better off coding it from scratch again.

PR #1928, opened 2017-02-04
Last known issue with the stash/bitbucket server remote. Will now not…

josmo:
… disable and override other hooks. It’ll add to the list of hooks and remove from the list of hooks on the project up to the max of 20 hooks

This should be the last "known" issue with the remote :)

### DIFF
--- a/remote/bitbucketserver/internal/client.go
+++ b/remote/bitbucketserver/internal/client.go
@@ -12,18 +12,20 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/drone/drone/model"
 	"github.com/mrjones/oauth"
+	"strings"
 )
 
 const (
-	currentUserId   = "%s/plugins/servlet/applinks/whoami"
-	pathUser        = "%s/rest/api/1.0/users/%s"
-	pathRepo        = "%s/rest/api/1.0/projects/%s/repos/%s"
-	pathRepos       = "%s/rest/api/1.0/repos?start=%s&limit=%s"
-	pathHook        = "%s/rest/api/1.0/projects/%s/repos/%s/settings/hooks/%s"
-	pathSource      = "%s/projects/%s/repos/%s/browse/%s?at=%s&raw"
-	hookName        = "com.atlassian.stash.plugin.stash-web-post-receive-hooks-plugin:postReceiveHook"
-	pathHookEnabled = "%s/rest/api/1.0/projects/%s/repos/%s/settings/hooks/%s/enabled"
-	pathStatus      = "%s/rest/build-status/1.0/commits/%s"
+	currentUserId    = "%s/plugins/servlet/applinks/whoami"
+	pathUser         = "%s/rest/api/1.0/users/%s"
+	pathRepo         = "%s/rest/api/1.0/projects/%s/repos/%s"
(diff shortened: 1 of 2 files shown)


remote/bitbucketserver/internal/client.go
```diff
@@ -214,3 +250,142 @@ func (c *Client) paginatedRepos(start int) ([]*Repo, error) {
 	}
 	return repoResponse.Values, nil
 }
+
+func Filter(vs []string, f func(string) bool) []string {
```
bradrydzewski: this should be private (ie lowercase)

remote/bitbucketserver/internal/client.go
```diff
@@ -214,3 +250,142 @@ func (c *Client) paginatedRepos(start int) ([]*Repo, error) {
...
+	for _, v := range vs {
+		if f(v) {
+			vsf = append(vsf, v)
+		}
+	}
+	return vsf
+}
+
+//TODO: find a clean way of doing these next two methods- bitbucket server hooks only support 20 cb hooks
+func arrayToHookSettings(hooks []string) HookSettings {
```
bradrydzewski: Can you provide a bit of context? The code looks a bit unorthodox, but I get the sense that this is because the Stash API is not user-friendly. So I think the code is probably fine to merge (thanks!) but I just want to understand a bit better.

josmo: It's part of the api for updating the hooks.  I wish it was better :/ I imagine there's a way to not have to list out every struct property in goLang and make it a bit less verbose(probably using introspection), I just haven't had a chance to look at it yet.

bradrydzewski: yea, I figured it had something to do with the stash api being a bit odd. I see nothing wrong with the current approach, since Go reflection might reduce the amount of code, but at the cost of readability.

another option could be just to unmarshal to a `map[string]string` so you can build the key dynamically with `map["hook-url-"+i]=...`

josmo: yeah, that's the only way I could think of doing it otherwise the readability would take a hit which it was I did it this way, until I could think of a "simpler" way.  Do you need me to remove the comments?

josmo:
@bradrydzewski Just wanted to see if you needed anything else on this one? :)

remote/bitbucketserver/internal/client.go
```diff
@@ -12,18 +12,20 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/drone/drone/model"
 	"github.com/mrjones/oauth"
+	"strings"
```
bradrydzewski: the `strings` package should be grouped with the other standard library packages. This is a recommended convention for Go projects.

```diff
import (
	"bytes"
	"encoding/json"
	"fmt"
	"io"
	"io/ioutil"
	"net/http"
	"strconv"
+	"strings"

	log "github.com/Sirupsen/logrus"
	"github.com/drone/drone/model"
	"github.com/mrjones/oauth"
-	"strings"
)
```

bradrydzewski:
I added a comment, but in retrospect, I don't want to hold up this pull request for a style change related to package ordering :)  I'll update the package ordering next time I'm in this part of the code.

josmo:
@bradrydzewski thanks for that!  if you don't get to it before me, next time I'm in there I'll re-arrange.

PR #1929, opened 2017-02-05
Remove size limit

josmo:
temp fix for #1783 and until #1793 or the alternative get pulled in. This should deal with larger number of repos on non SQLite dbs


bradrydzewski:
The size limit is necessary because sqlite cannot handle more than 999 parameters. This limitation is discussed at http://www.sqlite.org/limits.html

Therefore `toList` cannot be increased for sqlite

bradrydzewski:
Note that I'm fine with increasing for Postgres and MySQL, but we need to keep this limit in place for installations using SQLite.

josmo:
I thought by adding what you had earlier suggested of 
+	switch {
 +	case meddler.Default == meddler.SQLite && size > 999:
 +		size = 999
 +		listof = listof[:999]
 +	case size > 15000:
 +		size = 15000
 +		listof = listof[:15000]
 +	}
would only increase non SQLite.   I haven't check on SQLite but only on postgres

josmo:
I thought the last commit I had (based on a past comment ) only made the increase applicable to non SQLite but I'll double check.

josmo:
@bradrydzewski after look at it, I have it so that it will increase the limit for mySQL (toList) and postgres but not SQLite (toList makes the check).  Let me know if I'm missing something :)

store/datastore/utils.go
```diff
@@ -55,11 +59,15 @@ func toList(listof []*model.RepoLite) (string, []interface{}) {
 
 // helper function that converts a simple repository list
 // to a sql IN statement compatible with postgres.
-func toListPosgres(listof []*model.RepoLite) (string, []interface{}) {
+func toListPostgres(listof []*model.RepoLite) (string, []interface{}) {
 	var size = len(listof)
-	if size > 999 {
+	switch {
+	case meddler.Default == meddler.SQLite && size > 999:
```
bradrydzewski: I don't think this change is required because `toListPostgres` should only be invoked when `meddler.Default == meddler. PostgreSQL`. If you are seeing otherwise in the code let me know

bradrydzewski:
@josmo you are correct, looks like you've got your bases covered. I did add one last comment

store/datastore/utils.go
josmo: You're right it does get covered since it's only in the PostgreSQL portion. I updated to not check again.

josmo:
@bradrydzewski updated for your comment so it should be ready :)

Issue #1932, opened 2017-02-07
stomp client: recover panic: send on closed channel while streaming live build logs


jmccann:
Been working with @bradrydzewski on this.

I've been running into the following error randomly:
```
Feb  3 17:13:39 drone5-server drone5[9888]: time="2017-02-03T17:13:39Z" level=in
fo ip=x.x.x.x latency=1m30.248471505s method=GET path="/ws/logs/repo/name/402/1" status=200 time="2017-02-03T17:13:39Z" user-agent="Mozilla/5.0 (Ma
cintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/5
5.0.2883.95 Safari/537.36"
Feb  3 17:13:39 drone5-server drone5[9888]: 1:M 03 Feb 17:13:39.638 # stomp client: recover panic: send on closed channel
Feb  3 17:13:41 drone5-server drone5[9888]: time="2017-02-03T17:13:41Z" level=info ip=x.x.x.x latency=3.417983ms method=GET path="/api/builds" status=200 time="2017-02-03T17:13:41Z" user-agent="Go-http-client/1.1"
```

This seems to stem from writing to `logs` when it is closed @ https://github.com/drone/drone/blob/master/server/stream.go#L76

This does not happen often but as the environment has grown has started to become a daily (or more) occurrence.  When it does happen new builds are not processed and a restart of the drone server container is required to get things working again.

Brad may have more info he'd like to attach to this issue as well as I think he understands it better.

jmccann:
Want to call out https://github.com/drone/mq/pull/22 was helpful in tracking this down.

PR #1933, opened 2017-02-07
WIP: Try and avoid race condition when streaming logs

jmccann:
This is an attempt to fix/workaround issue #1932 

I think we are basically trying to better check the state of `logs` in the handler func passed to `Subscribe`.  This does seem to reduce the occurrence of the issue presented in #1932 but did not eliminate it.

So the other "workaround" implemented was to add a buffer to `logs` and not close it.  This allows it to continue to be written to while things are winding down.

Between these 2 changes I've seen no issues in my dev environment with my stress testing.  I'm going to test more in dev tonight and hopefully deploy these changes to my prod shortly to see how it behaves there.

But I wanted to get this PR out there for others to review/comment on.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -65,15 +65,31 @@ func LogStream(c *gin.Context) {
 	ticker := time.NewTicker(pingPeriod)
 	defer ticker.Stop()
 
-	logs := make(chan []byte)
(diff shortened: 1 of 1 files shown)


jmccann:
Closing in favor of #1940

Issue #1931, opened 2017-02-07
$DRONE_PULL_REQUEST not defined for MRs from non-forks


k0nserv:
Hey 👋,

We are using Gitlab and Drone 0.4. I've found from testing that when drone runs for MRs that orginated in the same repo `$DRONE_PULL_REQUEST` is not defined whereas when a build is ran for an MR from a fork it is defined. Is this expected or do we have a problem with our setup somewhere?

bradrydzewski:
The `DRONE_PULL_REQUEST` is set IF `DRONE_EVENT == pull_request` AND IF the ref includes the pull request number. The pull request number is parsed from `refs/merge-requests/%d/head` using a regular expression: https://github.com/drone/drone/blob/master/agent/agent.go#L311

I should point out that in some cases pushing to a feature branch and opening a pull request will trigger two builds. One for the push to the feature branch, and a second for the pull request. The former will never include the pull request number, the latter should. Check `DRONE_EVENT` to confirm the event type is`pull_request`.

Once you have verified the event is correct you can check and see what ref is being used. The ref isn't displayed in the user interface, but you can easily check the value using the API. You can just type `http://drone.mycompany.com/api/repos/{owner}/{name}/builds/{number}` into your browser to check. Make sure the ref looks something like refs/merge-requests/%d/head` or has a digit that can be extracted with the regexp.

If the reference does not look correct someone needs to look at the hook that GitLab sends to Drone and see what fields are available, and make sure Drone is using the correct field. If it is determined drone is sourcing the ref value from an incorrect fields, a PR should be issued to update Drone.

Please note that the GitLab implementation is community contributed and not something I personally use. So while I can definitely help point you in the right direction, it is not something I can easily debug of resolve. So community involvement here will be required.

k0nserv:
Thanks for the response @bradrydzewski the `DRONE_EVENT` is indeed `pull_request`, but the ref is a branch ref not a merge request one. Could you point me in the direction of the Gitlab implementation and I'll have a look at that?

bradrydzewski:
@k0nserv it looks like this might be the specific section in the code:
https://github.com/drone/drone/blob/master/remote/gitlab/gitlab.go#L506:L510

I recommend we check with @Bugagazavr who authored this code. The initial GitLab integration was created before merge refs existed. It is possible that, at the time of writing this code, merge refs were not available for intra-repository merge requests. Or maybe there is an edge case we don't know about.

[...]

PR #1934, opened 2017-02-08
Use refs/merge-requests

Bugagazavr:
https://github.com/drone/drone/issues/1931

Verified, gitlab creates a merge-request ref for local non-fork branches


k0nserv:
FYI from my research the required changesin GItlab seems to have been in the 8.0 release and this is already the minimum version that works with Drone

PR #1942, opened 2017-02-15
Hotfix url for "material.min.js"

magna-z:
Relative-path from repo drone/drone-ui


bradrydzewski:
material-ui was removed in master in favor of a new UI framework. We now also use a bundled file so that a CDN is no longer required.

PR #1949, opened 2017-02-23
Fix to check if the repo is enabled and don’t make a call to the sett…

josmo:
@bradrydzewski if you don't mind taking a look and merging as soon as you can.  I didn't realize that the hook worked different the first time you ever enable the repo vs once it's ever been enabled even if it's off :/  Right now you can't enable brand new repos but this fixes it.

### DIFF
--- a/remote/bitbucketserver/internal/client.go
+++ b/remote/bitbucketserver/internal/client.go
@@ -136,7 +136,7 @@ func (c *Client) CreateHook(owner string, name string, callBackLink string) erro
 	if err != nil {
 		return err
 	}
-	hooks := make([]string, 0)
+	var hooks []string
 	if hookDetails.Enabled {
 		hookSettings, err := c.GetHooks(owner, name)
 		if err != nil {
@@ -181,13 +181,9 @@ func (c *Client) GetHookDetails(owner string, name string) (*HookPluginDetails, 
 		return nil, err
 	}
 	defer response.Body.Close()
-	contents, err := ioutil.ReadAll(response.Body)
 	hookDetails := HookPluginDetails{}
-	err = json.Unmarshal(contents, &hookDetails)
-	if err != nil {
-		return nil, err
-	}
-	return &hookDetails, nil
+	err = json.NewDecoder(response.Body).Decode(&hookDetails)
+	return &hookDetails, err
 }
 
 func (c *Client) GetHooks(owner string, name string) (*HookSettings, error) {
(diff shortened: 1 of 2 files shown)


remote/bitbucketserver/internal/types.go
```diff
@@ -170,6 +170,19 @@ type RefChange struct {
...
+			Name string `json:"name"`
+			Type string `json:"type"`
+			Description string `json:"description"`
+			Version string `json:"version"`
+			ConfigFormKey string `json:"configFormKey"`
+		} `json:"details"`
+	Enabled bool `json:"enabled"`
+	Configured bool `json:"configured"`
+}
+
```
bradrydzewski: needs gofmt

remote/bitbucketserver/internal/client.go
```diff
@@ -131,15 +132,23 @@ func (c *Client) FindFileForRepo(owner string, repo string, fileName string, ref
 }
 
 func (c *Client) CreateHook(owner string, name string, callBackLink string) error {
-	hookSettings, err := c.GetHooks(owner, name)
+	hookDetails, err := c.GetHookDetails(owner, name)
 	if err != nil {
 		return err
 	}
-	hooks := hookSettingsToArray(hookSettings)
+	hooks := make([]string, 0)
```
bradrydzewski: you can declare like this
```
var hooks []string
```

remote/bitbucketserver/internal/client.go
```diff
@@ -165,6 +174,22 @@ func (c *Client) DeleteHook(owner string, name string, link string) error {
...
 }
 
+func (c *Client) GetHookDetails(owner string, name string) (*HookPluginDetails, error) {
+	urlString := fmt.Sprintf(pathHookDetails, c.base, owner, name, hookName)
+	response, err := c.client.Get(urlString)
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+	contents, err := ioutil.ReadAll(response.Body)
```
bradrydzewski: should handle this error

bradrydzewski: also you can combine reading the body and unmarshaling into a single command, which will be a bit more efficient:

```
json.NewDecoder(response.Body).Decode(&hookDetails)
```

remote/bitbucketserver/internal/client.go
```diff
@@ -165,6 +174,22 @@ func (c *Client) DeleteHook(owner string, name string, link string) error {
...
+	hookDetails := HookPluginDetails{}
+	err = json.Unmarshal(contents, &hookDetails)
+	if err != nil {
+		return nil, err
+	}
+	return &hookDetails, nil
```
bradrydzewski: in this case I would `return &hookDetails, err`. The calling function will generally check` err != nil` so there is no harm returning both

bradrydzewski:
sorry had a few more minor comments :)

josmo:
@bradrydzewski no worries and thanks for the comments.  Hopefully I'll get more consistent with my PRs :)

PR #1954, opened 2017-03-05
implement jsonrpc2 agent behind feature flag

bradrydzewski:
This can be enabled by passing the __DRONE_CANARY=true__  environment variable to the agent and the server. This will use a jsonrpc2 protocol for agent to server communication and a simplified queue and log manager, replacing the former mq package.

The purpose of this chance is improved scalability and reliability. The agents are better equipped to recover from failures and less likely to end up in an inconsistent state. This also introduces a new yaml parser and compiler, which should enable compatibility with other yaml files (compatibility with gitlab, bitbucket pipelines, travis, etc) and with multiple runners (kubernetes, docker, etc)

This alternate implementation has a number of issues and missing features.

- [x] re-enable secret concealing
- [ ] re-enable log limits
- [x] make sure all `drone_` environment variables still exist
- [x] handle yaml parsing and lint failures
- [x] extra blank lines in build output
- [ ] drone-exec not yet implemented. old code still executed
- [x] re-start existing build not yet implemented. old code still executed
- [x] cancel build is not yet hooked up
- [ ] github status link may be blank or incorrect (ask brad for more detail)

This alternate implementation does contain breaking changes:

* services are now accessed by hostname (similar to docker-compose) instead of localhost
* secrets are only interpolated

The following items need more testing:

- [ ] secrets
- [ ] service containers and linking
- [ ] restart with fork

Please note this patch __does not__ impact existing installations unless you set the __DRONE_CANARY__ flag. If you do not set this flag this patch will not impact your existing installation.


tboerger:
Awesome stuff!

bradrydzewski:
merging initial implementation behind feature flag

PR #1956, opened 2017-03-07
add ability to specify last as build number for deploy

vtolstov:

### DIFF
--- a/drone/deploy.go
+++ b/drone/deploy.go
@@ -25,6 +25,21 @@ var deployCmd = cli.Command{
 			Usage: "format output",
 			Value: tmplDeployInfo,
 		},
+		cli.StringFlag{
+			Name:  "branch",
+			Usage: "branch filter",
+			Value: "master",
+		},
+		cli.StringFlag{
+			Name:  "event",
+			Usage: "event filter",
+			Value: model.EventPush,
+		},
+		cli.StringFlag{
+			Name:  "status",
+			Usage: "status filter",
+			Value: model.StatusSuccess,
+		},
 		cli.StringSliceFlag{
 			Name:  "param, p",
 			Usage: "custom parameters to be injected into the job environment. Format: KEY=value",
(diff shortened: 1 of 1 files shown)


vtolstov:
ping...

tboerger:
LGTM

bradrydzewski:
my concern with this logic is that if the last build is event type `pull_request` it could be deployed. It is probably also desirable to allow the user to specify a branch, and if no branch is specified maybe use the default branch?

drone/deploy.go
```diff
@@ -38,16 +38,37 @@ func deploy(c *cli.Context) error {
...
+	buildArg := c.Args().Get(1)
+	var number int
+	if buildArg == "last" {
+		// Fetch the build number from the last build
+		builds, err := client.BuildList(owner, name)
+		if err != nil {
+			return err
+		}
+		for _, build := range builds {
+			if build.Status == model.StatusSuccess && build.Event != model.EventDeploy {
```
bradrydzewski: what about `pull_request` events?

bradrydzewski: also should consider teams that may want to filter out certain branches, to avoid deploying a feature branch. In the documentation, for example, we recommend the following code for launching jobs which takes branch into account:

```
drone build start --fork octocat/hello-world \
  $(drone build last --format="{{ .Number }}" --branch=master octocat/hello-world)
```

vtolstov: i don't know about pull_requests , but may be good to filter it out, because in case of pr we can take exact number...
About branch suggestion - ok

vtolstov: or may be simple add params for event type and branch ? so by default it uses only push event type, but if user wants it can be deploy event and pull request...? what do you think?

vtolstov:
@bradrydzewski i'm change params and use sane defaults. So if user does not specify anything and use last keyword, deploy uses push event, with status success and master branch.

vtolstov:
gentle ping... (sorry i need this feature and not want to keep own repo for long time..)

drone/deploy.go
```diff
@@ -25,36 +25,85 @@ var deployCmd = cli.Command{
...
+			Value: model.StatusSuccess,
+		},
 		cli.StringSliceFlag{
 			Name:  "param, p",
 			Usage: "custom parameters to be injected into the job environment. Format: KEY=value",
 		},
 	},
 }
 
+func eventContains(events []string, search string) bool {
```
bradrydzewski: is this being used?

bradrydzewski:
changes look good, just 1 question about (maybe) unused function?

drone/deploy.go
vtolstov: Thanks! old code. fixed now

vtolstov:
i'm fix issue and force pushed, so this is ready to merge! Thanks!

PR #1957, opened 2017-03-08
add last keyword for specify build number in build info

vtolstov:

### DIFF
--- a/drone/build_info.go
+++ b/drone/build_info.go
@@ -32,14 +32,26 @@ func buildInfo(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	number, err := strconv.Atoi(c.Args().Get(1))
-	if err != nil {
-		return err
-	}
+	buildArg := c.Args().Get(1)
 
 	client, err := newClient(c)
 	if err != nil {
 		return err
+	}
+
+	var number int
+	if buildArg == "last" {
+		// Fetch the build number from the last build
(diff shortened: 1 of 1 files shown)


vtolstov:
ping... @tboerger @bradrydzewski ?

tboerger:
LGTM

drone/build_info.go
```diff
@@ -32,16 +32,28 @@ func buildInfo(c *cli.Context) error {
...
+		if err != nil {
+			return err
+		}
+		number = build.Number
+	} else {
+		number, err = strconv.Atoi(buildArg)
+		if err != nil {
+			return err
+		}
+	}
```
bradrydzewski: I would recommend altering the logic slightly to only invoke `client.Build` once when `last` is the build argument. The second call is redundant.

```
var build *model.Build
if buildArg == "last" {
  build, err = client.BuildLast(owner, name, "")
  ...
} else {
  number, err = strconv.Atoi(buildArg)
  ...
  build, err = client.Build(owner, name, number)
}
```

bradrydzewski: also for consistency [1], we should let the user specify the branch using the `--branch` flag. This can default to an empty string, but if provided, can be passed to the `client.BuildLast` function

```diff
-build, err := client.BuildLast(owner, name, "")
+build, err := client.BuildLast(owner, name, c.String("branch"))
```

[1] https://github.com/drone/drone/blob/master/drone/build_last.go#L45

PR #1963, opened 2017-03-14
Direct support requests to the Discourse forum

jcarr-sailthru:
As the Gitter room advises users to go to the Discourse forum for posting
end-user support questions, best that the README points them to the right
place in the first case.


tonglil:
I think this is no longer valid.

PR #1964, opened 2017-03-14
Update error message for invalid buildArg option

iamjarvo:
Command: `drone deploy owner/repo-name build-number`
This is the current message `strconv.ParseInt: parsing "build-number": invalid syntax`. I think it is more user friendly to show this error message `Please enter a valid build number. I.E 78`

Thanks for the great tool!


bradrydzewski:
code was moved to drone/drone-cli and should be resolved now.

PR #1966, opened 2017-03-16
Adding github release deployment for cli.

Tzrlk:
If all goes as planned, the build should pick up the tars created in the 'compile' step, then upload that to the relevant github release.

See https://github.com/drone-plugins/drone-github-release/blob/master/DOCS.md for steps on adding the `GITHUB_RELEASE_API_KEY` secret.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -3,6 +3,7 @@ workspace:
   path: src/github.com/drone/drone
 
 pipeline:
+
   test:
     image: golang:1.8
     environment:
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -35,14 +35,23 @@ pipeline:
...
-    image: plugins/docker
-    repo: drone/drone
-    username: ${DOCKER_USERNAME}
-    password: ${DOCKER_PASSWORD}
-    tag: [ 0.6 ]
-    when:
-      branch: master
-      event: push
+
+    docker:
```
tboerger: Invalid indentation

tboerger:
I'm not sure if that makes sense at all. All tags/versions are uploaded to s3 so far.

paddie:
Having a versioned CLI in Github makes a lot of sense IMO. You could even use the matrix to build one for each platform (which Go makes so delightfully easy).

For most go projects I either `go get` the origin/master or head to the github releases branch to either figure out the hash of the latest stable release or download the binaries straight away.

.drone.yml
Tzrlk: What do you mean? I set the publish step indentation to one level higher than the key. The indentation on the file looks weird to me in the first-place, but I didn't want to completely re-format the file, because that's a great way of being a pain and getting changes rejected.

paddie: Ooh, I didn't know you could nest other steps under each step in the pipeline, that is neat.

Tzrlk: Ah, I was going by the 0.4 documentation instead of the newer stuff. That could really use a clarification. I'll get that fixed up.

tboerger:
@paddie we got more or less versioned CLI on S3. Hopefully brad will tag releases more often in the future :)

paddie:
Aye, that seems clear from the conversation but I always look for assets in
github first. What is the path to the S3 assets though, is it in the readme
(I'm on the phone right now)?

bradrydzewski:
I ended up moving the CLI to drone/drone-cli, but I did take the suggesting of this PR and am now publishing to github releases. Cheers

PR #1982, opened 2017-03-28
Introducing `EXTERNAL_NETWORKS` env var to connect Drone containers to additional Docker networks

ael-g:
Hi!
This patch add the ability for Drone containers to connect to existing Docker networks through an env var specified at drone server's startup: `EXTERNAL_NETWORKS`.

You can specify multiple networks by using comma as a separator: `EXTERNAL_NETWORKS=my_overlay,my_second_overlay`.

This is particularly useful if you want to use names that resolve only on a particular Docker network from inside your build containers, instead of public names/ip.

I'm using `backend.Conn` to ask Drone to connect to external networks and I'm not creating associated `backend.Network` struct. This way the patch is lighter and I don't have to make special cases to avoid Drone trying to create/delete those external networks.
Or instead, should I create `backend.Network` from the env param, and create `backend.Conn` from them in `createProcess` func, handling external and non external networks?

Is there a way I can update Drone's documentation to append this parameter?
Should I backport those changes to the `https://github.com/cncd/pipeline/` repo?

Thanks for reviewing!

### DIFF
--- a/server/hook.go
+++ b/server/hook.go
@@ -489,6 +489,7 @@ func (b *builder) Build() ([]*buildItem, error) {
 			compiler.WithProxy(),
 			// TODO ability to set global volumes for things like certs
 			compiler.WithVolumes(),
+			compiler.WithNetworks(),
 			compiler.WithWorkspaceFromURL("/drone", b.Curr.Link),
 		).Compile(parsed)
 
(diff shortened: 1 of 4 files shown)


bradrydzewski:
Thanks, can you send a pull request to https://github.com/cncd/pipeline ?

The `/vendor` directory is a dependency cache which we don't want to update manually. Instead we should update the dependencies at the source, and then I can refresh the dependency cache to include your changes.

Thanks! 😄

ael-g:
Oh I see! Doing.

PR #1985, opened 2017-03-30
[fea] native execution mode for windows and osx

olegator77:
This MR add engine for "native" execution mode for drone agent: e.g. start commands directly on host. This is useful for compiling windows application with msbuild and osx/ios application with xcodebuild.

Also native engine have very limited support for plugins - it tries to run native executable with plugin name directly on host. I've tested this with :

- drone-git - it works as native application on windows and osx
- drone-sftp-cache - it works ok as native application on osx, but requires little patch for windows

Please note, that this PR depends on PR in pipeline   https://github.com/cncd/pipeline/pull/2

### DIFF
--- a/drone/agent/agent.go
+++ b/drone/agent/agent.go
@@ -6,12 +6,14 @@ import (
 	"log"
 	"math"
 	"net/url"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/cncd/pipeline/pipeline"
 	"github.com/cncd/pipeline/pipeline/backend"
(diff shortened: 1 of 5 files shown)


bradrydzewski:
Closing this in favor of native window container support, as described here https://discourse.drone.io/t/windows-support-volunteers-wanted/2018

Support for virtual machines (i.e. not containers) would require a custom yaml format and a custom runtime. If someone is interested in spearheading such an effort, we can start a discussion in our discourse forum to gather initial feedback, and follow it up with an officail rfc https://discourse.drone.io

PR #1992, opened 2017-04-06
add registry management

bradrydzewski:


gtaylor:
What kind of registry is this?

bradrydzewski:
Docker registries. Let me explain :)

Let's say you are using private images in your build pipeline. You need a way to authenticate downloading the private image from your registry.

```
pipeline:
  build:
    image: gcr.io/app/foo
```

Currently if you want to provide these credentials to Drone you need to add specially named "magic" variables to your secret store:

```
drone secret add <repo> --image=gcr.io/* REGISTRY_USERNAME xxx
drone secret add <repo> --image=gcr.io/* REGISTRY_PASSWORD yyy
```

This specific patch logically separates secrets and registry credentials. Unlike secrets they will never be exposed to build containers and will only be used by the drone agent to pull images.

So now you can execute the following command:

```
drone registry add <repo> --hostname=gcr.io --username=xxx --password=yyy
```

Drone compares the hostname to the hostname of each image in the yaml. If the hostname matches, Drone knows to authenticate when pulling images. We use an official library for this, so the logic matches exactly what docker uses locally when deciding which cached auth credentials to use when pulling an image.

So really this is just about improved UX for pulling private images :)

gtaylor:
Very cool. This will be handy. ECR users will be happy.

themaz:
@gtaylor I don't think this'll actually help with ECR authentication. ECR requires a call to `aws ecr get-login` to actually get the (temporary) registry username/password.

gtaylor:
@themaz You can set up a cron (or Kubernetes job) to set the secret before the tokens expire. You could do that before, but this makes it more graceful.

PR #2004, opened 2017-04-14
restore queue on restart. fixes #1195

### LINKED ISSUE #1195: Recreate Queue on Restart
I'm stress test drone 0.4.0 and docker. If i forcebly restart docker and drone, build container may not be destroyed, and restart build does not work, because docker says that container with this name already exists.
I think drone needs check for container existance before creating, and destroy it when rebuild.

bradrydzewski:
Basic implementation. Not perfect. Better than what we had before.

### DIFF
--- a/model/queue.go
+++ b/model/queue.go
@@ -1,0 +1,71 @@
+package model
+
+import (
+	"context"
+
+	"github.com/cncd/queue"
+)
+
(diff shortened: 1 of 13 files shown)


bradrydzewski:
Basic implementation. Not perfect. Better than what we had before.

Issue #1021, opened 2015-05-18
Report files changed to limit testing scope


donny-dont:
Currently we have a long running test suite for a product. This gets enough commits that running the full suite would back up the build server. To limit that it would be nice to get back what files are changed.

As a contrived example usage assume that we have 3 sections to the app, A, B, and C, and each of these has a corresponding test suite. If any files within A's space change then just the A test suite needs to run.

This will allow for quicker turn around time for any functional/integration tests.

```diff
pipeline:
  frontend:
    image: node
    commands:
      - cd app
      - npm run test
+   when:
+     changeset:
+       includes: [ **.js, **.css, **.html ]
  backend:
    image: golang
    commands:
      - go build
      - go test -v
+   when:
+     changeset:
+       includes: [ **.go ]

+changeset:
+  includes: [ **.go ]
```

tonglil:
The same idea in another perspective, based on directories that have been changed, vs file extension.

bradrydzewski:
I think there two interesting use cases here, from what I can tell:

1. the ability to support mono repos, where a single repository consists of multiple loosely coupled projects. The team effectively wants to treat each sub-project separately, with its very own drone.yml file, or at a minimum its own set of steps
2. the ability to limit test execution to reduce build time for a standard repository, where the term standard is used to mean "not a mono repository". I think, in general, we can probably agree here that the goal is reducing test execution time, and limiting test execution is perhaps one solution. Another solution might be running all tests, but with parallelization to limit test execution time.

I think perhaps we need to step back and fully discuss the problems we are trying to solve. The solution (described in this issue) may very well be the correct solution. But it would help to fully understand the different problems we are trying to solve before we start writing code.

mrueg:
As it might help someone in the mean time, this is what I use as a workaround right now:
`` - if [[ "${DRONE_PULL_REQUEST}" ]]; then CHANGEDPATHS=$(git --no-pager diff --name-only FETCH_HEAD FETCH_HEAD~1); else CHANGEDPATHS=$(git --no-pager diff --name-only HEAD~1); fi ``
and then filter on the CHANGEDPATHS variable.

gtaylor:
I guess I can toss in two example problems that we have:

* A mono repo with various package manifests in it. I haven't bothered setting it up with Drone since it'd need to build every AMI unless I do some of the path stuff like @mrueg is talking about.
* Ditto for a Dockerfile repo of related Dockerfiles. 

[...]

PR #2012, opened 2017-04-27
Make the cc.xml CCMenu include a Project entry per Branch

donatj:
CCMenu xml responses are able to include multiple projects. We often have multiple branches running at the same time and right now we only get one ccmenu status back.

This leaves the existing CCProject entry and adds additional Project rows for the individual branches.

It'd be nice for me to be able to just track our `dev` and `master` branches separately, so I've altered the cc.go to allow this.

I have not yet altered the tests, I wanted to make sure there was actual interest in doing this first, but I'd certainly be happy to.

### DIFF
--- a/model/cc.go
+++ b/model/cc.go
@@ -7,8 +7,8 @@ import (
 )
 
 type CCProjects struct {
-	XMLName xml.Name   `xml:"Projects"`
-	Project *CCProject `xml:"Project"`
+	XMLName  xml.Name     `xml:"Projects"`
+	Projects []*CCProject `xml:"Project"`
 }
 
 type CCProject struct {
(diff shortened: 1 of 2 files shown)


donatj:
I would like to see this one considered. 

I corrected the major complaints about breaking backwards compatibility by making the first CCProject entry the same as the existing entry and subsequent entries individual branches. I think that should quell fears.

bradrydzewski:
Unfortunately I can't see my prior comments, and I apologize if they were vague, but in order to merge this the behavior should be opt-in, meaning we should preserve the default behavior and with an added flag or different URL we should show more branches.

For example, if `?branches=all`

bradrydzewski:
actually, we may end up adding multiple projects to a single repository [1] which could impact this feature. Should probably think through the design implications first.

[1] https://github.com/drone/drone/issues/1021#issuecomment-299603963

bradrydzewski:
closing in favor of v1.0 triggers

PR #2015, opened 2017-04-30
Fix segfault in model.Tree(procs)

mikkeloscar:
Fixes crash in model.Tree(procs) which occurs when the list of procs
passed, isn't in the expected order. E.g. if the first element doesn't
have `PPID == 0`, it would crash writing to the uninitialized
`parent.Children` slice.

### DIFF
--- a/model/proc.go
+++ b/model/proc.go
@@ -44,16 +44,17 @@ func (p *Proc) Failing() bool {
 // Tree creates a process tree from a flat process list.
 func Tree(procs []*Proc) []*Proc {
 	var (
-		nodes  []*Proc
-		parent *Proc
+		nodes    []*Proc
+		parent   *Proc
+		children []*Proc
 	)
 	for _, proc := range procs {
 		if proc.PPID == 0 {
 			nodes = append(nodes, proc)
 			parent = proc
-			continue
+			parent.Children = children
 		} else {
-			parent.Children = append(parent.Children, proc)
+			children = append(children, proc)
 		}
 	}
 	return nodes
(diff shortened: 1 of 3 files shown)


bradrydzewski:
I think this might need some unit tests, because I'm not sure it will work as desired. A process can have multiple parent nodes where ppid == 0. I don't see anything that clears the list of children. This means if we have 2 parent processes, parent process 2 will have all of parent process 1's children in the list?

I also wonder if this could be solved by sorting? The procs are created with the correct order which has me wondering if the database can return them out of order? I haven't see this issue myself, but I'm mostly tested with sqlite.

mikkeloscar:
To be completely honest I didn't fully understand the expected output of model.Tree, I just fixed it so it worked the same (as far as I could tell) without crashing.

Your explanation is helpful, is there anywhere else I can look to get a better understanding of how the tree structure should look like?

I'm using postgresql fwiw.

bradrydzewski:
@mikkeloscar will do. I should have some more time tomorrow. Maybe I can add some initial unit tests and you can build off those when implementing the fix?

mikkeloscar:
@bradrydzewski sounds good to me! :+1:

bradrydzewski:
can you test if adding an `order by proc_pid ASC` solves the issue for you?
https://github.com/drone/drone/blob/master/store/datastore/sql/postgres/sql_gen.go#L149:L166

mikkeloscar:
@bradrydzewski yes, sorting by `proc_pid` solves the issue for me. I have reverted the other commit and added the changes to the postgres query.

bradrydzewski:
awesome, thanks! much appreciated

PR #2016, opened 2017-04-30
Add --env flag to drone exec for passing custom ENV vars

mikkeloscar:
This adds the possibility to pass custom ENV vars to the build when running `drone exec`. I use the same naming as in docker `--env, -e`, but I don't mind changing it if you prefer something else.

My use case for this is that I'm running `drone exec` in my companies custom build system, and need to pass env vars like `BUILD_NUMBER` from the build system so they are available in the drone pipeline steps.

### DIFF
--- a/drone/exec/exec.go
+++ b/drone/exec/exec.go
@@ -280,7 +280,7 @@ var Command = cli.Command{
 func parseEnvVars(envs []string) (map[string]string, error) {
 	envVars := map[string]string{}
 	for _, env := range envs {
-		e := strings.Split(env, "=")
+		e := strings.SplitN(env, "=", 2)
 		if len(e) != 2 {
 			return nil, fmt.Errorf("failed to parse environment variable: %s", env)
 		}
--- a/drone/exec/exec.go
+++ b/drone/exec/exec.go
@@ -2,6 +2,7 @@ package exec
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


drone/exec/exec.go
```diff
@@ -270,6 +275,20 @@ var Command = cli.Command{
 	},
 }
 
+// parseEnvVars parses a list of envVars of the format 'NAME=value' into a
+// map[string]string.
+func parseEnvVars(envs []string) (map[string]string, error) {
+	envVars := map[string]string{}
+	for _, env := range envs {
+		e := strings.Split(env, "=")
```
tboerger: SplitN

mikkeloscar: Fixed

bradrydzewski:
If you need to pass `BUILD_NUMBER` you can do `--build-number`. The full set of fields available to the server are also available to the CLI.

Check out https://github.com/drone/drone/blob/master/drone/exec/exec.go#L143

mikkeloscar:
`BUILD_NUMBER` was just an example. In reality the variable name is different from the one provided by default in drone e.g. `CUSTOM_CI_BUILD_NUMBER`.

The point is I want to pass *custom* (and *dynamic*) ENV vars to all build steps. I don't see the problem of adding a generic `--env` flag?

bradrydzewski:
The CLI is meant to emulate the server which does not provide this capability.

You could perhaps try to hack around this limitation by declaring your custom variable as a secret. For example with the following yaml:

```
pipeline:
  foo:
    image: bar
    commands:
      - echo $CUSTOM_CI_BUILD_NUMBER
    secrets: [ CUSTOM_CI_BUILD_NUMBER ]
```

You could then do this:

```
CUSTOM_CI_BUILD_NUMBER=99 drone exec
```

mikkeloscar:
The idea is that I don't want to repeat my self when it's not necessary.

If you don't think it belongs in `drone exec` then that's alright. I'll just keep a fork with this feature.

PR #2039, opened 2017-05-12
Expose parent build number

lucaslorentz:
@bradrydzewski Expose parent build number. 

Please, let me know what else to change, I didn't grasp entirely how data flows.

Related to: https://github.com/drone/drone/pull/1930

Also, it would be very nice to do an interpolation like `${DRONE_PARENT_BUILD_NUMBER:-DRONE_BUILD_NUMBER}`, using build number as fallback when it's not a deployment event.
Does go **envsubst** supports that? If it does, maybe I can change https://github.com/cncd/pipeline/blob/master/pipeline/frontend/metadata.go#L105 to only add env if `m.Curr.Parent != 0.`

### DIFF
--- a/drone/exec/exec.go
+++ b/drone/exec/exec.go
@@ -144,6 +144,10 @@ var Command = cli.Command{
 			Name:   "build-number",
 			EnvVar: "DRONE_BUILD_NUMBER",
 		},
+		cli.IntFlag{
+			Name:   "parent-build-number",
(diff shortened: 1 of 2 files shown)


bradrydzewski:
note that this change will be available in `drone/drone:latest` and in `0.7.0` when it is released

lucaslorentz:
Thanks, I got it working using latest.

PR #2046, opened 2017-05-17
Add saving email of Bitbucket account

langovoi:
I found that email of Bitbucket user not saving, so I was implemented this functionality.

Also I have problem with tests - I don't know how to cover two cases:

1. error of [client.ListEmail()](https://github.com/langovoi/drone/blob/69b19336d80e34d1d212765cf292e08040b0b965/remote/bitbucket/bitbucket.go#L75)
1. user [without verified email](https://github.com/langovoi/drone/blob/69b19336d80e34d1d212765cf292e08040b0b965/remote/bitbucket/bitbucket.go#L79)

I think tests for this must be [here](https://github.com/langovoi/drone/blob/69b19336d80e34d1d212765cf292e08040b0b965/remote/bitbucket/bitbucket_test.go#L79).

It is my first time with Go, so if something looks wrong I will be glad to get feedback.

### DIFF
--- a/remote/bitbucket/bitbucket.go
+++ b/remote/bitbucket/bitbucket.go
@@ -69,7 +69,17 @@ func (c *config) Login(w http.ResponseWriter, req *http.Request) (*model.User, e
 	if err != nil {
 		return nil, err
 	}
-	return convertUser(curr, token), nil
+
+	emails, err := client.ListEmail()
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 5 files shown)


langovoi:
@bradrydzewski could you take a look please?

langovoi:
@tboerger @appleboy @bradrydzewski

tboerger:
I'm not following the BitBucket behavior, so I can't say if this implementation really makes sense.

PR #2049, opened 2017-05-20
update varchar to 255 for the unique value to work for the migrations…

josmo:
… table.

@bradrydzewski this ended up fixing my migration issue to current latest.  Let me know if you see any issues with this.


bradrydzewski:
thanks!

PR #2095, opened 2017-06-29
improves authorization failure logging

favadi:
* does not show secrets in log
* logs client IP


bradrydzewski:
Thanks, but this code has already been removed in 0.8 and replaced with grpc. See the feature/grpc branch.

PR #2098, opened 2017-06-30
ability to serve website from disk

bradrydzewski:


bradrydzewski:
This change allows you to serve the website from disk by setting the `DRONE_WWW` to the location of the website on disk. The can be used in conjunction with the drone ui project.

It also prototypes an interface for defining the user interface so that we can eventually make it pluggable for teams looking to replace or customize the stock UI.

PR #2108, opened 2017-07-12
add test in secret validation

hokichaio:
complete test in secret validation.

### DIFF
--- a/model/secret_test.go
+++ b/model/secret_test.go
@@ -25,10 +25,26 @@ func TestSecret(t *testing.T) {
 			secret := Secret{}
 			g.Assert(secret.Match("pull_request")).IsTrue()
 		})
-		g.It("should pass validation")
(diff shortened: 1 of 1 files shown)


model/secret_test.go
```diff
@@ -25,10 +25,26 @@ func TestSecret(t *testing.T) {
...
+		g.It("should pass validation", func() {
+			secret := Secret{}
+			secret.Name = "secretname"
+			secret.Value = "secretvalue"
+			err := secret.Validate()
+			g.Assert(err).Equal(nil)
+		})
 		g.Describe("should fail validation", func() {
-			g.It("when no image")
-			g.It("when no event")
```
hokichaio: removing this because it looks like not required from here
https://github.com/drone/drone/blob/master/model/secret.go#L47

bradrydzewski:
thanks :)

PR #2118, opened 2017-07-18
Send both build and config in SenderAllowed

oszi:
Following up on Gitter: Make SenderAllowed send both the build data and the config data to make it easy to verify the drone.yml file.

### DIFF
--- a/plugins/sender/plugin.go
+++ b/plugins/sender/plugin.go
@@ -18,7 +18,11 @@ func NewRemote(endpoint string) model.SenderService {
 
 func (p *plugin) SenderAllowed(user *model.User, repo *model.Repo, build *model.Build, conf *model.Config) (bool, error) {
 	path := fmt.Sprintf("%s/senders/%s/%s/%s/verify", p.endpoint, repo.Owner, repo.Name, build.Sender)
-	err := internal.Send("POST", path, build, nil)
+	data := map[string]interface{}{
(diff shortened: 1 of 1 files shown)


plugins/sender/plugin.go
```diff
@@ -18,7 +18,11 @@ func NewRemote(endpoint string) model.SenderService {
 
 func (p *plugin) SenderAllowed(user *model.User, repo *model.Repo, build *model.Build, conf *model.Config) (bool, error) {
 	path := fmt.Sprintf("%s/senders/%s/%s/%s/verify", p.endpoint, repo.Owner, repo.Name, build.Sender)
-	err := internal.Send("POST", path, build, nil)
+	data := map[string]interface{}{
+		"build": build,
+		"config": conf,
+	}
+	err := internal.Send("POST", path, data, nil)
 	if err != nil {
```
bradrydzewski: I think it needs to be `&data` (pointer) or the json package will return a marshal error. Sorry I didn't mention this previously

oszi: Fixed and rebased.

oszi:
Could you restart the build, please?

```
# cd /go/src/golang.org/x/net; git pull --ff-only
fatal: unable to access 'https://go.googlesource.com/net/': Failed to connect to go.googlesource.com port 443: Connection timed out
```

PR #2123, opened 2017-07-19
Escape newlines in env var substitution

execjosh:
Multiline values for env var substitution were not being escaped and could therefore cause YAML parsing to fail with something like `yaml: line 5: could not find expected ':'`.

For example, `${DRONE_COMMIT_MESSAGE}` could have more than two lines (e.g., `aaa\nbbb`).

Prior to substitution, `.drone.yml` might look like this.

```yaml
pipeline:
  xxx:
    image: scratch
    yyy: ${DRONE_COMMIT_MESSAGE}
```

After substitution, it might then look like this, and the YAML parser will fail.

```yaml
pipeline:
  xxx:
    image: scratch
    yyy: aaa
bbb
```

In order to prevent this, use `fmt.Sprintf("%q")` to escape the newlines, as well as any control characters.

### DIFF
--- a/server/hook.go
+++ b/server/hook.go
@@ -8,6 +8,7 @@ import (
 	"math/rand"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


execjosh:
This is originally based off of `v0.6.0`.  Should it instead be based off of `master`?

execjosh:
I've rebased it to `master`

execjosh:
Okay.  Now the tests pass 🎉   This commit should be cherry-pickable if you will do any back-porting to the `0.6` branch.

bradrydzewski:
thanks!

PR #2139, opened 2017-07-24
enable QUIC protocol with flag

bradrydzewski:


bradrydzewski:
question was asked why the current implementation only works with lets encrypt. This is my response and plan to expand the current functionality.

> the main reason is that drone has limited support for custom certificates (no http to https redirect, for example) which means most people probably just end up using a reverse proxy, which would not support QUIC
> 
> the lets encrypt implementation, however, is a bit more advanced. It virtually eliminates the need for a reverse proxy for ssl termination or redirect.
> 
> so I think the next step would be to improve the custom cert implementation in drone, so that it has feature parity with the lets encrypt implementation. Then we can look at integrating QUIC

Issue #2068, opened 2017-06-11
Migrate Gitlab API client to API V4


thomasf:
API V3 will be removed in GitLab 9.5, to be released on August 22, 2017.

https://docs.gitlab.com/ce/api/v3_to_v4.html

thomasf:
This is just a month away now.. 

It would be nice if maybe there was docker compose files in the contrib dicrectory (or maybe a separate drone-dev-kit or drone-gitlab-dev-kit repo) which sets up a gitlab install for local development with drone,.. Currently everyone that wants to hack on gitlab API stuff has to do it by themselves.

This is personally a much larger metal blocker for starting the work than the actual API migration work itself. From what I have seen there is no clear indication how people who develop drone "core" integration practically do that work.

Actual integration tests against gitlab would of course also be nice but that's a larger task.

vanadium23:
Hello, @bradrydzewski @thomasf.
I'd like to work on this issue, because previously contributed to gitlab API and want to see how it worked from client side view.

thomasf:
I got stuck on a few details while setting up my development environment last time I wanted to work on this so I didn't actually write any code. This should probably be fixed soonish, I might have another go at it soon again if time allows.

thomasf:
this issue closes https://github.com/drone/drone/issues/2068 

btw. Gitlab decided to keep the v3 api around a bit longerbecause they had too many v3 api requests to their site.. So latest gitlab still works with drone which is good

bradrydzewski:
closed by #2126

PR #2161, opened 2017-08-11
Update gitlab API to version v4 #2068

vanadium23:
Hello, fellows.

I've moved drone gitlab client from api/v3 to api/v4, also slightly change unit test so they've been passing, but imo we need to refactor them in the way other clients do.
I've tested it against local gitlab and gitlab.com and both works, except commit status is not changing, but it's broken from gitlab API.

### DIFF
--- a/remote/gitlab/client/project.go
+++ b/remote/gitlab/client/project.go
@@ -1,17 +1,16 @@
 package client
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"strconv"
 	"strings"
 )
 
 const (
-	searchUrl         = "/projects/search/:query"
 	projectsUrl       = "/projects"
 	projectUrl        = "/projects/:id"
-	repoUrlRawFile    = "/projects/:id/repository/blobs/:sha"
-	repoUrlRawFileRef = "/projects/:id/repository/files"
+	repoUrlRawFileRef = "/projects/:id/repository/files/:filepath"
 	commitStatusUrl   = "/projects/:id/statuses/:sha"
 )
 
@@ -47,6 +46,7 @@ func (c *Client) Projects(page int, per_page int, hide_archives bool) ([]*Projec
 	projectsOptions := QMap{
 		"page":     strconv.Itoa(page),
 		"per_page": strconv.Itoa(per_page),
+		"membership": "true",
 	}
 
 	if hide_archives {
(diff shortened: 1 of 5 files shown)


Bugagazavr:
👍 Good job

bradrydzewski:
thanks for this!  so it sounds like status API is broken on the gitlab end? can you provide some more details so we can document as a known issue and maybe link to the gitlab issue?

vanadium23:
@bradrydzewski don't know what is wrong. According to [gitlab docs](https://docs.gitlab.com/ce/api/commits.html#post-the-build-status-to-a-commit), the implementation is correct. Also I've rebased code onto master. I can deal with commit status in separate issue, if you want (:

bradrydzewski:
@vanadium23 maybe check to make sure you have DRONE_HOST configured properly? If not properly configured it can impact the ability to create statuses.

amq:
Adding the scheme to `DRONE_HOST` solved the status updates for me.

vanadium23:
@bradrydzewski, yes, problem was in the scheme, because I set it as localhost (:

remote/gitlab/client/project.go
```diff
@@ -1,17 +1,16 @@
 package client
 
 import (
+	b64 "encoding/base64"
```
bradrydzewski: remove alias and use `base64` package name below

vanadium23: Done.

thomasf:
this issue closes https://github.com/drone/drone/issues/2068   (don't know why github doesnt link these issues when the issue # is in the title)

btw. Gitlab decided to keep the v3 api around a bit longerbecause they had too many v3 api requests to their site.. So latest gitlab still works with drone which is good

vanadium23:
@bradrydzewski everything is fine?

bradrydzewski:
ok, sorry for the delays. thanks again for this!  I plan to snapshot rc.4 on monday for those waiting for a tagged release.

Spedge:
So we've been working on upgrading our Drone version to 0.8 - installing and delivering on rc3. rc4 is released and all of a sudden doesn't work with our supported version of GitLab? So to use Drone, we now need to do a major version upgrade of GitLab? 

As @thomasf suggests - v3 is still around. By migrating to v4 and not supporting any backward compatibility, you are really making this difficult for us.

thomasf:
I don't think it's unfair to require the  v4 API,  the v3 API will not be maintained anymore (whatever that means). Gitlab announced that v3 was going away like a year before it happened or something and v3/v4 has been working in parallell in Gitlab for a long time giving people time to migrate.

Spedge:
If we wrote some code to allow us to add a flag to the startup in order to use the v3 version, would you accept it?

bradrydzewski:
@Spedge no worries. I just merged https://github.com/drone/drone/pull/2215

I ended up just copying the old gitlab code into a gitlab3 folder so that we have old and new code in sibling packages. This way I didn't have to make any major code changes in drone.

```diff
remote/gitlab
+remote/gitlab3
```

I thine added an environment variable that you can use to switch:

```
DRONE_GITLAB_V3_API=true
```

This change is available in the `drone/drone:latest` image. Would it be possible to upgrade your server and verify this change by end of day tomorrow? I only ask because I am planning on snapshotting a final release on Monday and would like to make sure there are no critical issues.

gjtempleton:
I've just tried out the `:latest` tag with `DRONE_GITLAB_V3_API=true
` set and successfully triggered builds and had the status reported back to Gitlab successfully. Thanks for the swift response.

bradrydzewski:
Thanks Guy for the quick reply, happy to help!

Spedge:
Brilliant, thanks Brad - really, really appreciate the quick turnaround on this.

Issue #2057, opened 2017-05-25
[Feature Request] Be able to set a starting build number


josmo:
It would be great to be able to set the starting build number. 

Use case:
When you migrate from another build system or even upgrade drone sometimes you want to set the build number to the last of the other system.

Seems like in the settings or enable location we could add a starting build number.

@bradrydzewski If you have any ideas around what's the best way to handle this, I'm down to take a stab at it.

bradrydzewski:
You can do `update repos set repo_counter = ? where repo_id = ?`.  You could use a sql script to set the counter in batch for all repositories. Example script:

```text
cat <<EOF > setup.sql

update repos set repo_counter = ? where repo_full_name = ?;
update repos set repo_counter = ? where repo_full_name = ?;
update repos set repo_counter = ? where repo_full_name = ?;
update repos set repo_counter = ? where repo_full_name = ?;

EOF
```

Example command to load the script (sqlite):

```text
> sqlite3 /path/to/drone.sql
> sqlite> .read setup.sql
```

josmo:
ah yeah totally works. I was more meaning getting a UI component for folks to use and be able to set the next build counter :)

bradrydzewski:
I would suggest we add this to the `drone repo update` CLI command. These sort of features get incubated in the CLI (see `drone repo repair` and `drone repo chown`) before we consider adding them to the user interface.

Setting the next build number definitely has some risks, so I think we should probably provide a warning message that we force the user to acknowledge:

```
$ drone repo update foo/bar --build-counter=25
Setting the build counter is an unsafe operation that could put your
repository in an inconsistent state. Please use --unsafe to proceed

$ drone repo update foo/bar --build-counter=25 --unsafe
```

We may also want to consider, at least initially, restricting the ability to set this field to system administrators (similar to what we do for timeout and trusted). We can always loosen up the restrictions in the future, but perhaps a conservative approach here makes sense to start?

tonglil:
I think the UI component could only increment after a build has started right? Because the next build has to increment the build counter. If you change it to anything less, what implication does that have on the build that just ran?

josmo:
@tonglil right now the only way to do it is once the PR are merged and via the command line. The idea is that the admin should make sure manually to only update to a higher number since there's no verification in the code. Hence why you need the unsafe flag that Brad mentioned should be there :)  This honestly solves most problems I've seen but we could open a new feature request once it's all merged in if it makes sense to be in the UI

[...]

PR #2163, opened 2017-08-13
allow in a patch update to change the build number if the user is an …

josmo:
…admin

Server side update for the cli https://github.com/drone/drone-cli/pull/60 to be able to change the build-counter for issue https://github.com/drone/drone/issues/2057 @bradrydzewski I'm not really able to build drone anymore with the enterprise changes and what not so I haven't been able to test this part of the code :( but from what I can tell it's the only updates that are needed to make the cli updates work.

### DIFF
--- a/model/repo.go
+++ b/model/repo.go
@@ -57,13 +57,14 @@ func (r *Repo) Update(from *Repo) {
 
 // RepoPatch represents a repository patch object.
 type RepoPatch struct {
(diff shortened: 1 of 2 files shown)


josmo:
actually I was able to hack a build together and it works :)

bradrydzewski:
fyi, you can now build with the standard go command toolchain

```
go get github.com/drone/drone/...
```

Issue #2167, opened 2017-08-16
[Feature Request] Ability to change the name of a current project/repo


josmo:
Allow changing the name of an org or repo name.

This seems like a good candidate for the cli. Something allow the lines of
```
drone repo update --full-repo-name=newproj/supercat1 proj/supercat
```
would move the repo to the new name.

@bradrydzewski I'm not sure if this is really feasible but I'm willing to give it a go.
from what I can tell critical things would include

in the repo table - repo_owner, repo_name, repo_full_name
  *unknowns- repo_link, repo_clone
in the builds table -
  * unknowns if it even matters but build_link, build_remote (I doubt these would matter nearly as much as the ones in the repo table)

josmo:
is that something the repair hook would handle?

bradrydzewski:
drone now stores the unique repository ID (from github, bitbucket, etc) in the database so that it can automatically detect repository name changes and update accordingly.

PR #2168, opened 2017-08-16
Adding the ability to update the repo owner and name via patch with t…

josmo:
…he owner and name props

@bradrydzewski It looks like I've got everything working from being able to patch. I'll add a cli PR that's missing the model till that gets merged that way if you have any comments I can address it right.

The cli has the following warning when not unsafe of "validate updating the full repo name is unsafe - build status in repos will reflect old name"
since that seems to be the only side affect.  I also have the cli repair the hooks after it renames it.

for https://github.com/drone/drone/issues/2167

### DIFF
--- a/model/repo.go
+++ b/model/repo.go
@@ -67,4 +67,6 @@ type RepoPatch struct {
 	AllowDeploy  *bool   `json:"allow_deploy,omitempty"`
 	AllowTag     *bool   `json:"allow_tag,omitempty"`
 	BuildCounter *int    `json:"build_counter,omitempty"`
+	Name         *string `json:"name",omitempty`
+	Owner        *string `json:"owner,omitempty"`
 }
--- a/model/repo.go
+++ b/model/repo.go
@@ -1,5 +1,10 @@
 package model
 
+import (
+	"strings"
+	"fmt"
+)
+
 type RepoLite struct {
 	Owner    string `json:"owner"`
 	Name     string `json:"name"`
@@ -36,6 +41,25 @@ type Repo struct {
 	Config      string `json:"config_file"              meddler:"repo_config_path"`
 	Hash        string `json:"-"                        meddler:"repo_hash"`
 	Perm        *Perm  `json:"-"                        meddler:"-"`
+}
+
+func (r *Repo) ResetVisibility() {
+	r.Visibility = VisibilityPublic
+	if r.IsPrivate {
+		r.Visibility = VisibilityPrivate
+	}
+}
+
+// ParseRepo parses the repository owner and name from a string.
+func ParseRepo(str string) (user, repo string, err error) {
+	var parts = strings.Split(str, "/")
+	if len(parts) != 2 {
+		err = fmt.Errorf("Error: Invalid or missing repository. eg octocat/hello-world.")
+		return
+	}
+	user = parts[0]
+	repo = parts[1]
+	return
 }
 
 // Update updates the repository with values from the given Repo.
--- a/model/repo.go
+++ b/model/repo.go
@@ -67,6 +67,4 @@ type RepoPatch struct {
 	AllowDeploy  *bool   `json:"allow_deploy,omitempty"`
 	AllowTag     *bool   `json:"allow_tag,omitempty"`
 	BuildCounter *int    `json:"build_counter,omitempty"`
-	Name         *string `json:"name,omitempty"`
-	Owner        *string `json:"owner,omitempty"`
 }
--- a/model/repo.go
+++ b/model/repo.go
@@ -67,6 +67,6 @@ type RepoPatch struct {
 	AllowDeploy  *bool   `json:"allow_deploy,omitempty"`
 	AllowTag     *bool   `json:"allow_tag,omitempty"`
 	BuildCounter *int    `json:"build_counter,omitempty"`
-	Name         *string `json:"name",omitempty`
+	Name         *string `json:"name,omitempty"`
 	Owner        *string `json:"owner,omitempty"`
 }
--- a/server/repo.go
+++ b/server/repo.go
@@ -15,7 +15,6 @@ import (
 	"github.com/drone/drone/shared/httputil"
 	"github.com/drone/drone/shared/token"
 	"github.com/drone/drone/store"
-	"strings"
 )
 
 func PostRepo(c *gin.Context) {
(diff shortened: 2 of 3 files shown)


model/repo.go
```diff
@@ -67,4 +67,6 @@ type RepoPatch struct {
 	AllowDeploy  *bool   `json:"allow_deploy,omitempty"`
 	AllowTag     *bool   `json:"allow_tag,omitempty"`
 	BuildCounter *int    `json:"build_counter,omitempty"`
+	Name         *string `json:"name",omitempty`
```
mrueg: not sure if the quoting is right here: shouldn't it be "name,omitempty"?

josmo: You're right it should be "name,omitempty" I'm not sure why it worked while testing but I'll fix and push

server/repo.go
```diff
@@ -137,6 +138,19 @@ func PatchRepo(c *gin.Context) {
...
+		from, err := remote.Repo(user, *in.Owner, *in.Name)
+		if err != nil {
+			c.AbortWithError(http.StatusInternalServerError, err)
+		}
+		repo.Name = from.Name
+		repo.Owner = from.Owner
+		repo.FullName = from.FullName
+		repo.Avatar = from.Avatar
+		repo.Link = from.Link
+		repo.Clone = from.Clone
```
bradrydzewski: maybe also update `private` and `visibility`

josmo: Ok I'll totally add that.  Right now I have the CLI run a repair after the update which seems to work seamlessly from the user perspective. At least with what I've tried it works.

josmo:
@bradrydzewski Let me know what you lean towards. Right now in the CLI I'm doing a update and then repair as part of one command, but I guess both of those could take place on the server as well and part of the repair endpoint?

bradrydzewski:
I think I am in favor for including this in the repair endpoint directly. When you change the name of a repository it is sort of in a broken state, and the repair function fixes the state. So maybe something like this:

```
drone repo repair foo/bar --renamed=foo/baz
```

We should also fetch the user permissions for the new repository, to ensure they have admin access to the repository with the provided name.

josmo:
I think it can make sense to use the repair endpoint.  On the CLI I tend to like the``` drone repo mv --unsafe currentproject/currentrepo newproject/newrepo ``` since it's a bit more intuitive from a user. Sound reasonable?

bradrydzewski:
Yes I like the idea of adding a new move command and endpoint

```
POST /api/repos/{owner}/{repo}/move?to={owner}/{repo}
```

The endpoint should probably handle updating the webhook, since some people might use the API directly and not the CLI

josmo:
@bradrydzewski Let me know if this seems reasonable.  Looks like it works on my side.

server/repo.go
```diff
@@ -235,3 +220,53 @@ func RepairRepo(c *gin.Context) {
...
+		c.AbortWithError(http.StatusInternalServerError, err)
+	}
+	repo.Name = from.Name
+	repo.Owner = from.Owner
+	repo.FullName = from.FullName
+	repo.Avatar = from.Avatar
+	repo.Link = from.Link
+	repo.Clone = from.Clone
+	repo.IsPrivate = from.IsPrivate
+	repo.Visibility = from.Visibility
```
bradrydzewski: this needs to be calculated using the following logic

```
	if repo.Visibility == "" {
		repo.Visibility = model.VisibilityPublic
		if repo.IsPrivate {
			repo.Visibility = model.VisibilityPrivate
		}
	}
```

bradrydzewski: maybe we add a `repo.ResetVisibility()` function?

server/repo.go
```diff
@@ -235,3 +220,53 @@ func RepairRepo(c *gin.Context) {
...
+	if errStore != nil {
+		c.AbortWithError(http.StatusInternalServerError, errStore)
+		return
+	}
+
+	RepairRepo(c)
+}
+
+// ParseRepo parses the repository owner and name from a string.
+func ParseRepo(str string) (user, repo string, err error) {
```
bradrydzewski: this might be nice to add to `model/repo.go` so that we can re-use in other places?

server/repo.go
```diff
@@ -235,3 +220,53 @@ func RepairRepo(c *gin.Context) {
...
+
+func MoveRepo(c *gin.Context) {
+	remote := remote.FromContext(c)
+	repo := session.Repo(c)
+	user := session.User(c)
+
+	to, exists := c.GetQuery("to")
+	if !exists {
+		err := fmt.Errorf("Missing required to query value")
+		c.AbortWithError(http.StatusInternalServerError, err)
```
bradrydzewski: I think we need to always `return` after an `AbortWithError`.  If you see any examples in the code where we are not, they are probably bugs :)

server/repo.go
josmo: Sounds good I'll add since I'll need to use it when I create a PR to have patch behave the same way

server/repo.go
```diff
@@ -235,3 +220,53 @@ func RepairRepo(c *gin.Context) {
...
+	}
+	repo.Name = from.Name
+	repo.Owner = from.Owner
+	repo.FullName = from.FullName
+	repo.Avatar = from.Avatar
+	repo.Link = from.Link
+	repo.Clone = from.Clone
+	repo.IsPrivate = from.IsPrivate
+	repo.Visibility = from.Visibility
+
```
bradrydzewski: we should fetch the user permissions for the repository and verify they have admin access, and aren't trying to rename their repository to something they don't have access to.

server/repo.go
bradrydzewski: note we should probably only re-calculate visibility if the `Private` value changed

```
if repo.IsPrivate != from.IsPrivate {
  repo.ResetVisibility()
}
```

The visibility value is set by the user, and lets you mark a private repository as publicly visible and vice versa. I think it makes sense to change the visibility automatically when a repository private value is changed.

server/repo.go
josmo: oops, I think it's just me being stupid :), fixed in the next commit

server/repo.go
josmo: done in the next commit

server/repo.go
josmo: done in the next commit

server/repo.go
josmo: I think what I have works. But Let me know if there's a better/preferred way

josmo:
@bradrydzewski I think all the comments you add are taken care of :)

bradrydzewski:
thanks, I think we are good to go :)

PR #2184, opened 2017-08-29
Repo rename change

josmo:
@bradrydzewski I think this fits more in line with what you envisioned from https://github.com/drone/drone-cli/pull/61 Let me know if I took a wrong turn ;)


server/repo.go
```diff
@@ -238,52 +238,86 @@ func RepairRepo(c *gin.Context) {
...
+		destinationRepo.ResetVisibility()
 	}
+	destinationRepo.Counter = currentRepo.Counter
+	destinationRepo.AllowDeploy = currentRepo.AllowDeploy
+	destinationRepo.AllowPull = currentRepo.AllowPull
+	destinationRepo.AllowPush = currentRepo.AllowPush
+	destinationRepo.AllowTag = currentRepo.AllowTag
+	destinationRepo.IsTrusted = currentRepo.IsTrusted
+	destinationRepo.IsGated = currentRepo.IsGated
+	destinationRepo.IsActive = currentRepo.IsActive
```
josmo: @bradrydzewski I noticed that isActive doesn't get set from remote.Activate  I imagine I should disable the currentRepo and save that change?

bradrydzewski:
thanks, let me think about this a bit more and get back to you :)

josmo:
sounds good. Let me know if a different route seems more compelling :)

bradrydzewski:
one additional item we should set:

```
destinationRepo.Hash = currentRepo.Hash
```

josmo:
@bradrydzewski I added Hash. I did wonder if I should disable the currentRepo
```
currentRepo.IsActive = false
```
otherwise it stays in the database as active..  Not sure if that's a big deal or not.

josmo:
@bradrydzewski Just checking what you were thinking about this. I'm about to rename a bunch of repos namespace and would hold off making a custom build of drone if it gets merged in :)

bradrydzewski:
closing since renaming was built-in to the sync process in 1.0

PR #2187, opened 2017-08-30
Adding a contributing.md guide.  Still lots of TBD but wanted a start…

josmo:
…ing place. [ci skip]

I just wanted to get a starting point for a contribution doc.  Seems like the drone-core repo was the best place for it.


CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
...
+
+ * Checkout Golangs [Common code review comments](https://github.com/golang/go/wiki/CodeReviewComments)
+
+### Git Commit Messages
+
+* Use the present tense ("Add feature" not "Added feature")
+* Limit the first line to 72 characters or less
+* Reference issues and pull requests liberally after the first line
+* When only changing documentation, include `[ci skip]` in the commit description
+* Consider starting the commit message with an applicable emoji:
```
metalmatze: I don't see any emojis in the history of commit messages. Additionally I don't see a reason to start with it.

josmo: Fair, just grabbed it from a past one :) I'll remove

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+
+## I don't want to read this whole thing I just have a question!!!
+
+> **Note:** [Please don't file an issue to ask a question.](http://docs.drone.io/getting-help/)
+
+## What should I know before I get started?
+
+### Drone and Plugins
+
+Drone in itself is comprised of [several projects](https://github.com/drone), [Community Plugins](https://github.com/drone-plugins) and [Public plugins](http://plugins.drone.io/)
```
tonglil: https://github.com/drone-plugins is the official plugins, and http://plugins.drone.io/ is the plugin index listing plugins and docs.

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+
+## How Can I Contribute?
+
+### Your First Code Contribution
+
+Feel free to look through the issues list and find something to start working on!
```
tonglil: Please include discussing the implementation before starting to work on something.

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+### Go common code review comments
+
+ * Checkout Golangs [Common code review comments](https://github.com/golang/go/wiki/CodeReviewComments)
+
+### Git Commit Messages
+
+* Use the present tense ("Add feature" not "Added feature")
+* Limit the first line to 72 characters or less
+* Reference issues and pull requests liberally after the first line
+* When only changing documentation, include `[ci skip]` in the commit description
```
tonglil: The docs are not kept in here, and they shouldn't skip CI otherwise they won't get built and published.

CONTRIBUTING.md
tboerger: the drone-plugins org have been given to the community, just the core plugins like docker, git and slack got limited write access.

tonglil: Ah gotcha. How does that work? We should probably document how plugins can be added.

tboerger: It's planned to write up some guidelines, but we have not done so far. Generally the plugins should be already in use and there should be some general need for the plugin, beside that it should be written in Go and have a similar structure like the other plugins :)

ashwilliams1:
I think we are standardizing on the following document, which we have started using in some of our other projects:
https://github.com/drone-runners/drone-runner-exec/blob/master/.github/contributing.md

Issue #2171, opened 2017-08-18
Add Drone Agent build filtering


gtaylor:
Particularly in cloud environments with instance-based IAM systems, it's important to make sure that builds run on the instance with the most appropriate IAM profiles relative to the build. For example:

* If repo1 requires pushing artifacts to an S3 or GS bucket, it'll need to run on an instance that has the correct IAM profile.
* If repo2 does not require elevated privs, it should not run on an instance with IAM privs that it does not need.

While this could be managed with .drone.yaml labels, doing this at the agent level for critical/sensitive builds would lessen the likelihood of external change. For example:

* I set some labels in my .drone.yaml, pointing at a particular subset of my cluster that is only for this build.
* Someone sneaks a label change into one of my hundreds of other repos, thus giving the malicious party an open avenue for messing with whatever privs my IAM profile grants.

I won't be very prescriptive about implementation details, but here are the things that would be great to be able to do:

* Glob org and/or repo names: `myorg/some*`, `my*/someprefix-*`. This can be potentially dangerous in environments without strict control over repo creation, so we may consider including a note warning as such.
* Set multiple patterns to match: `myorg/repo1,myorg/repo2`

ozbillwang:
In our case, I need Drone agents to report itself with a name (label or tag), for example, `agent-dev`, `agent-stag`, agent-prod`

So if I need it to be built for dev environment drone server will send the job to these `drone-dev` agents.

Pipeline file should be something like this

```diff
pipeline:
  backend:
    image: golang
    commands:
      - go build
      - go test
+   agent:
+    - drone-dev
```

sergekov:
I would vote up for this feature. It is extremely important to be able to select agent based on certain criteria in the cloud. Without it I am unable to limit a developer to certain environment such as dev/uat/staging. Therefore bad code can be executed and deployed in production.

gtaylor:
@ozbillwang Only problem with defining this in the .drone.yaml is that I can sneak a change in there to run my build on a privileged agent.

If it's an agent env var, it gets a bit harder to sneak escalated builds in.

bradrydzewski:
The solution we will (re) implement was briefly and successfully tested when we used stomp as the rpc protocol which accepts sql-like query expressions to filter messages. Some of the ground work has already been laid to enable this capability.

Why expressions? Because people have diverse needs including limiting by random fields such as ram, cpu count, gpu, installed software packages and more.

### Agent Changes

[...]

PR #2199, opened 2017-09-08
Add repo to task labels for build filtering

praxist:
Second PR with regards to https://github.com/drone/drone/issues/2171#issuecomment-327589521.

We only have an immediate need for filtering based on repo, so I didn't add the environment to the labels as mentioned in the proposed code.

### DIFF
--- a/server/hook.go
+++ b/server/hook.go
@@ -322,10 +322,11 @@ func PostHook(c *gin.Context) {
 		task := new(queue.Task)
 		task.ID = fmt.Sprint(item.Proc.ID)
 		task.Labels = map[string]string{}
-		task.Labels["platform"] = item.Platform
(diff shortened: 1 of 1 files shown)


bradrydzewski:
one minor suggestion ... maybe we set `repo` and `platform` variables after the user-defined labels are added. This way a user cannot override these fields from their yaml.

```diff
		task.Labels = map[string]string{}
-  		task.Labels["platform"] = item.Platform
  		for k, v := range item.Labels {
  			task.Labels[k] = v
  		}
+ 		task.Labels["platform"] = item.Platform
+		task.Labels["repo"] = b.Repo.FullName
```

bradrydzewski:
thanks!

PR #2200, opened 2017-09-09
Add expression parser evaulator for build filter

praxist:
continuation of https://github.com/drone/drone/issues/2171#issuecomment-327589521, using the expression library from https://github.com/drone/expr

### DIFF
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"time"
 
 	oldcontext "golang.org/x/net/context"
 
@@ -22,7 +23,8 @@ import (
 	"github.com/drone/drone/model"
 	"github.com/drone/drone/remote"
 	"github.com/drone/drone/store"
-	"time"
+
+	"github.com/drone/expr"
 )
 
 // This file is a complete disaster because I'm trying to wedge in some
@@ -89,13 +91,9 @@ func (s *RPC) Next(c context.Context, filter rpc.Filter) (*rpc.Pipeline, error) 
 		}
 	}
 
-	fn := func(task *queue.Task) bool {
-		for k, v := range filter.Labels {
-			if task.Labels[k] != v {
-				return false
-			}
-		}
-		return true
+	fn, err := createFilterFunc(filter)
+	if err != nil {
+		return nil, err
 	}
 	task, err := s.queue.Poll(c, fn)
 	if err != nil {
@@ -467,6 +465,32 @@ func (s *RPC) checkCancelled(pipeline *rpc.Pipeline) (bool, error) {
 		return true, nil
 	}
 	return false, err
+}
+
+func createFilterFunc(filter rpc.Filter) (queue.Filter, error) {
+	var st *expr.Selector
+	var err error
+
+	if filter.Expr != "" {
+		st, err = expr.ParseString(filter.Expr)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return func(task *queue.Task) bool {
+		if st != nil {
+			match, _ := st.Eval(expr.NewRow(task.Labels))
+			return match
+		}
+
+		for k, v := range filter.Labels {
+			if task.Labels[k] != v {
+				return false
+			}
+		}
+		return true
+	}, nil
 }
 
 //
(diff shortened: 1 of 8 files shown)


server/rpc.go
```diff
@@ -89,12 +91,27 @@ func (s *RPC) Next(c context.Context, filter rpc.Filter) (*rpc.Pipeline, error)
 		}
 	}
 
+	var st *expr.Selector
+	var err error
+	if filter.Expr != "" {
```
praxist: Thoughts on moving this logic into the `fn` declaration? That would reduce the need for having the extra variable declarations. Or is catching a potential error from the selector initialization important?

bradrydzewski: yes, for caching, otherwise it would re-parse the sql string every time :)

.drone.yml
```diff
@@ -7,6 +7,7 @@ pipeline:
     image: golang:1.8
     commands:
       - go get -u github.com/drone/drone-ui/dist
+      - go get -u github.com/drone/expr
```
bradrydzewski: we can vendor this repository, and then remove this line. Will add a comment with instructions shortly

bradrydzewski:
ok, quick guide to vendoring ...

download the govendor tool

```
go get -u github.com/kardianos/govendor
```

and make sure you have a local copy of the expr library

```
go get -u github.com/drone/expr
```

and then you can run this command:

```
govendor add github.com/drone/expr
```

this will add the code to the vendor directory, which gets checked into the repository.

bradrydzewski:
> Thoughts on moving this logic into the fn declaration? That would reduce the need for having the extra variable declarations. Or is catching a potential error from the selector initialization important?

I'm guessing you were looking to clean up / simplify the code. That file is definitely getting a bit unruly after a lot of changes and years of evolution ...

If you wanted to clean things up a bit, you could break out the function like this:

```Go
func createFilterFunc(filter rpc.Filter) (queue.Filter, error) {
	var st *expr.Selector
 	var err error

 	if filter.Expr != "" {
 		st, err = expr.ParseString(filter.Expr)
 		if err != nil {
 			return nil, err
 		}
 	}

	return func(task *queue.Task) bool {
 		if st != nil {
 			match, _ := st.Eval(expr.NewRow(task.Labels))
 			return match
 		}
 
  		for k, v := range filter.Labels {
  			if task.Labels[k] != v {
  				return false
  			}
  		}
  		return true
	}, nil
}
```

You could then invoke this above function to get the filter, and pass to the poller:

```Go
fn, err := createFilterFunc(filter)
if err != nil {
	return nil, err
}
task, err := s.queue.Poll(c, fn)
```

I am fine either way, so I'll leave the decision up to you :)

praxist:
good recommendation!

ozbillwang:
Nice job, well done!

@bradrydzewski 

Could you provide some instructions on how to use this feature?

bradrydzewski:
I am planning on writing official docs, but in the mean time, here is a quick overview.

The first step is to define a query used to filter which builds the agent will accept. The query expression is similar to sql. For example:

```sql
platform = 'linux/amd64' AND repo GLOB octocat/*
```

There are two fields available by default:

* `repo` is the repository name
* `platform` is the target platform for the build

The user can define additional fields in the labels section of the .drone.yml file. These fields can then be evaluated by the query expression:

```
pipeline:
  build:
    image: golang
    commands: go test

label:
  - ram=32
  - cpu=16
```

For example:

```sql
platform = 'linux/amd64' AND ram >= 32 AND cpu >= 16
```

The query expression string is passed to the agent using the `DRONE_FILTER` environment variable. The agent will only process builds when the expression evaluates to true.

yoanisgil:
@bradrydzewski is it possible to define labels at each step so individual steps are executed by a given agent? It does not look like is possible right now but I just wanted to confirm.

bradrydzewski:
>  It does not look like is possible right now but I just wanted to confirm

It is not possible, but perhaps this is something that would be possible when we launch the .drone.js file, which will have more advanced scripting and conditional capabilities.

yoanisgil:
Quick question though: this new upcoming functionality in .drone.js, is something that can be tested? Like we're in the process of choosing our CI tool and I personally like Drone and features such as this one would really help in the decision process (as well as the monorepo one ;)).

bradrydzewski:
There should be a release candidate for 0.9 end of month, which will include mono-repo support. The .drone.js will probably be part of 0.10. So will be a while.

PR #2201, opened 2017-09-09
Hyper.sh Integration into Drone Agent

LFlare:
Hello,

I've done some towards integrating https://github.com/hyperhq/hyper-api with the agent binary, to allow for dynamically provisioned virtual machines priced secondly for the optimal build agent configurations.

However, I've ran into some issues with the build process especially during environmental setting, as shown below,

```
{"time":"2017-09-09T16:48:52+08:00","level":"debug","repo":"xxxxxx","build":"12","id":"413","error":"Error response from daemon: Environment is invalid, should be [A-Za-z_][A-Za-z0-9_]*","exit_code":1,"message":"updating pipeline status"}
```

Anyone able to help chip in?

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -12,6 +12,7 @@ pipeline:
       - go get -u golang.org/x/net/context/ctxhttp
       - go get -u github.com/golang/protobuf/proto
       - go get -u github.com/golang/protobuf/protoc-gen-go
+      - go get -u github.com/hyperhq/hyper-api
       - go test -cover $(go list ./... | grep -v /vendor/)
 
   test_postgres:
--- a/cmd/drone-agent/agent.go
+++ b/cmd/drone-agent/agent.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cncd/pipeline/pipeline"
 	"github.com/cncd/pipeline/pipeline/backend"
 	"github.com/cncd/pipeline/pipeline/backend/docker"
+	"github.com/cncd/pipeline/pipeline/backend/hyper"
 	"github.com/cncd/pipeline/pipeline/interrupt"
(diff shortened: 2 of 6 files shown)


.drone.yml
```diff
@@ -12,6 +12,7 @@ pipeline:
       - go get -u golang.org/x/net/context/ctxhttp
       - go get -u github.com/golang/protobuf/proto
       - go get -u github.com/golang/protobuf/protoc-gen-go
+      - go get -u github.com/hyperhq/hyper-api
```
tboerger: This should not be required.

bradrydzewski:
thanks @LFlare 

I think the first step is to send a pull request to cncd/pipeline with an updated driver, as opposed to directly updating the vendor directory.

Note that the cncd/pipeline directory has a command line utility that allow you to test pipeline backends directly from the terminal, along with a bunch of samples that you you should be able to execute directly on hyper.

So lets get a PR for the hyper backend in cncd/pipeline:
https://github.com/cncd/pipeline/tree/master/pipeline/backend

And then you can also update the pipec binary to include the hyper token and optionally use the hyper backend:
https://github.com/cncd/pipeline/tree/master/pipec

Once you compile pipec you should be able to directly test one of the samples:

```
cd samples/sample_7_redis
pipec --hyper-access-key=... --hyper-secret-key=...
```

Once this is working in the pipeline library, we can then update the pipeline vendor code using the `govendor` tool and update the drone server and agent accordingly. THanks!

LFlare:
Fair enough. I'm still going to kindly request for your assistance @bradrydzewski with the pipeline part of the code with hyper, since there's that annoying error but I'll close this PR for now.

PR #2203, opened 2017-09-10
Reduce build time

mavimo:
- [ ] Move DB folder to tmpfs (need drone 0.8.0:latest)
 - [x] Create `test-integration` group (parallelize tests on mysql/postgres)
 - [x] Create `dockerimage` group (parallelize tasks)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -114,7 +114,7 @@ services:
       - POSTGRES_USER=postgres
     tmpfs:
       - /var/lib/postgresql/data
-mysql:
+  mysql:
     image: mysql:5.6.27
     environment:
       - MYSQL_DATABASE=test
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -104,8 +112,12 @@ services:
...
       - POSTGRES_USER=postgres
+    tmpfs:
+      - /var/lib/postgresql/data
   mysql:
     image: mysql:5.6.27
     environment:
       - MYSQL_DATABASE=test
       - MYSQL_ALLOW_EMPTY_PASSWORD=yes
+    tmpfs:
+      - /var/lib/mysql
```
bradrydzewski: at this time, tmpfs is only available to trusted repositories. Because drone is a public repository and we accept untrusted pull requests, we cannot enable trusted mode.

mavimo:
@bradrydzewski `tmpfs` removed

PR #2209, opened 2017-09-12
Provide a basic agent healthcheck

bradrydzewski:


tboerger:
I personally would name it version and healthz. But you are the boss :)

Dockerfile.agent
```diff
@@ -1,7 +1,9 @@
-# docker build --rm -t drone/drone .
+# docker build --rm -f Dockerfile.agent -t drone/agent .
 
 FROM centurylink/ca-certs
 ENV GODEBUG=netdns=go
 ADD release/drone-agent /bin/
 
 ENTRYPOINT ["/bin/drone-agent"]
+
+HEALTHCHECK CMD ["/bin/drone-agent", "ping"]
```
tboerger: If you move healthcheck and entrypoint before add the layers can get cached :)

bradrydzewski:
Is healthz a documented standard or convention?

I am currently using the dockerflow convention defined here:
https://github.com/mozilla-services/Dockerflow

I don't care what format or naming convention we use, as long as it is a documented convention used by multiple projects.

bradrydzewski:
made all recommended changes, and I've added some more advanced healthcheck logic with unit tests. I'm going to merge this so I can let it bake for a while before we (hopefully today) tag rc.5

tboerger:
> Is healthz a documented standard or convention?

/healthz is pretty common in with k8s ecosystem, you can see it here as an example: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#advanced-liveness-probe-example

bradrydzewski:
yep, I found some more information about it :)

I also added `/varz` which exposes some nice runtime information:

```
{
  "polling_count": 0,
  "running_count": 1,
  "running": {
    "2626": {
      "id": "2626",
      "repository": "bradrydzewski\/foobar",
      "build_number": "554",
      "build_started": "2017-09-12T21:28:42.006962574Z",
      "build_timeout": 7920000000000
    }
  }
}
```

PR #2210, opened 2017-09-13
Add server env var to specify .drone.yml location

praxist:
We would like a global default for specifying a different path/name for `.drone.yml`. 

This PR allows setting `DRONE_REPO_CONFIG` for the server to change the default path to `.drone.yml`. One consideration discussed was that setting this default does not retroactively change the default path for repos previously created.

Implementation-wise, I couldn't really think of a better place to stick this global - let me know if you'd prefer this elsewhere.

### DIFF
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -60,6 +60,7 @@ var Config = struct {
 		Host           string
 		Port           string
 		Pass           string
+		RepoConfig     string
 		SessionExpires time.Duration
 		// Open bool
 		// Orgs map[string]struct{}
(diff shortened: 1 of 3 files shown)


server/rpc.go
```diff
@@ -60,6 +60,7 @@ var Config = struct {
 		Host           string
 		Port           string
 		Pass           string
+		RepoConfig     string
```
gtaylor: `go fmt` may fix the alignment here

praxist: should already be formatted!

bradrydzewski:
Thanks

Issue #2217, opened 2017-09-14
Filter events by tag and its base ref


msuterski:
Our use case requires that we would only run certain builds on tag events based on a specific branch. 

Specifically, it's to ensure that the tag event triggers a prod deployment only from master, which is a protected branch with restricted write access. 

We're using github, so my references will be limited to it.

I couldn't find a way to do it in the drone's yml syntax or a reference to `base_ref` in the drone's source. 

I'd be happy to file a PR with this change. I could either add a new option to `events` (`base_branch` ?) or override the `Build.Ref` with a value of `base_ref` in case the event is for a tag.

bradrydzewski:
ok yes, I see what you mean. I see the `base_ref` field in the payload

```js
{
  "ref": "refs/tags/v0.8.0-rc.5",
  "before": "0000000000000000000000000000000000000000",
  "after": "538f25683450489a96808bc11cb5400c1939ac82",
  "created": true,
  "deleted": false,
  "forced": false,
  "base_ref": "refs/heads/master",
  "compare": "https://github.com/drone/drone/compare/v0.8.0-rc.5",
  "commits": [

  ]
  /* truncated */
}
```

I think we can safely set the `build.Branch` to the `base_ref` for tag events. I do not think we need a new field here, and I don't think there will be any adverse effects using the branch field.

You will then be able to do something like this:

```
when:
  event: tag
  branch: master
```

Note that we probably want to implement the following basic logic just to be safe:

* only use the `base_ref` if starts with `refs/heads/`
* when setting `base_ref` to the branch, we should trim prefix `refs/heads/`

I would definitely accept a PR for this as described above, if you think it would solve your use case.

msuterski:
Great. I'll start working it. Thanks.

tboerger:
But branch is in case of a tag already `refs/tags/v0.8.0-rc.5`?

tboerger:
I'm already using something like this:

```
when:
  event: [push, tag]
  branch: [master, refs/tags/*]
```

bradrydzewski:
@tboerger I just added ref as a proper constraint:
https://github.com/cncd/pipeline/commit/b7427ae1fb86bb04e2997e5f3a1eec75c4b056ef

so we can do something like this:

```
when:
  event: [ push, tag ]
  ref: [ refs/heads/master, /refs/tags/* ]
```

I think this makes a bit more sense, since storing the ref in the branch only works in github. This approach should work with all providers.

[...]

PR #2218, opened 2017-09-15
Use 'base_ref' as branch name for github tag events.

msuterski:
Implements https://github.com/drone/drone/issues/2217

This change adds ability to filter builds by the source branch name for tag
events from Github.

### DIFF
--- a/remote/github/convert.go
+++ b/remote/github/convert.go
@@ -204,6 +204,11 @@ func convertPushHook(from *webhook) *model.Build {
 		// just kidding, this is actually a tag event. Why did this come as a push
 		// event we'll never know!
 		build.Event = model.EventTag
+		// For tags, if the base_ref (tag's base branch) is set, we're using it
+		// as build's branch so that we can filter events base on it
+		if strings.HasPrefix(from.BaseRef, "refs/heads/") {
+			build.Branch = strings.Replace(from.BaseRef, "refs/heads/", "", -1)
+		}
 	}
 	return build
 }
--- a/remote/github/convert_test.go
+++ b/remote/github/convert_test.go
@@ -247,5 +247,25 @@ func Test_helper(t *testing.T) {
 			g.Assert(build.Event).Equal(model.EventTag)
 			g.Assert(build.Ref).Equal("refs/tags/v1.0.0")
 		})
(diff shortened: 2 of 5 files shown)


remote/github/convert.go
```diff
@@ -204,6 +204,11 @@ func convertPushHook(from *webhook) *model.Build {
 		// just kidding, this is actually a tag event. Why did this come as a push
 		// event we'll never know!
 		build.Event = model.EventTag
+		// For tags, if the base_ref (tag's base branch) is set, we're using it
+		// as build's branch so that we can filter events base on it
+		if strings.HasPrefix(from.BaseRef, "refs/heads/") {
+			build.Branch = strings.Replace(from.BaseRef, "refs/heads/", "", -1)
```
bradrydzewski: you can use `strings.TrimPrefix` here

remote/github/convert_test.go
```diff
@@ -247,5 +247,25 @@ func Test_helper(t *testing.T) {
...
+		})
+
+		g.It("should not convert tag's base_ref from webhook if not prefixed with 'ref/heads/'", func() {
+			from := &webhook{}
+			from.Ref = "refs/tags/v1.0.0"
+			from.BaseRef = "refs/refs/master"
+
+			build := convertPushHook(from)
+			g.Assert(build.Event).Equal(model.EventTag)
+			g.Assert(build.Branch).Equal("v1.0.0")
```
bradrydzewski: In the event that we have no match the `build.Branch` is set to ref. To pass the unit test you will need the following adjustment:

```diff
-g.Assert(build.Branch).Equal("v1.0.0")
+g.Assert(build.Branch).Equal("refs/tags/v1.0.0")
```

We could trim prefix`refs/tags/` from the branch since we are cleaning things up. Or we could leave the current behavior in place. I'm open to either option.

msuterski:
why is go test -cover $(go list ./... | grep -v /vendor/)`  not running all the tests (including the converter ones) locally? Do I need to have something set up?

remote/github/convert.go
msuterski: Ah, sure. Will do.

PR #2219, opened 2017-09-15
graceful shutdown on SIGTERM

bradrydzewski:


gtaylor:
Beat me to it

tonglil:
Thanks!

Wondering why this was not just implemented in cncd/pipeline?

bradrydzewski:
I am planning on splitting the pipeline codebase into a `drone-runtime` and `drone-yaml` package, because we'll end up supporting multiple yaml formats, including from other vendors, able to compile down to the drone binary format.

PR #2236, opened 2017-09-29
Use yaml anchors to simplify drone config

tonglil:
Should make it easier to edit just the relevant release values in the future.

Happy to rebase on https://github.com/drone/drone/pull/2203 if need be.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,3 +1,19 @@
+---
+docker_config: &docker_config
+  image: plugins/docker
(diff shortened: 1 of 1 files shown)


tboerger:
I'm personally not a fan of it. More up and down scrolling, not directly obvious what happens...

ptagr:
It definitely makes things a lot less redundant. I'd rather use it only for the image though.

tonglil:
Timing this one out.

PR #2241, opened 2017-10-12
Urgent fix for changed auth behavior on Gitea

tboerger:
Within the PR https://github.com/go-gitea/gitea/pull/2184 on Gitea the
project dropped the standard behavior to use x-auth-basic personal
access token. So we should use a proper username and the token from now
on to support Gitea >= 1.2.0. Since this driver anyway only supports
Gitea >= 1.2.0 this won't be a breaking change on Drone side, the only
breaking change is on side of Gitea!

### DIFF
--- a/remote/gitea/gitea.go
+++ b/remote/gitea/gitea.go
@@ -271,8 +271,8 @@ func (c *client) Netrc(u *model.User, r *model.Repo) (*model.Netrc, error) {
 		}, nil
 	}
 	return &model.Netrc{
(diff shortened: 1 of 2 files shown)


metalmatze:
And I was wondering what changed in drone 0.8 that it didn't work anymore. :confused: 
But I was running 1.2 RCs and apparently that was the problem.

ptman:
When is this going to be part of a release?

tboerger:
With the next upcoming tag

Issue #2261, opened 2017-11-08
Feature request - Force HTTPS if LetEncrypts is enabled


khanhicetea:
Could we redirect all http (except ACME `~/.well-known/*`) to https if LetEncrypts option is enabled ?

And adding `HTTP Strict Transport Security`

Thanks.

tboerger:
If Let's Encrypt is enabled it is already redirecting the requests to HTTPS, it's just not setting the HSTS header.

khanhicetea:
Hi @tboerger ,

I re-check my drone server (v0.8.1) via curl, and this is what I got (I replaced hostname and ip)

```bash
$ curl -vvv http://drone.example.com/
*   Trying 172.17.17.2 ...
* Connected to drone.example.com (172.17.17.2) port 80 (#0)
> GET / HTTP/1.1
> Host: drone.example.com
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Access-Control-Allow-Origin: *
< Cache-Control: no-cache, no-store, max-age=0, must-revalidate, value
< Content-Type: text/html; charset=UTF-8
< Expires: Thu, 01 Jan 1970 00:00:00 GMT
< Last-Modified: Wed, 08 Nov 2017 15:47:23 GMT
< X-Content-Type-Options: nosniff
< X-Drone-Version: 0.8.1+build.1318
< X-Frame-Options: DENY
< X-Xss-Protection: 1; mode=block
< Date: Wed, 08 Nov 2017 15:47:23 GMT
< Content-Length: 370
```

Then I try to access via http url on browser => it redirected to Github OAuth page with redirect_url was an http url

Thanks !

khanhicetea:
We already have HSTS header provide by middleware `Secure` 

```go
// Secure is a middleware function that appends security
// and resource access headers.
func Secure(c *gin.Context) {
	c.Header("Access-Control-Allow-Origin", "*")
	c.Header("X-Frame-Options", "DENY")
	c.Header("X-Content-Type-Options", "nosniff")
	c.Header("X-XSS-Protection", "1; mode=block")
	if c.Request.TLS != nil {
		c.Header("Strict-Transport-Security", "max-age=31536000")
	}

	// Also consider adding Content-Security-Policy headers
	// c.Header("Content-Security-Policy", "script-src 'self' https://cdnjs.cloudflare.com")
}
```

tboerger:
[...]

PR #2263, opened 2017-11-14
Redirect HTTP to HTTPS when SSL is enabled.

lodge93:
In our current drone setup, we are not using a proxy, thus letting drone handle
SSL termination. In addition, we are not exposing port 80 (effectively disabling
insecure drone access). When new engineers join and attempt to access drone,
they are not sent a 301 and often complain that they either do not have access
or that drone is broken (when in reality they are just accessing drone via the
incorrect protocol/port).

This commit changes the default behavior when running drone with a server-cert
by only sending redirects on port 80 rather than allowing both secure and
insecure access.

### DIFF
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -565,7 +565,7 @@ func server(c *cli.Context) error {
 	// start the server with lets encrypt enabled
 	// listen on ports 443 and 80
 	g.Go(func() error {
-		return http.ListenAndServe(":http", handler)
+		return http.ListenAndServe(":http", http.HandlerFunc(redirect))
 	})
 
 	g.Go(func() error {
@@ -681,6 +681,9 @@ func redirect(w http.ResponseWriter, req *http.Request) {
 	serverHost = strings.TrimPrefix(serverHost, "https://")
 	req.URL.Scheme = "https"
 	req.URL.Host = serverHost
+
+	w.Header().Set("Strict-Transport-Security", "max-age=31536000")
+
 	http.Redirect(w, req, req.URL.String(), http.StatusMovedPermanently)
 }
 
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -536,7 +536,7 @@ func server(c *cli.Context) error {
 	// start the server with tls enabled
 	if c.String("server-cert") != "" {
 		g.Go(func() error {
-			return http.ListenAndServe(":http", handler)
+			return http.ListenAndServe(":http", http.HandlerFunc(redirect))
 		})
 		g.Go(func() error {
 			serve := &http.Server{
@@ -673,6 +673,15 @@ func (a *authorizer) authorize(ctx context.Context) error {
 		return errors.New("invalid agent token")
 	}
 	return errors.New("missing agent token")
+}
+
+func redirect(w http.ResponseWriter, req *http.Request) {
+	var serverHost string = droneserver.Config.Server.Host
+	serverHost = strings.TrimPrefix(serverHost, "http://")
+	serverHost = strings.TrimPrefix(serverHost, "https://")
+	req.URL.Scheme = "https"
+	req.URL.Host = serverHost
+	http.Redirect(w, req, req.URL.String(), http.StatusMovedPermanently)
 }
 
 func cacheDir() string {


bradrydzewski:
cc @tboerger can you provide a review / thoughts?

cmd/drone-server/server.go
```diff
@@ -675,6 +675,12 @@ func (a *authorizer) authorize(ctx context.Context) error {
 	return errors.New("missing agent token")
 }
 
+func redirect(w http.ResponseWriter, req *http.Request) {
+	req.URL.Scheme = "https"
+	req.URL.Host = req.Host
```
tboerger: Maybe it's better to set this with the parse result of the server-host flag.

lodge93: In general, I would prefer to be redirected to the host the request was made to (mostly for debugging, being able to go to `drone-server001.company.com` directly instead of `drone.company.com` is often useful). Given the oauth flow redirects back the the server-host anyways, I can see why redirecting here would be desired. Will update!

tboerger: But this potentially lead to invalid certificates, especially if let's encrypt is enabled.

khanhicetea:
#2261 

Could we implement this for LetEncrypt http handler too ?

lodge93:
Sorry @khanhicetea, I didn't see issue #2261 when I originally created the PR. Please take a look and verify the additional changes resolve the issue.

cmd/drone-server/server.go
```diff
@@ -675,6 +675,18 @@ func (a *authorizer) authorize(ctx context.Context) error {
...
 }
 
+func redirect(w http.ResponseWriter, req *http.Request) {
+	var serverHost string = droneserver.Config.Server.Host
+	serverHost = strings.TrimPrefix(serverHost, "http://")
+	serverHost = strings.TrimPrefix(serverHost, "https://")
+	req.URL.Scheme = "https"
+	req.URL.Host = serverHost
+
+	w.Header().Set("Strict-Transport-Security", "max-age=31536000")
```
khanhicetea: We already had HSTS header in `Secure` middleware : https://github.com/drone/drone/blob/master/router/middleware/header/header.go#L38

lodge93: The secure middleware is currently being applied to gin routes and therefore the header would not be added to a `net/http` handler: https://github.com/drone/drone/blob/bb8c569249db8fe42cac7fee6d8c50e3b46b68ca/router/router.go#L26

In addition, the current middleware only sends the header to requests that already have TLS enabled:
https://github.com/drone/drone/blob/8eacbe382b73f4f51f2d2042086a469f2e6a8028/router/middleware/header/header.go#L43-L45

bradrydzewski:
looks good, thanks!

cmd/drone-server/server.go
bradrydzewski: I will improve this a bit later when I'm refactoring some of our routing code in 0.9

khanhicetea: I mean `Strict-Transport-Security` header is no need for http response, just need for https response (after redirecting)

lodge93: That's a good point @khanhicetea. I actually didn't know the RFC mandates that the header is ignored on non-secure requests, which means setting the header is actually useless:
> If an HTTP response is received over insecure transport, the UA
      MUST ignore any present STS header field(s).

https://tools.ietf.org/html/rfc6797#section-8.1

tboerger:
Instead of stripping prefixes on your own you could also simply use net/url, that would make it better readable.

lodge93:
@tboerger That's a good point. I originally used `net/url` but I ran into a few issues. If the scheme was missing, the host was placed into the path field of the struct and would not error. If the scheme was available, the host was placed into the host field as expected. For example:
```
package main

import (
	"log"
	"net/url"
)

func main() {
	u, err := url.Parse("https://dev.example.com")
	if err != nil {
		log.Println(err)
		return
	}
	log.Println(u.Scheme)
	log.Println(u.Opaque)
	log.Println(u.User)
	log.Println(u.Host)
	log.Println(u.Path)
	log.Println(u.RawPath)
	log.Println(u.ForceQuery)
	log.Println(u.RawQuery)
	log.Println(u.Fragment)
}
```
Would output differently depending on if the scheme was included. This may not have been a big deal but I could not find any other validation of server-host and wanted to make sure the redirect would still work with scheme-less server-hosts.

The other issue was I was unsure how to handle the error. I originally thought about simply ignoring the error, but given that there is no other validation on this configuration option, I decided that was not a good solution. I was then leaning towards throwing a 500 with a message body informing the user of the misconfigured user host, but I would not be using a the standard error format that the rest of the application uses.

Looks like this could use some work, I am more then willing to submit another PR to do proper validation and fold the route into the standard routes if you think it's worth it.

tboerger:
IMHO this is a misconfiguration that should be handled much earlier so that you don't have to care about it at this place.

tboerger:
The Let's Ecrypt host match lao just relies on a properly predefined value: https://github.com/drone/drone/blob/master/cmd/drone-server/server.go#L572-L584

Issue #1980, opened 2017-03-26
[Feature] Builds auto cancellation


PeterDaveHello:
As discussion in:
http://discourse.drone.io/t/feature-build-auto-cancel/401

It'll be good to have a feature to automatically cancel the builds on a single branch or pull request, this can really help to save resources on CI cluster when a user is keep committing or changing something especially on GitHub web GUI since they can't work on multiple fine at a single time, may trigger 5~10 builds or even more, which can cause a huge resource wasting(especially on the builds having huge building works or on the fat repo like @cdnjs).

luxifer:
I'll have a look at this

ozbillwang:
Got updates from @bradrydzewski , quote here 

> note that we would not cancel a running build when new commits are received, because it could negatively impact things like deployments. Instead we will cancel pending (non-running) builds in favor of the newer commits

My reply:

>Thanks. I still prefer the feature to cancel a running build, because in our pipeline, build in feature branches are not really apply the changes, only unit test, integration test, and only apply the changes in dev environment. So it is fine to cancel the running job in some cases. It will save a lot of time for developers.

casiodk:
How far is the feature from release?

leandromatos:
Any update of this feature?

bradrydzewski:
we do not have any updates to share.  if / when we have updates to share we will comment on the issue accordingly. No comments should indicate the issue is not being worked on, and not part of our short term roadmap.

bradrydzewski:
we have an initial implementation of this feature available in master / latest. You can automatically cancel pending push or pull request events if a newer build is scheduled. You can configure this behavior from the command line using the following:

```
$ drone repo update --auto-cancel-pull-requests=false octocat/hello-world
$ drone repo update --auto-cancel-pushes=false        octocat/hello-world
```

note that we do not cancel running builds. I know this was mentioned above, however, this is not something we plan to implement. If you need this feature, you can implement a custom webhook to handle canceling running builds. https://github.com/drone/boilr-webhook

shqear93:
Just a reminder, don't forget to document it please ^^

shqear93:
+1 to implement this feature

PR #2267, opened 2017-11-19
Add support for automatic cancellation of running/pending builds

oldmantaiter:
By using two new variables (flagged as EXPERIMENTAL), new builds
for a branch will check to see whether it should cancel any preceding
builds.

Cancelling pending builds uses the same logic that killing zombie processes
does, and cancelling running builds uses the same logic that cancelling from
the UI does.

Deals, in part, with the ideas discussed in #1980

### DIFF
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -461,6 +461,19 @@ var flags = []cli.Flag{
 		Name:   "coding-skip-verify",
 		Usage:  "coding skip ssl verification",
 	},
+	//
+	// experimental parameters
+	//
+	cli.BoolFlag{
+		EnvVar: "DRONE_EXPERIMENTAL_AUTO_CANCEL_PENDING",
+		Name:   "experimental-auto-cancel-pending",
+		Usage:  "automatically cancel pending builds for same branch",
+	},
+	cli.BoolFlag{
+		EnvVar: "DRONE_EXPERIMENTAL_AUTO_CANCEL_RUNNING",
+		Name:   "experimental-auto-cancel-running",
+		Usage:  "automatically cancel running builds for same branch",
+	},
 }
 
 func server(c *cli.Context) error {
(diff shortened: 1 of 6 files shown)


oldmantaiter:
@bradrydzewski - Have you had time to take a look at this? I've been running it in my stack since I opened the PR and it doesn't seem to have any issues that I've noticed.

bradrydzewski:
Sorry for the delayed reply. Auto cancel is something I would like to see implemented at the repository-level as opposed to a global setting. It would also only cancel pending builds (not yet possible in Drone) as opposed to running builds. The planned design is based on how other systems handle this feature (see below screenshot for prior art)

screenshot

oldmantaiter:
@bradrydzewski - Thanks for the update. I've seen a couple threads asking for auto-cancel of running builds, so I am wondering if once those 0.9 features land I could take a crack at implementing auto-cancel of running builds? I think if people understand that once enabled it is not responsible for gracefully stopping the running build, then maybe it can be labelled as an "at your own risk" feature that will only be allowed in the community edition and hidden behind a server config enabling the server to look for the repo config variable that allows cancelling running builds for the same branch. What do you think?

casiodk:
how far is this feature from release?

damm:
Crickets this feature would be quite helpful as one of our tests spawns a bunch of docker containers (using dind) and if I push too fast I can take down our drone host easily 

32gigs of ram it's not huge but would be helpful if it would auto cancel the previous build in favor of a new build if i push into a branch.

thunderball1:
Is this going to be merged?

bradrydzewski:
@thunderball1 see comment here https://github.com/drone/drone/pull/2267#issuecomment-350341928  tldr we want per-repository implementation, not global

oldmantaiter:
@bradrydzewski Is there a branch for the new queuing system you mentioned in https://github.com/drone/drone/pull/2267#issuecomment-350341928 ?

PR #2282, opened 2017-12-16
BUILDING: add export CGO_ENABLED=0

dlamotte:
While this will work fine without it, you'll hit an issue when you go to
run a docker container built with this because the resulting binary will
look for some dynamic library on start and fail to find it which yields
a "no such file or directory" error.

This at least hints someone down the right track.  If you're using
sqlite3 locally, then you'll need CGO enabled (AFAICT).

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -9,5 +9,6 @@
 
 3. Install binaries to $GOPATH/bin
 
+    export CGO_ENABLED=0
     go install github.com/drone/drone/cmd/drone-agent
     go install github.com/drone/drone/cmd/drone-server


tboerger:
This have not been set intentionally, if we set `CGO_ENABLED=0` the default database settings won't work anymore as it gets enabled if CGO is available.

dlamotte:
Yeah, I noticed that. But when I built it with `go build...` it didn't work
after I built the container. So I figured I'd done something wrong.

- Dan

tboerger:
As you can see at https://github.com/drone/drone/blob/master/.drone.sh#L25-L28 only the agent gets built without CGO.

dlamotte:
Good call.  Missed that.  Sorry, still quite new to drone (especially when building it).

Thanks!

PR #2283, opened 2017-12-16
server: fully qualify Status URI

dlamotte:
Required by Bitbucket Server to properly post status updates.

FWIW, I don't really think this should be merged in this state, but hoped this would be a good starting point for a conversation.  It appears that all other points in the code where status updates occur, we steal the protocol based on the incoming request.  I'm not sure if we can use that same pattern here or if we need to do something else.  If this is an easy fix, please feel free to close this out.  If not, I'd be happy to work on it if you give me some direction on where I should go.

### DIFF
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -425,7 +425,7 @@ func (s *RPC) Done(c context.Context, id string, state rpc.State) error {
 					s.store.UpdateUser(user)
 				}
 			}
-			uri := fmt.Sprintf("%s/%s/%d", s.host, repo.FullName, build.Number)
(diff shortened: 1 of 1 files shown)


tboerger:
I don't think that this is required. The env variable `DRONE_HOST` got to include the protocol, than you will get proper status updates. The only thing that can be improved is the documentation of this environment variable to make it more clear.

dlamotte:
You are correct.  Looks like we were missing it there.  Was hoping it was
something like that.

Thanks!

- Dan

PR #2286, opened 2017-12-20
Validate DRONE_HOST to ensure it contains a scheme and not a trailing slash

jesselang:
Based on discussions here: https://discourse.drone.io/t/incorrect-linkback-from-bitbucket-cloud/1438/5

And instructions here: http://docs.drone.io/installation/

Code often is more effective than documentation.


bradrydzewski:
great idea, thanks for the pull request

PR #2287, opened 2017-12-20
Log the exact error in case of 'error setting commit status'

rumyantseva:
I'm a happy Drone user but now I see a lot of "error setting commit status for..." error in my logs. Unfortunately, without additional information, it might be hard to investigate the problems in this case. Maybe it would be helpful to log the exact error in a case of "error setting commit status" to be able to find the root of the problems faster?

Thank you.


bradrydzewski:
Thanks taking the time to send a pull request, much appreciated. Also check out [this thread](https://discourse.drone.io/t/error-setting-commit-status-for/1229) which might help save you some time debugging. Cheers!

rumyantseva:
@bradrydzewski thank you so much! actually, this is what I was looking for, yeah :)

PR #2313, opened 2018-01-22
fix missing semicolon in Postgres script's files

Valeran86:


bradrydzewski:
The semi-colons are not necessary because we generate .go code from the .sql files. Check out: https://github.com/drone/drone/blob/master/store/datastore/ddl/postgres/ddl_gen.go

tboerger:
But it's just consistent to add them as you have written them on most migration sources :)

bradrydzewski:
once the CLI is signed we can merge

tboerger:
Seems to be signed

PR #2320, opened 2018-01-29
Fix Gitea remote pull-request webhook parsing

mjwwit:
The Gitea remote references non-existent properties of the pull request webhook payload to determine the base branch (and refspec). Using Gitea v1.3.2 (latest stable) the pull request webhook payload does not include the `pull_request.base_branch` or `pull_request.head_branch` properties. It does however, include `pull_request.base.ref` and `pull_request.head.ref` properties that contain the desired information. 

This pull request changes the parsing of pull request webhooks to use the existent properties.
An assertion has been added to test the parsing of the pull request refspec.

### DIFF
--- a/remote/gitea/fixtures/hooks.go
+++ b/remote/gitea/fixtures/hooks.go
@@ -103,13 +103,11 @@ const HookPullRequest = `{
       "email": "gordon@golang.org",
       "avatar_url": "http://gitea.golang.org///1.gravatar.com/avatar/8c58a0be77ee441bb8f8595b7f1b4e87"
     },
-    "base_branch": "master",
     "base": {
       "label": "master",
       "ref": "master",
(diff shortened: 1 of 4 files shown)


mjwwit:
We just checked these changes against our Gitea 1.3.2 instance and PRs now play nice with branch filters again.

lunny:
LGTM

tboerger:
@bradrydzewski I think from the Gitea side this should be fine for merge :)

mxmehl:
I cannot test it in our production system but from your description it sounds like it would solve the referenced issue

mjwwit:
@bradrydzewski we're kind of waiting for this fix to be merged. PR builds from Gitea don't work properly without it. What do you need before you can merge it?

PR #2321, opened 2018-01-30
Drone cli usage

tonglil:
Closes https://github.com/drone/docs/issues/267.

Tried my best to add missing usage fields, added comments where I didn't know what they referred to.

### DIFF
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -107,7 +107,7 @@ var flags = []cli.Flag{
 	cli.StringSliceFlag{
 		EnvVar: "DRONE_ESCALATE",
 		Name:   "escalate",
-		Usage:  "images run in privileged mode",
+		Usage:  "images to run in privileged mode",
 		Value: &cli.StringSlice{
 			"plugins/docker",
 			"plugins/gcr",
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -43,7 +43,7 @@ var flags = []cli.Flag{
 	cli.StringFlag{
 		EnvVar: "DRONE_SERVER_HOST,DRONE_HOST",
 		Name:   "server-host",
-		Usage:  "server host (<scheme>://<hostname>)",
+		Usage:  "server fully qualified url (<scheme>://<host>)",
 	},
 	cli.StringFlag{
 		EnvVar: "DRONE_SERVER_ADDR",
@@ -101,7 +101,7 @@ var flags = []cli.Flag{
 	cli.DurationFlag{
 		EnvVar: "DRONE_SESSION_EXPIRES",
 		Name:   "session-expires",
-		Usage:  "set the session expiration time default 72h",
+		Usage:  "session expiration time",
 		Value:  time.Hour * 72,
 	},
 	cli.StringSliceFlag{
@@ -170,6 +170,7 @@ var flags = []cli.Flag{
 	cli.Int64Flag{
 		EnvVar: "DRONE_LIMIT_SHM_SIZE",
 		Name:   "limit-shm-size",
+		Usage:  "docker compose /dev/shm allowed in bytes",
 	},
 	cli.Int64Flag{
 		EnvVar: "DRONE_LIMIT_CPU_QUOTA",
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -185,6 +185,7 @@ var flags = []cli.Flag{
 	cli.StringFlag{
 		EnvVar: "DRONE_LIMIT_CPU_SET",
 		Name:   "limit-cpu-set",
+		Usage:  "set the cpus allowed to execute containers",
 	},
 	//
 	// remote parameters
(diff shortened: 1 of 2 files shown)


cmd/drone-server/server.go
```diff
@@ -157,28 +158,32 @@ var flags = []cli.Flag{
...
 	cli.Int64Flag{
-		Name:   "limit-mem",
 		EnvVar: "DRONE_LIMIT_MEM",
+		Name:   "limit-mem",
+		Usage:  "maximum memory allowed in bytes",
 	},
 	cli.Int64Flag{
-		Name:   "limit-shm-size",
 		EnvVar: "DRONE_LIMIT_SHM_SIZE",
+		Name:   "limit-shm-size",
```
tonglil: No usage here, couldn't find any docs about this on https://docs.docker.com/engine/admin/resource_constraints/.

cmd/drone-server/server.go
```diff
@@ -157,28 +158,32 @@ var flags = []cli.Flag{
...
 	cli.Int64Flag{
-		Name:   "limit-cpu-shares",
 		EnvVar: "DRONE_LIMIT_CPU_SHARES",
+		Name:   "limit-cpu-shares",
+		Usage:  "change the cpu shares",
 	},
 	cli.StringFlag{
-		Name:   "limit-cpu-set",
 		EnvVar: "DRONE_LIMIT_CPU_SET",
+		Name:   "limit-cpu-set",
```
tonglil: No usage here, couldn't find any docs about this on https://docs.docker.com/engine/admin/resource_constraints/.

cmd/drone-server/server.go
```diff
@@ -38,12 +38,12 @@ var flags = []cli.Flag{
...
 		EnvVar: "DRONE_DEBUG",
 		Name:   "debug",
-		Usage:  "start the server in debug mode",
+		Usage:  "enable server debug mode",
 	},
 	cli.StringFlag{
 		EnvVar: "DRONE_SERVER_HOST,DRONE_HOST",
 		Name:   "server-host",
-		Usage:  "server host",
+		Usage:  "server host (<scheme>://<hostname>)",
```
tboerger: Maybe call it `server fqdn`? That should be more descriptive since it's not only the host name which led to conusion in the past.

cmd/drone-server/server.go
```diff
@@ -101,12 +101,13 @@ var flags = []cli.Flag{
 	cli.DurationFlag{
 		EnvVar: "DRONE_SESSION_EXPIRES",
 		Name:   "session-expires",
-		Usage:  "Set the session expiration time default 72h",
+		Usage:  "set the session expiration time default 72h",
```
tboerger: I think we can skip the `default 72h` because it's anyway part of the help output:

```
   --session-expires value            Set the session expiration time default 72h (default: 72h0m0s) [$DRONE_SESSION_EXPIRES]
```

cmd/drone-server/server.go
```diff
@@ -101,12 +101,13 @@ var flags = []cli.Flag{
...
 		Value:  time.Hour * 72,
 	},
 	cli.StringSliceFlag{
 		EnvVar: "DRONE_ESCALATE",
 		Name:   "escalate",
+		Usage:  "images run in privileged mode",
```
tboerger: `images and services to run in privileged mode`?

cmd/drone-server/server.go
```diff
@@ -124,7 +125,7 @@ var flags = []cli.Flag{
 	cli.StringFlag{
 		EnvVar: "DRONE_AGENT_SECRET,DRONE_SECRET",
 		Name:   "agent-secret",
-		Usage:  "agent secret passcode",
+		Usage:  "drone server-agent shared password",
```
tboerger: `shared password for agent communication`?

cmd/drone-server/server.go
tboerger: https://docs.docker.com/compose/compose-file/#shm_size

cmd/drone-server/server.go
tboerger: https://docs.docker.com/compose/compose-file/compose-file-v2/#cpu-and-other-resources

cmd/drone-server/server.go
tonglil: services are images, no? ;)

cmd/drone-server/server.go
tonglil: That got me to https://docs.docker.com/engine/reference/run/#cpuset-constraint.

Does this refer to `--cpuset-cpus` or `--cpuset-mems`?

cmd/drone-server/server.go
tonglil: Added

cmd/drone-server/server.go
tboerger: Yeah true, thought about plugins and services. :D

cmd/drone-server/server.go
tboerger: I'm not sure about this one, need to dig into the code

mavimo: Is cpu set for [CPUs](https://github.com/cncd/pipeline/blob/82706703c87f88a632674484e095e494c0e30e34/pipeline/backend/docker/convert.go#L43)

tboerger: Thanks

tonglil: Thanks, added.

tonglil:
Should be ready for review @bradrydzewski.

bradrydzewski:
LGTM, thanks for cleaning this up!

PR #2331, opened 2018-02-02
Build List Pagination

jbrockopp:
This PR pertains to [this thread](https://discourse.drone.io/t/add-pagination-to-builds-list/374) on discourse.

I imagine more work will be required on the UI side of things, but figured I would start with the API.

Let me know if anything else is needed, thanks!

### DIFF
--- a/server/badge.go
+++ b/server/badge.go
@@ -72,7 +72,7 @@ func GetCC(c *gin.Context) {
 		return
 	}
 
-	builds, err := store.GetBuildList(c, repo)
+	builds, err := store.GetBuildList(c, repo, 1)
 	if err != nil || len(builds) == 0 {
 		c.AbortWithStatus(404)
 		return
--- a/server/build.go
+++ b/server/build.go
@@ -25,7 +25,13 @@ import (
 
 func GetBuilds(c *gin.Context) {
 	repo := session.Repo(c)
-	builds, err := store.GetBuildList(c, repo)
+	page, err := strconv.Atoi(c.DefaultQuery("page", "1"))
+	if err != nil {
+		c.AbortWithError(http.StatusBadRequest, err)
+		return
+	}
+
+	builds, err := store.GetBuildList(c, repo, page)
 	if err != nil {
 		c.AbortWithStatus(http.StatusInternalServerError)
 		return
--- a/store/datastore/builds.go
+++ b/store/datastore/builds.go
@@ -45,9 +45,9 @@ func (db *datastore) GetBuildLastBefore(repo *model.Repo, branch string, num int
 	return build, err
 }
 
-func (db *datastore) GetBuildList(repo *model.Repo) ([]*model.Build, error) {
+func (db *datastore) GetBuildList(repo *model.Repo, page int) ([]*model.Build, error) {
 	var builds = []*model.Build{}
-	var err = meddler.QueryAll(db, &builds, rebind(buildListQuery), repo.ID)
+	var err = meddler.QueryAll(db, &builds, rebind(buildListQuery), repo.ID, 50*(page-1))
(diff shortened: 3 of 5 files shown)


mavimo:
@jbrockopp I started a similar task some month ago (but not have time to complete and PR it), let me know if you like some help on https://github.com/drone/drone-ui side.

wass3r:
Unless this is available already, I think it would be good to expose the total number of builds somehow (maybe custom header?). It would be helpful on the UI side to be able to determine what to show for paging. Otherwise, the options are limited to lazy loading more content as you reach the bottom of the list.

bradrydzewski:
>  I think it would be good to expose the total number of builds somehow

We can actually calculate this client-side by looking at the build number. The repository object stores the last build number allowing us to calculate the total number of pages.

robertlabrie:
Piling on, please don't hard code the 50 limit, make it a var I can override. Also if you're returning paginated results, isn't there usually some attribute to tell you to fetch more?

gtaylor:
> please don't hard code the 50 limit

Why? If a user sets a large/unreasonable page size limit that causes issues with table locking or query size, we end up with GH issues and Discourse spam. We also have to ask everyone what their page size is if they are encountering issues, rather than picking a reasonable default.

If we _do_ make it configurable, I'd suggest capping at 200 or 250 like AWS does with some of their APIs to prevent self-harm.

Though, I think 50 is just fine.

robertlabrie:
AWS API adds "MaxResults" and "NextToken" to take some of the suckage out of pagination.

Do you have any idea of what the upper limit is or is 50 utterly arbitrary?

bradrydzewski:
It is common to support `?page` and `? per_page ` where per_page is the number of results that are returned in each page. GitHub and GitLab both allow a maximum per_page value of 100. I would be open to a pull request that implements per_page, as long as it enforces a maximum value.

jmccann:
We are only extending `OFFSET` to the current query which already defaults `LIMIT` to `50`.  This was done based on discussions that have already occurred in discourse that we found.

I think a separate PR/Issue can cover adding the ability to adjust the number of records returned from API.

Please let us know if there is anything wanted/needed in regards to the scope of this PR.  Thanks!

jbrockopp:
@bradrydzewski just curious where exactly we left off with this and if anything else was required for this pr? I don't want this to become stale so if there is more work for me to do to get this merged just let me know 👍

bradrydzewski:
@jakejscott hey, I was just reviewing (merging a bunch of PRs today) and I was wondering if the offset should be 0 by default instead of 1? I was testing with sqlite and noticed that offset 1 skipped the first result in the list. 

```
> select repo_id from repos order by repo_id asc limit 5 offset 0;
3
5
6
8
9
```

```
> select repo_id from repos order by repo_id asc limit 5 offset 1;
5
6
8
9
10
```

jmccann:
We'll take a look at that and get it addressed.  Thanks!

bradrydzewski:
@jakejscott my bad, I just realized you have accounted for this:

```
var err = meddler.QueryAll(db, &builds, rebind(buildListQuery), repo.ID, 50*(page-1))
```

bradrydzewski:
Ok I think we are good to merge. I know there are some different ideas for continued improvement both in terms of usability and flexibility, which I am very open to. Lets plan on addressing these in the 0.9 release. We can create a separate issue to track.

In terms of the user interface (based on some changes I was planning as well) I was thinking we can keep it simple to start, and use a simple "more" button and keep appending to the list. We can open a separate issue in the drone-ui repository to discuss details, for those interested.

tonglil:
a simple "more" button is a good start.

PR #2359, opened 2018-03-01
Update router.go

techmexdev:
Removes unused code


tboerger:
I don't think that this should be dropped.

PR #2361, opened 2018-03-02
Execute database migrations and their logging in a transaction

bonifaido:
Otherwise some inconsistencies may occur, and those can leave the migrations in a non-resumable state.

### DIFF
--- a/store/datastore/ddl/mysql/ddl_gen.go
+++ b/store/datastore/ddl/mysql/ddl_gen.go
@@ -1,11 +1,11 @@
 // Copyright 2018 Drone.IO Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //      http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


tboerger:
Looks like your patch doesn't work, the tests are failing because the tables don't exist.

bonifaido:
Yes, now I have fixed the `tx.Commit()`call and works fine locally, but not on the CI.

bradrydzewski:
a few issues:

* not all database providers support transactions for ddl statements. Have you tested and verified against all 3 supported database providers?
* this code is auto-generated so you would need to modify the code generation utility that we are using. By modifying the ddl_gen.go file directly, your changes will be reverted the next time we generate the code
* there should always be a `defer.Rollback()`. There are some blocks of code that return on error which in turn leaves the transaction open

bonifaido:
Something is wrong with the system, because now the build is green.

bonifaido:
Yes you are right, I forgot implicit commit for DDL exists. I will revise this PR and come back with a better solution (because the problem hits us quite regularly).
Thank you for the other comments as well.

tphoney:
Hi @bonifaido thanks for the work on this, but it has been a while and the PR is now conflicting. I am going to close the PR and if you want to rebase and resubmit, we can re-review.

thanks for your patience.

PR #2363, opened 2018-03-03
Add in-memory Vault test server

usererror:
I have been following discussion around the Vault plugin for awhile and with the code released I took a peek. I noticed that you were running Vault in a Docker container for tests but that isn't necessary as Hashicorp provides an in-memory test server implementation. This PR exposes that with the `TestServer()` function.

The relevant test in `vault_test.go` was updated to use the new in-memory test server and the `vault` package is now `go test`-able. The change was very small but there is no external dependency anymore (outside of the go tool).

I also included tests for the test server itself. `TestTestServerWithConf` is a good example of how to go about testing "non-default" backends. This may not seem helpful initially but I think it could prove useful later on if Drone decides to branch out and use other authentication methods than just `token`.


plugins/secrets/vault/testing.go
```diff
@@ -0,0 +1,85 @@
+package vault
+
+import (
+	"testing"
+
+	"github.com/hashicorp/vault/api"
+	vaulthttp "github.com/hashicorp/vault/http"
+	vaulttest "github.com/hashicorp/vault/vault"
```
usererror: I had to call this vaulttest due to `vault` being a type defined in `vault.go`. I'm not sure how this will effect vendoring deps or the godoc for this package. Any thoughts?

usererror:
It also looks like there may be some dependency issues although I just tried to `govendor add` both packages the failed build is complaining about not having.

bradrydzewski:
@usererror yea this is strange. I am going to switch to vgo since that seems to be the direction the community has chosen. That might take some time. I recently ported the drone-cli and it was a complete mess. I promise I will get back to this, but just bear with me. Note that I really like this change, so thanks for the pull request :)

tboerger:
I think the direction of dep or vgo is the way to go isn't clear yet.

ashwilliams1:
we ended up moving the vault code to github.com/drone/drone-vault

PR #2366, opened 2018-03-06
Initial support of Docker secrets

donny-dont:
This patch just takes a stab at using FilePath within Urfave/cli to load docker secrets and hard codes secret paths.

### DIFF
--- a/cmd/drone-agent/main.go
+++ b/cmd/drone-agent/main.go
@@ -1,11 +1,11 @@
 // Copyright 2018 Drone.IO Inc.
-// 
+//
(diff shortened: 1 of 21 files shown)


krasi-georgiev:
nice! 

I guess some docs should also be updated?

bradrydzewski:
@donny-dont can you sync and then I can merge? sorry about that

ashwilliams1:
1.0 switch to [kelseyhightower/envconfig](https://github.com/kelseyhightower/envconfig) which unfortunately does not support loading variables from file. There is an open issue to support this feature at https://github.com/kelseyhightower/envconfig/issues/130.

Issue #2334, opened 2018-02-08
Validation when adding a registry


bradrydzewski:
The add registry endpoint expects a __hostname__ only and should reject any values with special characters such as slashes. Otherwise they cannot be deleted.

shivamMg:
@bradrydzewski There's a nice hostname validator in [goa](https://github.com/goadesign/goa/blob/master/validation.go#L58) framework. Their validators do not come in a separate package though; so to use them we'll have to import the goa package. If it seems superfluous we can implement it ourselves. They've linked the RFC they used for creating that regex: https://www.ietf.org/rfc/rfc1035.txt (Page 7).

shivamMg:
Used `IsHost` validator from [govalidator](https://github.com/asaskevich/govalidator) instead. The one in goa is just for `<label>` from the RFC, and not for `<domain>`. Regex for domain becomes pretty complex. Also an IP can also be a possible registry address.

bradrydzewski:
Merged into 0.9 release. Pending verification.

PR #2370, opened 2018-03-10
Add validation for Registry Address

shivamMg:
Registry address should be a valid hostname apart from being non-empty.

---

Fixes: https://github.com/drone/drone/issues/2334

Uses host validator from [govalidator](https://github.com/asaskevich/govalidator) to validate hostnames. Added it as a dependency.


xordspar0:
@bradrydzewski, You said in the relevant issue that this is merged in 0.9. Does that mean that this PR isn't needed and can be closed?

PR #2375, opened 2018-03-21
Urgent fix for changed auth behavior on Gitea

tboerger:
Within the PR https://github.com/go-gitea/gitea/pull/2184 on Gitea the
project dropped the standard behavior to use x-auth-basic personal
access token. So we should use a proper username and the token from now
on to support Gitea >= 1.2.0. Since this driver anyway only supports
Gitea >= 1.2.0 this won't be a breaking change on Drone side, the only
breaking change is on side of Gitea!

### DIFF
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -1,11 +1,11 @@
 // Copyright 2018 Drone.IO Inc.
-// 
+//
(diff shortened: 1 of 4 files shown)


tboerger:
Totally wrong messages and I don't know where this comes from...

PR #2377, opened 2018-03-21
Just fixed format with go fmt ./...

tboerger:


shivamMg:
+1 for this. Many people, including myself, work with editors that automatically run go fmt on saved files. Need to do git add in patch mode then if you don't want to include irrelevant formatted stuff.

tboerger:
Yeah just run into that while preparing another PR, so I thought let's fix that entirely ;)

PR #2381, opened 2018-03-23
WIP : YAML fallback URL

gboddin:
I did a bit of cleaning on the code proposed a while ago to support remote YAML fallback.

This is (most likely) still a work in progress since I would like @bradrydzewski ( or any drone expert ) to review this code before merging.

I'll rebase my commits when review is complete.


gboddin:
And side note, just like that I replicated the pull_request uses old code issue : https://discourse.drone.io/t/moved-drone-not-pulling-latest-commit/1811/2

hook.go got updated to remove the `os` import but the last build still sees it

tboerger:
What is the specific use case for this change?

gboddin:
Hi @tboerger !  I updated the PR description with those details

tboerger:
I think this feature could be useful, but instead of hard-coding the parameters it could make more sense to handle the value of the env variable as a template which accepts `{{.event}}`, `{{.repo}}`, `{{.branch}}` and `{{.refspec}}`, than it got the biggest flexibility and the env variable could look like `https://my.handler.example.com/{{.repo}}/{{.branch}}/{{.event}}?refspec={{.refspec}}` or anything else.

bradrydzewski:
I think we need the design for triggers before we can more forward. Implementing triggers could render something like this obsolete.

gboddin:
@tboerger I like the idea, I'll go ahead and try to pass `repo` and `build` in the template .

That would even allow `http://server/{{.repo.org}}/{.build.event}.yml` and make a simple webserver quite powerful :)

@bradrydzewski it's the JS related engine right ?

gboddin:
The PR has been updated.

The template engine can now be used to reference any of `Repo`, `Build` and `User` , for instance : 

`DRONE_REPO_CONFIG_FALLBACK_URL=http://172.20.0.1:8001/?event={{ .Build.Event }}&branch={{ .Repo.Branch }}&name={{ .Repo.FullName }}&user={{ .User.Login }}`

```
172.20.0.3 - - [16/May/2018 13:05:00] "GET /?event=push&branch=master&name=gboddin/drone-test&user=gboddin HTTP/1.1" 200 -
```

xordspar0:
@bradrydzewski Is there an issue or forum thread that follows the discussion of the triggers design? If so, a link here would be helpful to keep this PR up to date.

mavimo:
@xordspar0 this feature should be easily implemented as plugin in drone 0.9, see:

 * [#2450 (comment)](https://github.com/drone/drone/issues/2450#issuecomment-417883114)
 * [#1021 (comment)](https://github.com/drone/drone/issues/1021#issuecomment-417819643)
 * [drone/drone-config-plugin-starter](https://github.com/drone/drone-config-plugin-starter) (starter project for custom config plugins)
 * [drone/drone-jsonnet-config](https://github.com/drone/drone-jsonnet-config) (custom plugin that uses jsonnet as the yaml)

PR #2382, opened 2018-03-23
Allow .drone.sh to build OSS versions when not in Drone's org

gboddin:


bradrydzewski:
this file is only needed when building the enterprise edition. If you are compiling drone, you can just use the standard build commands:

```
```

https://github.com/drone/drone/blob/master/BUILDING

gboddin:
Yes and no, let me explain : 

Forking drone to work on your own version and getting CI working is an issue if the `.drone.yml` and `.drone.sh` are left as it is in the current repo.

The goal here is to keep the same code base to build it whatever fork it comes frome

gboddin:
Good point

PR #2386, opened 2018-03-27
Fix EventStreamSSE Memory Leak

fbcbarbosa:
The `Config.Services.Pubsub.Subscribe` interface at EventStreamSSE is being initialized with the global Gin Context. This causes the publisher object to [hang at line 58](https://github.com/cncd/pubsub/blob/master/pub.go#L58) and the goroutine remains opened indefinetly, which can be a source of memory leak.

Setting it to the locally defined ctx object correctly causes the goroutine to close when the function exits due to the deferred close.

See https://discourse.drone.io/t/memory-leak-on-drone-server/1884/4 for further information.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -78,7 +78,7 @@ func EventStreamSSE(c *gin.Context) {
 
 	go func() {
 		// TODO remove this from global config
-		Config.Services.Pubsub.Subscribe(c, "topic/events", func(m pubsub.Message) {
+		Config.Services.Pubsub.Subscribe(ctx, "topic/events", func(m pubsub.Message) {
 			name := m.Labels["repo"]
 			priv := m.Labels["private"]
 			if repo[name] || priv == "false" {


fbcbarbosa:
Some more info on the issue.

Here's the full Subscribe implementation:

```
func (p *publisher) Subscribe(c context.Context, dest string, receiver Receiver) error {
	p.Lock()
	t, ok := p.topics[dest]
	p.Unlock()
	if !ok {
		return ErrNotFound
	}
	s := &subscriber{
		receiver: receiver,
	}
	t.subscribe(s)
	select {
	case <-c.Done():
	case <-t.done:
	}
	t.unsubscribe(s)
	return nil
}
```

It should end when either "c" or "t" channels are closed. So it follows that the memory leak is happening because the Gin Context and the "topic" object would never (or almost never) close.

I'm not familiar with Gin, so not sure if this is expected. From the documentation, golang net/http context should close when the session ends, not sure if Gin has a different implementation. Just bringing it up because there might be some underlying issues here (i.e. is this a bug with Gin? should the topic close itself? No idea).

Anyway, I did some further debugging by adding some logs around https://github.com/drone/drone/blob/master/server/stream.go#L96:

```
	for {
		select {
		case <-rw.CloseNotify():
			logrus.Debugf("user feed: rw.CloseNotify")
			return
                ...
```

Testing shows that when I close my browser, it soon triggers a "rw.CloseNotify()", exiting the function. This deferred function is then called:

```
	defer func() {
		cancel()
		close(eventc)
		logrus.Debugf("user feed: connection closed")
	}()
```

The local context ctx is canceled, however since Subscribe is waiting for the Gin context to close it remains hanging. _With_ this fix, as expected, a rw.CloseNotify() will cancel ctx which _then_ closes the Publisher. 

So we get the c.Writer.CloseNotify() signal, but never get the c.Done() signal, apparently.

PR #2387, opened 2018-03-28
Allow overriding build message on forked builds

frebib:
This is a feature we've desired for some time now. It would be useful to have the ability to set the build message to something resembling:

`Build forked from owner/parent-image #4`

I've tested this locally and works with the existing cli, specifying the message as a parameter:

```bash
drone build start --fork -p message="new message" owner/child-image 3
```


tboerger:
I like this functionality, but is this really enough to get displayed in the ui?

frebib:
Yes it is actually. Any parameters that are specified in the API request are available in the context. I spent a good half an hour tracing through the various code trying to work out how it could be done.

PR #2394, opened 2018-04-05
Add ability to restart build of any state

delfer:
Build in any state can be restarted.
If it was an error - external environment (dependencies) can changes till now.
If build already running/pending - it is not a problem to schedule new build.

### DIFF
--- a/server/build.go
+++ b/server/build.go
@@ -496,11 +496,8 @@ func PostBuild(c *gin.Context) {
 	}
 
 	switch build.Status {
-	case model.StatusPending,
-		model.StatusRunning,
-		model.StatusDeclined,
-		model.StatusBlocked,
-		model.StatusError:
+	case model.StatusDeclined,
+		model.StatusBlocked:
 		c.String(500, "cannot restart a build with status %s", build.Status)
 		return
 	}


server/build.go
```diff
@@ -495,15 +495,6 @@ func PostBuild(c *gin.Context) {
 		return
 	}
 
-	switch build.Status {
-	case model.StatusPending,
-		model.StatusRunning,
-		model.StatusDeclined,
-		model.StatusBlocked,
```
bradrydzewski: I think we might want to keep `Declined` and `Blocked` since this typically indicates a build has a security-related change that needs manual evaluation. But otherwise, we should be able to remove all others.

delfer:
@bradrydzewski, PR updated: error on StatusRunning and StatusBlocked retrieved.

daniel-ac-martin:
Really keen to get hold of this feature. - Any chance of a release?

tboerger:
This will be part of v0.9

daniel-ac-martin:
@tboerger: Do you know when v0.9 will be coming? We're still very keen to get hold of this feature.

tboerger:
@daniel-ac-martin as usual, when it's done :P but to be serious I can't answer that.

korigod:
It's great this commit made its way to v0.8.6, thanks!

tboerger:
True, totally forgot this commit

PR #2395, opened 2018-04-10
receive release hook from gogs

roryamos:
i have a problem when gogs release a version that trigger release event.
i hope drone can process this event so i add some code will process release event as like tag event.


rushairer:
Is there anyone merge this commit?

bradrydzewski:
code moved to github.com/drone/go-scm

PR #2406, opened 2018-04-24
Enable Vault auth through kubernetes auth method

praxist:
Added a feature to obtain the initial Vault token from the Kubernetes auth method.

This works by making a request to the vault server at the specified auth method mount point's login path and presenting the JWT located in a file on a running pod, along with the Kubernetes role to authenticate as.

Vault will then respond with a token and its TTL, if the request is valid.

My Go is pretty rusty, so I expect this code will require a bit of massaging before getting integrated.

### DIFF
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -175,6 +175,24 @@ var flags = []cli.Flag{
 		Usage:  "token to secure prometheus metrics endpoint",
 		Value:  "",
 	},
+	cli.StringFlag{
+		EnvVar: "DRONE_VAULT_AUTH_TYPE",
+		Name:   "drone-vault-auth-type",
+		Usage:  "auth backend type used for connecting to vault",
+		Value:  "token",
+	},
+	cli.StringFlag{
+		EnvVar: "DRONE_VAULT_AUTH_MOUNT_POINT",
+		Name:   "drone-vault-auth-mount-point",
+		Usage:  "mount point for desired vault auth backend",
+		Value:  "",
+	},
+	cli.StringFlag{
+		EnvVar: "DRONE_VAULT_KUBERNETES_ROLE",
+		Name:   "drone-vault-kubernetes-role",
+		Usage:  "role to authenticate as for vault kubernetes auth",
+		Value:  "",
+	},
 	//
 	// resource limit parameters
 	//
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -179,7 +179,7 @@ var flags = []cli.Flag{
 		EnvVar: "DRONE_VAULT_AUTH_TYPE",
 		Name:   "drone-vault-auth-type",
 		Usage:  "auth backend type used for connecting to vault",
-		Value:  "token",
+		Value:  "",
 	},
 	cli.StringFlag{
 		EnvVar: "DRONE_VAULT_AUTH_MOUNT_POINT",
--- a/plugins/secrets/vault/kubernetes.go
+++ b/plugins/secrets/vault/kubernetes.go
@@ -16,22 +16,22 @@ Vault JSON Response
   }
 }
 */
-type VaultAuth struct {
+type vaultAuth struct {
 	Token string `json:"client_token"`
 	Lease string `json:"lease_duration"`
 }
-type VaultResp struct {
-	Auth VaultAuth
+type vaultResp struct {
+	Auth vaultAuth
 }
 
-func getKubernetesToken(addr, role, mountPoint, tokenFile string) (string, time.Duration, error) {
+func getKubernetesToken(addr, role, mount, tokenFile string) (string, time.Duration, error) {
 	b, err := ioutil.ReadFile(tokenFile)
 	if err != nil {
 		return "", 0, err
 	}
 
-	var resp VaultResp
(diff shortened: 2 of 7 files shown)


cmd/drone-server/server.go
```diff
@@ -175,6 +175,24 @@ var flags = []cli.Flag{
 		Usage:  "token to secure prometheus metrics endpoint",
 		Value:  "",
 	},
+	cli.StringFlag{
+		EnvVar: "DRONE_VAULT_AUTH_TYPE",
+		Name:   "drone-vault-auth-type",
+		Usage:  "auth backend type used for connecting to vault",
+		Value:  "token",
```
praxist: envisioning a case where we'd switch on this auth type, "token" as the current default, "kubernetes" for this particular auth method, and any others from https://www.vaultproject.io/docs/auth/index.html in the future

plugins/secrets/vault/opts.go
```diff
@@ -24,3 +28,21 @@ func WithRenewal(d time.Duration) Opts {
 		v.renew = d
 	}
 }
+
+func WithKubernetesAuth() Opts {
+	return func(v *vault) {
+		addr := os.Getenv("VAULT_ADDR")
+		role := os.Getenv("DRONE_VAULT_KUBERNETES_ROLE")
+		mount := os.Getenv("DRONE_VAULT_AUTH_MOUNT_POINT")
+		jwtFile := "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
praxist: unsure the best place to put this. From my knowledge the location of this file will always be the same for Kubernetes pods.

plugins/secrets/vault/opts.go
```diff
@@ -24,3 +28,21 @@ func WithRenewal(d time.Duration) Opts {
 		v.renew = d
 	}
 }
+
+func WithKubernetesAuth() Opts {
```
bradrydzewski: we should make this `WithKubernetesAuth(addr, role, mount string)` and then in the main.go (enterprise version) I will wire everything up and ensure the variables are passed to the function

plugins/secrets/vault/opts.go
```diff
@@ -24,3 +28,21 @@ func WithRenewal(d time.Duration) Opts {
...
+		token, ttl, err := getKubernetesToken(addr, role, mount, jwtFile)
+		if err != nil {
+			logrus.Debugf("vault: failed to obtain token via kubernetes-auth backend: %s", err)
+			return
+		}
+
+		v.client.SetToken(token)
+		v.ttl = ttl
+		v.renew = ttl / 2
+	}
```
bradrydzewski: typically this logic would not happen inside the option, the option would set variables in the `vault` struct and then we would do something like this:

```
// New returns a new store with secrets loaded from vault.
func New(store model.ConfigStore, opts ...Opts) (secrets.Plugin, error) {
	client, err := api.NewClient(nil)
	if err != nil {
		return nil, err
	}
	v := &vault{
		store:  store,
		client: client,
	}
	for _, opt := range opts {
		opt(v)
	}
+	if v.auth == 'kubernetes' {
+		err = v.initKubernetes()
+		if err != nil {
+			return nil, err
+		}
+	} 
	v.start() // start the refresh process.
	return v, nil
}
```

note that we could store the kube configuration in vault like this:

```
type vault struct {
	store  model.ConfigStore
	client *api.Client
	ttl    time.Duration
	renew  time.Duration
+	auth string
+	kubeAuth
+	githubAuth // for example
+
	done   chan struct{}
}

+type kubeAuth struct {
+  addr, role, mount string
+}
```

plugins/secrets/vault/kubernetes.go
```diff
@@ -0,0 +1,51 @@
...
+/*
+Vault JSON Response
+{
+  "auth": {
+    "client_token" = "token",
+    "lease_duration" = "1234s"
+  }
+}
+*/
+type VaultAuth struct {
```
bradrydzewski: lets use `vaultAuth` and `vaultResp` since these are internal data structures

plugins/secrets/vault/vault.go
```diff
@@ -41,11 +41,17 @@ type vaultConfig struct {
...
-	store  model.ConfigStore
-	client *api.Client
-	ttl    time.Duration
-	renew  time.Duration
-	done   chan struct{}
+	store    model.ConfigStore
+	client   *api.Client
+	ttl      time.Duration
+	renew    time.Duration
+	auth     string
```
bradrydzewski: I think we might need to set `v.auth="kubernetes"` in the option

plugins/secrets/vault/opts.go
```diff
@@ -29,20 +25,13 @@ func WithRenewal(d time.Duration) Opts {
...
-			logrus.Debugf("vault: failed to obtain token via kubernetes-auth backend: %s", err)
-			return
-		}
-
-		v.client.SetToken(token)
-		v.ttl = ttl
-		v.renew = ttl / 2
+		v.kubeAuth.addr = addr
+		v.kubeAuth.role = role
+		v.kubeAuth.mount = mount
```
bradrydzewski:
```
if addr != "" {
  v.auth = "kubernetes"
}
```

PR #2411, opened 2018-05-02
Fix vault-k8s expected response type

praxist:
Vault server response for lease_duration is an integer, rather than a string

### DIFF
--- a/plugins/secrets/vault/kubernetes.go
+++ b/plugins/secrets/vault/kubernetes.go
@@ -12,13 +12,13 @@ Vault JSON Response
 {
   "auth": {
     "client_token" = "token",
-    "lease_duration" = "1234s"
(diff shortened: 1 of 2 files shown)


plugins/secrets/vault/kubernetes.go
```diff
@@ -42,10 +42,7 @@ func getKubernetesToken(addr, role, mount, tokenFile string) (string, time.Durat
 		return "", 0, err
 	}
 
-	ttl, err := time.ParseDuration(resp.Auth.Lease)
-	if err != nil {
-		return "", 0, err
-	}
+	ttl := time.Duration(resp.Auth.Lease) * time.Second
```
tboerger: Have you also tried to set the type directly to time.Duration? Than you could drop this line.

PR #2432, opened 2018-06-07
Fix agent panics by checking array size before indexing.

trevorriles:
Ran into an issue where agents would panic when due to an `index out of range` runtime error. This would cause the container to die and jobs would hang. 

I've update the agent code to check that both `Stages` and `Steps` are not empty lists, and if they are empty we should fail and return early from the job executor.

```
{"time":"2018-06-06T14:41:33Z","level":"debug","message":"request next execution"}
panic: runtime error: index out of range
goroutine 12 [running]:
main.(*runner).run(0xc42031a240, 0xbc0da0, 0xc420205780, 0x0, 0x0)
        /go/src/github.com/drone/drone/cmd/drone-agent/agent.go:182 +0x170d
main.loop.func2(0xc4202af030, 0xc4202aef98, 0xbc3300, 0xc42027cf40, 0xc4201cf980, 0x0, 0x0, 0xc4202aef80, 0xc4202aeff0)
        /go/src/github.com/drone/drone/cmd/drone-agent/agent.go:132 +0x193
created by main.loop
        /go/src/github.com/drone/drone/cmd/drone-agent/agent.go:137 +0x910
{"time":"2018-06-06T14:41:34Z","level":"debug","message":"request next execution"}
```

### DIFF
--- a/cmd/drone-agent/agent.go
+++ b/cmd/drone-agent/agent.go
@@ -17,6 +17,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


tboerger:
@bradrydzewski still relevant for 0.9?

PR #2454, opened 2018-07-21
vendor: update envsubst

int-tt:
Update envsubst for bug fix
ex) https://github.com/drone/envsubst/pull/7


bradrydzewski:
we need to bump envsubst, but we switched to go modules

Issue #1665, opened 2016-06-05
Long terminal output is trimmed


makkrnic:
When terminal output is very long, it gets trimmed after some time. See http://sartura-drone.makkrnic.com/makkrnic/openwrt-qemu-x86/6
The terminal output is missing after the first `[ 69%]` and since the build succeeded, there obviously should be more output.

bradrydzewski:
we have open issues https://github.com/drone/drone/issues/1429 https://github.com/drone/drone/issues/733 and https://github.com/drone/drone/issues/1563 concerning large log output

Issue #1563, opened 2016-04-04
Truncate large build logs to include the tail instead of the head


goober:
Thank you for taking the time to use Drone and file an issue or feature request. Before filing an issue please ensure the following boxes are checked, if applicable:

- [x] I have searched for existing issues
- [ ] I have discussed the issue with the community at https://gitter.im/drone/drone
- [ ] I have provided a sample `.drone.yml` file to help the team reproduce
- [ ] I have provided details from the build logs
- [ ] I have provided details from the server logs by running `docker logs drone`
- [x] I am not using the issue tracker to ask why my build failed

The last part of the build output is mostly more interesting for a user than the beginning. e.g. the final build status and build errors are usually printed at the end. When the build output is larger than the 2mb limit it will truncate the output and only include the first 2mb from the logs. I suggest that the truncation is done so the end of the output is included instead.

bradrydzewski:
Let's track #733 which is all about enabling / improving large build logs

PR #2456, opened 2018-07-23
Log truncation

tcely:
Related to #1563 and #1665

### DIFF
--- a/cmd/drone-agent/agent.go
+++ b/cmd/drone-agent/agent.go
@@ -145,7 +145,7 @@ func loop(c *cli.Context) error {
 // The maximum grpc payload size is 4194304. So until we implement streaming
 // for uploads, we need to set these limits below the maximum.
 const (
-	maxLogsUpload = 2000000 // this is per step
+	maxLogsUpload = 4000000 // this is per step
 	maxFileUpload = 1000000
 )
 
--- a/cmd/drone-agent/agent.go
+++ b/cmd/drone-agent/agent.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"io"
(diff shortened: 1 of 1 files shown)


tcely:
With these two changes I'm able to get about 25k lines of logs. (I don't have a build with more output currently.)

The log streaming is nice and fast and shows everything I'm used to. Once you go over the limit though the loop that's dropping lines from the beginning is fairly slow and delays log uploads by a fair amount. I'd like suggestions to make this faster / more efficient if anyone has them.

The next change that will help with supporting very large build logs is compressing the encoded json with something like deflate and sending the proper headers to have the browser inflate it for us.

xordspar0:
Both of the issues you referenced have been closed for quite a while. Is there any other discussion (issues, forum threads) that would provide some background on this?

tcely:
> Both of the issues you referenced have been closed for quite a while. Is there any other discussion (issues, forum threads) that would provide some background on this?

@xordspar0 The issue management around this problem has been quite poor.

The agents limited how much of the build output is sent to the server and drops all output once that limit is reached. This is a huge problem for anyone who has a failing build, as you can't see the cause of the failure and must rebuild manually.

The guessing around DB limits and server / browser limits was not a solution.

tboerger:
Such a huge build output seems to be an exception, otherwise more people would complain about it.

There had been browser crashs without a build log limitation, that's not just a blind guess, this has been tested with different browsers.

As you can see on other issues and pull requests there is generally a slow response time across the last months because brad is focused on the 0.9 release, this is a release with a huge delay already so it's a must to strictly focus on it.

Besides that, 0.9 will drop grpc and replace it with plain http calls, if it will be chunked or something like that to get around some limitations we will see soon.

tboerger:
Ah, and since this pr is only focused on agent to server communication, there are also somewhere documented limitations of the payload size for grpc.

Raffo:
@tboerger I haven't commented before, but the limitation is a major problem for us as well. I'm currently just following the PR and related topics, but I'd recommend not to assume that it's not a real problem because people are not complaining. Obviously just my two cents 😄

tcely:
@Raffo you are welcome to share my solution if you would like.

https://hub.docker.com/r/tcely/drone-agent/

tcely:
@tboerger is this the limit you're looking for?

https://github.com/drone/drone/blob/c48150767c2700d35dcc29b110a81c8b5969175e/cmd/drone-agent/agent.go#L145

bradrydzewski:
increased to 5mb in separate patch

PR #2474, opened 2018-08-30
set default accept to application/json

Jackong:
some oauth2 server only response json data, but response content-type is text/plain.


tboerger:
Maybe some other OAuth2 server, but the servers used by Drone are handling everything fine :)

PR #2479, opened 2018-09-05
Clean up status codes.

Alireza-Ta:
Hey
Just change all http status codes to numerical format in ```server``` directory.

### DIFF
--- a/server/badge.go
+++ b/server/badge.go
@@ -17,12 +17,13 @@ package server
 import (
 	"fmt"
 
-	log "github.com/Sirupsen/logrus"
-	"github.com/gin-gonic/gin"
+	"net/http"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/drone/drone/model"
 	"github.com/drone/drone/shared/httputil"
 	"github.com/drone/drone/store"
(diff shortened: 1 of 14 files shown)


xxxtonixxx:
why??

metalmatze:
I think this is quite the opposite of what you want to do in Go. The constants are way more descriptive and reference the same numbers you've just replaced them with.

Alireza-Ta:
Well, i'm new to go!
I thought if they be in one format would be fine, however you can ignore it if it's a bad practice!

metalmatze:
@Alireza-Ta Thanks for taking the time to contribute nevertheless! I hope you learned something, keep the PRs coming. 👍

metalmatze:
Oh wait. You actually want to use the constants. I'm not really sure how important that is to @bradrydzewski. 🤔

tboerger:
No, I had also been confused. Look at his first commit: https://github.com/drone/drone/pull/2479/commits/8183907c070301440ef8587c92cb7ba4ecb3dc8b

Alireza-Ta:
@metalmatze You said using consts is better than just plain numbers because of that I changed it!

PR #2482, opened 2018-09-07
Updated Alpine base version to 3.8.

jpds:


tboerger:
Is there any specific reason to switch? The current version should be still maintained.

jpds:
Yeah, keeping up to date with software versions is always a good idea! :)

mbger:
Well, [not necessarily](https://twitter.com/lemiorhan/status/935578694541770752).

Daviey:
3.8 has been out since June and shares the current latest tag.  Would seem like a good idea to track the latest upstream release, rather than pin to an ageing version forever?

@mbger The OSX security issue seems pretty irrelevant for this?

tboerger:
I would not pin to an aging version forever, but the currently used version is totally fine supported so far. So imho it's not really required to update it yet.

mbger:
@Daviey My comment not only seems irrelevant, it **is** irrelevant for this issue. Just wanted to make the point that a general statement like _keeping up with software versions is **always** a good idea_ imho is not true.

PR #2485, opened 2018-09-11
Missing equal sign in Dockerfile

szepeviktor:


bradrydzewski:
the equal sign is optional, see https://docs.docker.com/engine/reference/builder/#env

szepeviktor:
@bradrydzewski Thank you.
All other `ENV`-s are with equal sign.

PR #2488, opened 2018-09-12
.drone.yml: Bump to golang:1.11.

jpds:


xordspar0:
The build instructions at https://github.com/drone/drone/blob/master/BUILDING say that the minimum Go version is 1.9, so at the very least the tests should run for Go 1.9 or later. :slightly_smiling_face:

tboerger:
The builds are working fine with 1.8, so maybe we should state 1.8 within the build instructions.

That's also a reason why I would keep the used go version as ensures the promise is still true.

damm:
Should merge it; it would be nice to see some activity in this repo... other than closing support related issues

tboerger:
Looks like you ignored my comment. Besides that there is a pretty heavy ongoing work on the upcoming 0.9 release.

bradrydzewski:
fixed in separate patch.

PR #2492, opened 2018-09-16
Add a CLI flag to set GRPC address

hashworks:
Otherwise it's not possible to run multiple drone instances on one server (without docker).

### DIFF
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -76,6 +76,12 @@ var flags = []cli.Flag{
 		Name:   "server-key",
 		Usage:  "server ssl key path",
 	},
+	cli.StringFlag{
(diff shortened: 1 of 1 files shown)


vtolstov:
as i know new drone does not have grpc

tboerger:
I don't think this is required anymore, upcoming version will drop grpc replaced by an rest api. Besides that, officially done is only supported via docker installation, and within the docker container nobody needs to change the grpc port.

hashworks:
> upcoming version will drop grpc replaced by an rest api
What is the timetable here?

> officially done is only supported via docker installation, and within the docker container nobody needs to change the grpc port
Hm, trought so. However I then wondered why the HTTP port is changeable.

bradrydzewski:
closing since 1.0 no longer uses grpc

PR #2499, opened 2018-09-21
Mention the BUILDING file from README

strk:
Otherwise the only existing links take you to online docs
which are for different branches/versions...


strk:
proper (?) linking pushed with 2aaf50f

PR #2358, opened 2018-02-28
Add ability to configure and use tls with grpc traffic

jmccann:
This PR is to add ability to encrypt GRPC traffic.  I also split the grpc setup code to a `grpc.go` file for each the server and agent.  If you prefer not to have that done I can work to undo it.

At a minimum the following is required to be configured:

### Server
* `DRONE_GRPC_CERT_FILE`
* `DRONE_GRPC_KEY_FILE`

### Agent

* `DRONE_GRPC_TLS_ENABLE`

This code can also be used to configure another layer of authentication using server/client cert/key verification.  The following additional options are also provided to help facilitate that.

### Server

* `DRONE_GRPC_CA_FILE`
* `DRONE_GRPC_VERIFY_MODE`

### Agent

* `DRONE_GRPC_CERT_FILE`
* `DRONE_GRPC_KEY_FILE`
* `DRONE_GRPC_CA_FILE`

Finally, on the agent you can also specify `DRONE_GRPC_SERVERNAME` to override the hostname to use to match the certificate on the server.  This is primarily for testing purposes.

I have done some manual testing of this with a private drone instance built from source.

PR #2501, opened 2018-09-28
Enable TLS for gRPC

GuillaumeSmaha:
Hi,
I made this change to enable gRPC over TLS.

First, I had a parameter `DRONE_GRPC_TLS` in server and agent to enable it manually to avoid breaking change because drone agent needs one of this 2 parameters to be set: `DRONE_SERVER_HTTPS` or `DRONE_SERVER_CERT`.

- `DRONE_SERVER_CERT` is like the parameter on server-side, it permit to pass manually the public key of the certificate. This solution needs that the agent will be restart when the certificate will be renewed.
- `DRONE_SERVER_HTTPS` contains the server host:port of the drone HTTPS server. When the agent try to connect on the gRPC, a request is made to the  HTTPS server. After the request succeeded, the public certificate is extracted from the request and this certificate is returned in the TLS configuration for the gRPC connection.

### DIFF
--- a/cmd/drone-agent/agent.go
+++ b/cmd/drone-agent/agent.go
@@ -17,15 +17,12 @@ package main
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"encoding/json"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


GuillaumeSmaha:
I updated it as WIP. It seems the function `getConfigForClient` is never called

GuillaumeSmaha:
Finally, after multiple tests. gRPC client automatically get the public certificate of the server. So `DRONE_SERVER_HTTPS` is useless.

jmccann:
You may want to read https://github.com/drone/drone/pull/2358 to see the state of grpc

GuillaumeSmaha:
@jmccann Thanks, I check it :)

GuillaumeSmaha:
So, I close it :D

PR #2504, opened 2018-10-01
Use canonical logrus import

SamWhited:
Right now if you try to import Drone as a dependency in a system using [Go modules] you will see an error like this:

```
go: github.com/Sirupsen/logrus@v1.1.0: parsing go.mod: unexpected module path "github.com/sirupsen/logrus"                                                                                   
go: error loading module requirements
```

The canonical name is lowercase (see their `Go.mod` file):

https://github.com/sirupsen/logrus/blob/1ed61965b9e594bf37539680d7f63eccd060314f/go.mod#L1

The version in the vendor dir probably also needs to be updated to fix this, but I was less comfortable making potentially breaking changes to a repo that I'm not familiar with and didn't want to download and learn the vendor tool. For now this makes it possible (I think) for users who have vendoring disabled to use Drone as a dependency.

[Go modules]: https://golang.org/doc/go1.11#modules

### DIFF
--- a/cmd/drone-server/server.go
+++ b/cmd/drone-server/server.go
@@ -43,8 +43,8 @@ import (
 	droneserver "github.com/drone/drone/server"
 	"github.com/drone/drone/store"
 
-	"github.com/Sirupsen/logrus"
 	"github.com/gin-gonic/contrib/ginrus"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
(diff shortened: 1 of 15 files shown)


SamWhited:
Sorry, I din't notice that this repo had a CLA, I can't sign CLAs. There's no way this patch is copyrightable though (or if it is I release all rights to it, etc.) feel free to use it or copy/paste and commit yourself.

PR #2510, opened 2018-10-18
Upgrade envsubst to latest version

tboerger:
Update the github.com/drone/envsubst to the latest version to include
fixes to the substr function.


bradrydzewski:
looks like dep didn't correctly prune a folder that was re-named in a previous version

tboerger:
Not related to dep, related to govendor which is still used on master branch of drone :)

PR #2517, opened 2018-11-05
Fix gRPC DeadlineExceeded  error

hansedong:
Mainly solves a large number of error logs that may appear in the drone-agent

`grpc error: wait(): code: DeadlineExceeded: rpc error: code = DeadlineExceeded desc = context deadline exceeded`

and the drone-server:

`grpc: Server.processUnaryRPC failed to write status stream error: code = DeadlineExceeded desc = “context deadline exceeded”`

Please see https://discourse.drone.io/t/fix-grpc-deadlineexceeded-error/2884 for futher information.

### DIFF
--- a/cmd/drone-agent/agent.go
+++ b/cmd/drone-agent/agent.go
@@ -211,6 +211,7 @@ func (r *runner) run(ctx context.Context) error {
 		logger.Debug().
 			Msg("listen for cancel signal")
 
+		ctx, cancel := context.WithTimeout(ctxmeta, timeout+1*time.Minute)
 		if werr := r.client.Wait(ctx, work.ID); werr != nil {
 			cancelled.SetTo(true)
 			logger.Warn().


bradrydzewski:
I think this would be fixed by settings server keepalive:

```
DRONE_KEEPALIVE_MIN_TIME=20s
```

and agent keepalive values:

```
DRONE_KEEPALIVE_TIME=45s
DRONE_KEEPALIVE_TIMEOUT=30s
```

The deadline exceeded errors typically are a result of a lost network connection, which is typically the result of some sort of network policy (load balancer, etc) that is closing idle connections. This can be prevented with a keepalive.

hansedong:
@bradrydzewski Sorry, this setting doesn't work because the network connection is always smooth. The problem I want to fix with this pull request is not caused by the network failure.

hansedong:
I have reproduced and fixed this problem in our production environment, and now it works very well, so I would like to ask, can this pull request be merged into the official repository?

bradrydzewski:
closing since 1.0 no longer uses grpc

PR #2528, opened 2018-11-15
Update install documentation link in README.md

1davidmichael:
The current installation link here is 404ing: https://github.com/drone/drone/


bradrydzewski:
this was actually solved at the documentation-level, with support for some of the old urls to avoid broken links. Thanks though :)

PR #2545, opened 2018-12-08
implement recent builds

### LINKED ISSUE #2543: Add Endpoint for Builds
I would like to add an endpoint for an administrator to paginate through all builds at `/api/builds` including recent builds at `/api/builds/recent`. This could be useful for system administrators that want to quickly monitor recent builds for anomalies.

zetaab:
hopefully fixes #2543 

No idea is this what was asked. Also no idea should this have pagination or is it good like this? Now this will print 50 latest builds (it will ignore status of build) from drone server.


harrya:
What's the status of this PR?

tphoney:
With the new Drone 2.0 UI this should be implemented by default. Many apologies for getting to this so late.

PR #2563, opened 2019-01-01
Extend validation on secret names

72636c:
From Discourse:

> [This is because you secret names include a slash character, which should not be allowed. I need to add some validation on the server-side to the name field. The slash interferes with the REST API calls.](https://discourse.drone.io/t/drone-cloud-secret-deletion-fails-with-json-parse-error/3355/2)

I've only disallowed the forward slash for now, since I know that's broken in my secret names.

### DIFF
--- a/model/secret.go
+++ b/model/secret.go
@@ -17,6 +17,7 @@ package model
 import (
 	"errors"
 	"path/filepath"
+	"strings"
 )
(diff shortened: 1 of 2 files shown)


bradrydzewski:
fixed in separate patch.

PR #2570, opened 2019-01-10
Fixing of ineffassign issues

saromanov:
Hi! I fixed issues related to report from `ineffassign` linter. This is issues where errors from methods is not properly checked. You can review it


tboerger:
Thank you for your effort, but we won't release any 0.8.x version anymore, so we won't merge this pr as this code base will be replaced by the drone 1.0.0 code soonish.

You could check if github.com/drone/go-scm and github.com/drone/go-login got the same ineffassign issues to get it fixed for the next upcoming version.

saromanov:
ok, np

Issue #2596, opened 2019-02-18
Streams leaking resources


luis-silva:
I'm raising this issue and a PR (#2597) to address memory and goroutine leaks in drone server.

### The Problem

I've noticed that, when using Streams, Drone was leaking go routines and heap space. I was able to reproduce it by doing the following:
- Opening a new tab and access Drone
- Refreshing the page multiple times (or opening multiple tabs)
- Analysing the resource usage by accessing `/api/debug/pprof/`, `go tool pprof`, debug logs and chrome developer tools.

The logs showed multiple connections being opened but never closed, just by refreshing the page
![connections](https://user-images.githubusercontent.com/4629568/52945024-b357c580-3368-11e9-927a-59cee9da76c9.png)

Memory tracking showed that the heap never got garbage collected
![memory_pre](https://user-images.githubusercontent.com/4629568/52945400-8a840000-3369-11e9-8799-de6250d22385.png)

Go routines would never terminate
<img width="2400" alt="goroutines" src="https://user-images.githubusercontent.com/4629568/52945604-01b99400-336a-11e9-8c57-5dda43b8f08b.png">

Go routines and heap usage confirmation in pprof endpoint
![pprof](https://user-images.githubusercontent.com/4629568/52945440-a4254780-3369-11e9-86a9-e71e83ee94f9.png)

Idle streams not getting closed:
<img width="530" alt="stream duration" src="https://user-images.githubusercontent.com/4629568/52945765-64ab2b00-336a-11e9-806d-1836eea62822.png">

### Analysis

When profiling I found the two main issues to be:

**Timers memory leak**
<img width="288" alt="timer_leak" src="https://user-images.githubusercontent.com/4629568/52947069-9c67a200-336d-11e9-9dd9-cab454744523.png">

The 60 minute timer to close the stream was not actually working. This was due to the fact that a new timer was being created for every iteration of the for loop, which happened every 30 seconds when the ping timeout ran, which meant that new 60 minute timers were created every 60 seconds. This would also prevent the old timers from being garbage collected as Go won't consider them unused until they're triggered.

I solved that issue by creating the timers outside of the for loop and properly reseting them only when data was flowing through the stream.

**Stream not being closed**
<img width="435" alt="repolist" src="https://user-images.githubusercontent.com/4629568/52947459-accc4c80-336e-11e9-97f0-c2f1c15ff2e9.png">
[...]

PR #2597, opened 2019-02-18
Cleanup idle streams and release memory

luis-silva:
This PR addresses the problems raised in Issue #2596.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -54,12 +54,13 @@ func EventStreamSSE(c *gin.Context) {
 	io.WriteString(rw, ": ping\n\n")
 	flusher.Flush()
 
-	logrus.Debugf("user feed: connection opened")
-
 	user := session.User(c)
 	repo := map[string]bool{}
 	if user != nil {
-		repos, _ := store.FromContext(c).RepoList(user)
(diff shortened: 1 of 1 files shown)


tboerger:
Great that you gathered these leaks, but there are no plans to release another 0.8 version, instead it's focused on a final 1.0.0 release and encourage the people to upgrade.

luis-silva:
@tboerger I don't follow you reasoning. 
0.8 is the current major stable release so it should be maintained until, at the very least, a new stable release is out (and ideally a bit longer while businesses apply the migration plans to the new version).
The source code for 1.0 is not even public yet so there's no way to know if this issue is also present there.
I supplied a good deal of information in the related issue and already created the PR to fix the problem so this should only require a simple review and minor release and not extra coding efforts on your part.

bradrydzewski:
Thanks for the pull request and the very thoughtful and thorough analysis.  I have merged into a feature branch and pushed an image that teams can start testing if they are experiencing this issue. The image is `drone/drone:0.8.9-pr.2597`

> The source code for 1.0 is not even public yet so there's no way to know if this issue is also present there.

The event stream code was re-written in 1.0 and does not use a timer, so 1.0 would not be subject to this issue.

PR #2598, opened 2019-02-19
Updated user model comments

patrickjahns:
Updated the comments for the user model to reflect that the admin flag is again saved in the database


tboerger:
I have already told him that it's still the 0.8 source :P

PR #2600, opened 2019-02-21
feat(s3): allow user to specify s3 endpoint for non aws usage

JordanSussman:
This PR does the following:

- Updates the Dockerfiles to be based on a newer version of alpine
- Adds the `DRONE_S3_ENDPOINT` environment variable to allow users to specify the s3 endpoint to connect to
  - This is needed for services like [DigitalOcean spaces](https://www.digitalocean.com/products/spaces/) and other s3 compatible services
  - You could argue that the AWS library should support something like `AWS_ENDPOINT`, however, historically they haven't been willing to do this for any language or even the [cli](https://aws.amazon.com/cli). For example, https://github.com/aws/aws-sdk-php/issues/1239 has been open for almost two years.

**Note:** I've tested this update on AWS and DigitalOcean spaces.

### DIFF
--- a/cmd/drone-server/config/config.go
+++ b/cmd/drone-server/config/config.go
@@ -330,8 +330,9 @@ type (
 
 	// S3 provides the storage configuration.
 	S3 struct {
-		Bucket string `envconfig:"DRONE_S3_BUCKET"`
-		Prefix string `envconfig:"DRONE_S3_PREFIX"`
+		Bucket   string `envconfig:"DRONE_S3_BUCKET"`
+		Prefix   string `envconfig:"DRONE_S3_PREFIX"`
+		Endpoint string `envconfig:"DRONE_S3_ENDPOINT"`
 	}
 
 	// HTTP provides http configuration.
--- a/cmd/drone-server/inject_store.go
+++ b/cmd/drone-server/inject_store.go
@@ -85,6 +85,7 @@ func provideLogStore(db *db.DB, config config.Config) core.LogStore {
 	return logs.NewS3Env(
 		config.S3.Bucket,
 		config.S3.Prefix,
+		config.S3.Endpoint,
 	)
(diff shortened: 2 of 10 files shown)


bradrydzewski:
the build failure is a problem with the Go modules. I will patch this tomorrow morning.

tboerger:
Tools like Minio also require path style being set to true, otherwise it won't work.

bradrydzewski:
ok, thanks for this!  do you have an example environment variables for digital ocean spaces? I would love to include them in the docs.

bradrydzewski:
@tboerger I do not think digital ocean requires path style, but for minio we can definitely enable with a separate pull request. I think we will need to add a `DRONE_S3_SKIP_VERIFY` and `DRONE_S3_PATH_STYLE` variables, like this:
https://github.com/minio/cookbook/blob/master/docs/aws-sdk-for-go-with-minio.md

JordanSussman:
> ok, thanks for this! do you have an example environment variables for digital ocean spaces? I would love to include them in the docs.

The instructions for DigitalOcean Spaces is almost identical to AWS, you just need to specify `DRONE_S3_ENDPOINT=<region>.digitaloceanspaces.com`. For example, my space (the digitalocean term for bucket) is located in the digitalocean region of sfo2, so I would set `AWS_REGION=sfo2` and `DRONE_S3_ENDPOINT=sfo2.digitaloceanspaces.com`. 

As with AWS you still need to set `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `DRONE_S3_BUCKET` for everything to work. The credentials for `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` can be generated on the [api page](https://cloud.digitalocean.com/settings/api/tokens).

PR #2603, opened 2019-02-21
Fix check to avoid running agent on server

oldmantaiter:
- This was causing the agent to spin up on the server even though
  it was not enabled


bradrydzewski:
I think you might be misunderstanding.  `Agent.Enabled` enables remote agents.  When `Nomad`, `Kubernetes` and `Agent` are disabled, Drone starts a local build runner. I am pretty sure it is working as expected.

oldmantaiter:
Ah, that makes more sense. I'll close.

bradrydzewski:
also just wanted to say that I still very much appreciate that you took the time to send a pull request, even if it didn't make sense to merge 🙂

PR #2621, opened 2019-03-08
change build_author_name charset to utf8mb4

khanhicetea:
Fix the issue when build_author_name is UTF8 string (Github fullname). (Only MySQL)

This is error log from github webhook:

```
{"message":"Error 1366: Incorrect string value: '\\xE1\\xBB\\x85n H...' for column 'build_author_name' at row 1"}
```

Thanks!

### DIFF
--- a/store/shared/migrate/mysql/files/004_create_table_builds.sql
+++ b/store/shared/migrate/mysql/files/004_create_table_builds.sql
@@ -22,7 +22,7 @@ CREATE TABLE IF NOT EXISTS builds (
 ,build_source        VARCHAR(500)
 ,build_target        VARCHAR(500)
 ,build_author        VARCHAR(500)
-,build_author_name   VARCHAR(500)
+,build_author_name   VARCHAR(500) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci
 ,build_author_email  VARCHAR(500)
 ,build_author_avatar VARCHAR(1000)
 ,build_sender        VARCHAR(500)


khanhicetea:
Manual migration for those who used MySQL

```sql
ALTER TABLE builds MODIFY build_author_name VARCHAR(500) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```

tboerger:
Maybe it makes more sense to create the whole table in utf8mb4 format.

khanhicetea:
Yes ! MySQL 8.0 has default charset utf8mb4. I used this for docker compose MySQL 5.7

```
image: mysql:5.7
command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
```

tboerger:
But how to handle setups which doesn't have enabled utf8mb4?

khanhicetea:
I think we should have a migration sql file instead, so old versions will get upgrade

bradrydzewski:
> I think we should have a migration sql file instead, so old versions will get upgrade

I do not think this is necessary because a) this does not impact most Drone users and b) this only impacts 1.0 which is not released yet and has a very small number of installations. Most people have not migrated to 1.0 yet. Teams that migrate to 1.0 that need this column to be utf8mb4 can manually run ALTER TABLE.

> Maybe it makes more sense to create the whole table in utf8mb4 format.

We force specific columns to be UTF8 so this pull request is consistent with the existing approach. It is also possible to force UTF8 at the database level as mentioned above. 

```
image: mysql:5.7
command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
```

Issue #855, opened 2015-01-27
GitLab Refresh Tokens & Concurrency


bradrydzewski:
When the `access_token` is refreshed GitLab also sends a new `refresh_token`. This is a perfectly valid (but optional) implementation detail:

> the authorization server MAY issue a new refresh token, in which case the client MUST discard the old refresh token and replace it with the new refresh token.

When GitLab sends back a new `refresh_token` it also immediately expires the old `refresh_token`. This is also a perfectly valid (but optional) implementation detail:

> The authorization server MAY revoke the old refresh token after issuing a new refresh token to the client.

This becomes problematic for systems that are making concurrent requests to the API. It is possible for two threads (or in our case goroutines) to successfully refresh an access_token around the same time. This means one thread will have an expired set of tokens that it cannot refresh.

This gets more complicated because we are making these requests offline. This means if we achieve the above race condition, we could get locked out of the API forever until the user physically goes to the website and re-authenticates.

There really isn't a good way to handle this in a concurrent, offline environment. We could try to implement mutexes, but even this is subject to edge cases (ie program quits un-unexpectedly)

I believe we need to engage GitLab for this. One example of a middle ground could be providing `access_type=offline` (similar to Google) that would prevent sending a new `refresh_token` when refreshing an `access_token`.

Another option would be to more closely mirror GitHub and never expire an access token. It looks like this is possible (see https://github.com/doorkeeper-gem/doorkeeper/pull/76) but I'm not sure they would be willing to make such a drastic policy change.

Intrepidd:
I think we stumbled upon this. After recently upgrading drone, we had issues with build not launching.

After digging into the logs and source code, we found that the owner of the repository had to log in again.

Problem is, this issue happens several times a day. I imagine we're having the race condition you're describing

bradrydzewski:
gitlab removed refresh tokens by default, so no longer applicable

Intrepidd:
Indeed, after updating gitlab the issue is now gone, thanks

PR #2622, opened 2019-03-08
Add Gitea Oauth

techknowlogick:
With fallback to basic auth if Gitea server does not support oauth

Linked to https://github.com/drone/go-login/pull/3


bradrydzewski:
awesome!  one question: how does gitea handle the scenario where two separate go routines try to refresh the access token at the same time?  this is something that caused issues in early gitlab oauth implementations, so just want to understand a bit more :)

techknowlogick:
Tests fail because the above linked PR isn't merged

bradrydzewski:
I just merged :)

techknowlogick:
Thanks @bradrydzewski 😄

techknowlogick:
@jonasfranz could you answer @bradrydzewski's question above?

techknowlogick:
Marking this as WIP as a patch needs to be sent to Gitea so that if no redirect_uri is set then it'll default to what is set.

techknowlogick:
@bradrydzewski looking through Gitea OAuth code, refreshing token isn't implemented in Gitea (yet). I was able to compile drone and login with this PR and it pulled all the repos. I've also been able to submit an updated PR to go-login with being able to use RedirectURL.

cmd/drone-server/config/config.go
```diff
@@ -293,9 +293,11 @@ type (
 
 	// Gitea provides the gitea client configuration.
 	Gitea struct {
-		Server     string `envconfig:"DRONE_GITEA_SERVER"`
-		SkipVerify bool   `envconfig:"DRONE_GITEA_SKIP_VERIFY"`
-		Debug      bool   `envconfig:"DRONE_GITEA_DEBUG"`
+		ClientID     string `envconfig:"DRONE_GITEA_CLIENT_ID"`
```
appleboy:
```go
	Gitea struct {
		Server       string   `envconfig:"DRONE_GITEA_SERVER" default:"https://try.gitea.io"`
		ClientID     string   `envconfig:"DRONE_GITEA_CLIENT_ID"`
		ClientSecret string   `envconfig:"DRONE_GITEA_CLIENT_SECRET"`
		SkipVerify   bool     `envconfig:"DRONE_GITEA_SKIP_VERIFY"`
		Scope        []string `envconfig:"DRONE_GITEA_SCOPE" default:"repo,repo:status,user:email,read:org"`
		Debug        bool     `envconfig:"DRONE_GITEA_DEBUG"`
	}
```

cmd/drone-server/inject_login.go
```diff
@@ -91,6 +92,16 @@ func provideGiteaLogin(config config.Config) login.Middleware {
...
 		return nil
 	}
+	if config.Gitea.ClientID != "" {
+		return &gitea.Config {
+			ClientID:     config.Gitea.ClientID,
+			ClientSecret: config.Gitea.ClientSecret,
+			Server:       config.Gitea.Server,
+			Client:       defaultClient(config.Gitea.SkipVerify),
+			Logger:       logrus.StandardLogger(),
+			RedirectURL:  config.Server.Addr + "/login",
```
appleboy: Add

```go
Scope:        config.Gitea.Scope,
```

cmd/drone-server/config/config.go
techknowlogick: Thanks for feedback 😄 I've made this change.

appleboy:
also update the `provideClient` func:

```diff
    case config.Github.ClientID != "":
        return provideGithubClient(config)
-    case config.Gitea.Server != "":
+    case config.Gitea.ClientID != "":
        return provideGiteaClient(config)
    case config.GitLab.ClientID != "":
        return provideGitlabClient(config)
```

techknowlogick:
@appleboy I kept `config.Gitea.ClientID != ""` codition so this doesn't break for Gitea users < 1.8 so they can still use username/password

cmd/drone-server/inject_login.go
techknowlogick: done

appleboy:
@techknowlogick Good catch. Thanks for the PR. LGTM.

jonasfranz:
Refresh times are implemented. You can only use an access token or refresh token once to redeem a new token.

bradrydzewski:
For context, GitLab had a similar restriction a few years ago that they since removed. Here are some issues that we ran into https://github.com/drone/drone/issues/855

cmd/drone-server/config/config.go
```diff
@@ -293,9 +293,12 @@ type (
...
 	// Gitea provides the gitea client configuration.
 	Gitea struct {
-		Server     string `envconfig:"DRONE_GITEA_SERVER"`
-		SkipVerify bool   `envconfig:"DRONE_GITEA_SKIP_VERIFY"`
-		Debug      bool   `envconfig:"DRONE_GITEA_DEBUG"`
+		Server       string   `envconfig:"DRONE_GITEA_SERVER" default:"https://try.gitea.io"`
+		ClientID     string   `envconfig:"DRONE_GITEA_CLIENT_ID"`
+		ClientSecret string   `envconfig:"DRONE_GITEA_CLIENT_SECRET"`
+		SkipVerify   bool     `envconfig:"DRONE_GITEA_SKIP_VERIFY"`
+		Scope        []string `envconfig:"DRONE_GITEA_SCOPE" default:"repo,repo:status,user:email,read:org"`
```
jonasfranz: Gitea does not support scopes currently

techknowlogick: I decided that a documentation change is better than a code change for when scope is supported

tboerger:
Great to see that this gets finally integrated

cmd/drone-server/config/config.go
```diff
@@ -293,9 +293,12 @@ type (
 
 	// Gitea provides the gitea client configuration.
 	Gitea struct {
-		Server     string `envconfig:"DRONE_GITEA_SERVER"`
-		SkipVerify bool   `envconfig:"DRONE_GITEA_SKIP_VERIFY"`
-		Debug      bool   `envconfig:"DRONE_GITEA_DEBUG"`
+		Server       string   `envconfig:"DRONE_GITEA_SERVER" default:"https://try.gitea.io"`
```
bradrydzewski: this default needs to be removed. An empty string is the only way we know if gitea is configured or not.

techknowlogick: Good catch. Thanks :)

I've updated.

appleboy:
@techknowlogick conflicts.

jonasfranz:
@bradrydzewski do you propose to remove refresh token invalidation after a new token is signed?

bradrydzewski:
@jonasfranz yes, I think ideally invalidating existing refresh tokens after a new refresh token is generated would be opt-in.  Most oauth providers (GitHub, Google, GitLab, etc) only invalidate refresh tokens if the user revokes access to the application.

lunny:
@techknowlogick `go.mod` conflicted

jonasfranz:
@bradrydzewski With my latest PR, Gitea will not invalidate refresh tokens by default. But there is an option to enable token invalidation.

bradrydzewski:
@jonasfranz awesome, can you link to the pull request here. Once it is merged, we can merge this pull request. Also I will need to document the minimum version of Gitea that will be required -- can you let me know which version this will be? Thanks!!

jonasfranz:
@bradrydzewski The PR is already merged: https://github.com/go-gitea/gitea/pull/6584

It will be in version 1.8.0 (or 1.8.0-rc3 if we have a rc3).

techknowlogick:
@lunny conflict resolved.

appleboy:
@bradrydzewski Already merged in Gitea.

bradrydzewski:
excellent. one last request -- can we get an entry for this in the changelog? thanks!

lunny:
@bradrydzewski I think yes. This could be in release notes of v1.8.

bradrydzewski:
what I mean is that we need an entry for this change in Drone's changelog file https://github.com/drone/drone/blob/master/CHANGELOG.md

techknowlogick:
@bradrydzewski added

tboerger:
This works with gitea master? Or also with some stable release?

techknowlogick:
@tboerger it works with 1.8.0 which was just released yesterday, but Drone can still work with older versions of gitea using basic auth.

oliver006:
I've been using it successfully with docker images gitea:v1.8.0-rc3 (and 1.8.0 now that it's released) and drone/drone@sha256:7acfbba60ac6c35537b099ce612aac0af3cdbfa2e48d0ee96730cd4c904b89fa (the sha of the image right after this PR got merged).

But there seems to be a little problem with refreshing the gitea token. After a while, I'm no longer to e.g. enable a gitea repo for drone. I'm still logged in and can modify settings within drone. Once I log out and back in, everything is fine.

Should I open a separate issue?

techknowlogick:
@oliver006 between rc3 and final gitea released a patch that should fix the issue you described above. Please feel free to open an issue in gitea tracker if this comes up again.

oliver006:
Thanks for the quick response @techknowlogick . I tried with Gitea 1.8.0 and the drone/drone image I mentioned above but it's still not working. I get:

image

I checked in the DB, the `user_oauth_expiry` value is unchanged (and 2h in the past).

I'll open an issue in the Gitea project.

bradrydzewski:
I think the issue is that if Gitea uses a refresh token, a refresher needs to be configured. See https://github.com/drone/drone/blob/master/cmd/drone-server/inject_login.go#L166

frostieDE:
This features needs some documentation, or am I missing something on the docs? :smile:

ukos-git:
here are my quick notes for installing:

- go to your gitea instance application settings page: `https://gitea.io/user/settings/applications`
- add a oauth instance with arbitrary id and redirect to the login page: `https://gitea.io/login`

Screenshot from 2019-04-28 15-40-02

- you will get ClientID and ClientSecret:

Screenshot from 2019-04-28 15-41-13

- add the generated values to your drone environement:

```console
DRONE_GITEA_CLIENT_ID=XXX
DRONE_GITEA_CLIENT_SECRET=XXX
```

be sure to use latest `gitea/gitea` and `drone/drone` images (gitea version is displayed in ui bottom left. Today is Gitea Version: 345bc06)

Probably not enough for a PR to docs but better than no documentation.

PR #2624, opened 2019-03-14
Update windows dockerfiles and add 1809 versions

donny-dont:


tboerger:
Sorry, wanted to ask for the server, not controller.

donny-dont:
Brad said no server just agent. Only did controller cause it was there already.

PR #2631, opened 2019-03-21
Add postgres and mysql support to OSS version

### LINKED ISSUE #2630: Unable to execute db migrations in mariadb
running drone using following command against clean mariadb database 
Server version: 10.1.38-MariaDB MariaDB Server

```
docker run \
  --volume=/var/run/docker.sock:/var/run/docker.sock \
  --volume=/var/lib/drone:/data/dronefiles \
  --env=DRONE_GITHUB_SERVER=https://github.com \
  --env=DRONE_GITHUB_CLIENT_ID=xxx \
  --env=DRONE_GITHUB_CLIENT_SECRET=xx \
  --env=DRONE_RUNNER_CAPACITY=10 \
  --env=DRONE_SERVER_HOST=drone.foobar.com \
  --env=DRONE_SERVER_PROTO=https \
  --env=DRONE_GIT_ALWAYS_AUTH=true \
  --env=DRONE_SERVER_PORT=:443 \
  --env=DRONE_TLS_CERT=/var/lib/drone/key.crt \
  --env=DRONE_TLS_KEY=/var/lib/drone/key.pem \
  --env=DRONE_DATABASE_DRIVER=mysql \
  --env=DRONE_DATABASE_DATASOURCE=xxxx:xxxx@tcp\(172.17.0.1:3306\)/drone?parseTime=true \
  --publish=80:80 \
  --publish=443:443 \
  --restart=always \
  --detach=true \
  --name=drone \
  jesseh/drone:1-oss
```

{"error":"Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'user_login TEXT\n,user_email TEXT\n,user_admin BOOLEAN\n,us' at line 3","level":"fatal","msg":"main: cannot initialize server","time":"2019-03-21T13:16:28Z"}

I added users table manually, but the problem seems to be in all tables. So when starting new drone installation, it cannot create other tables than `migrations`

zetaab:
fixes #2630 

@bradrydzewski the main question now is that should there be support for postgre and mysql in OSS version? The code currently says that its not supported, but this page https://drone.io/enterprise/features/  says that opensource version should have support? So if there should be support, feel free to review otherwise you can just close this PR.


zetaab:
ah okay you just updated that page

PR #2636, opened 2019-03-26
Better license errors

pnovotnak:
The error message when a license is bad doesn't contain relevant context to help the user debug. This PR adds that context.

### DIFF
--- a/cmd/drone-server/inject_license.go
+++ b/cmd/drone-server/inject_license.go
@@ -38,7 +38,7 @@ func provideLicense(client *scm.Client, config config.Config) *core.License {
 		l = license.Trial(client.Driver.String())
 	} else if err != nil {
 		logrus.WithError(err).
-			Fatalln("main: invalid or expired license")
+			Fatalf("main: invalid or expired license at '%s'", config.License)
 	}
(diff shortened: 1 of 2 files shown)


tboerger:
IMHO it's not a good idea to print the license to the log.

pnovotnak:
Even the file path?

cmd/drone-server/inject_license.go
```diff
@@ -38,7 +38,7 @@ func provideLicense(client *scm.Client, config config.Config) *core.License {
 		l = license.Trial(client.Driver.String())
 	} else if err != nil {
 		logrus.WithError(err).
-			Fatalln("main: invalid or expired license")
```
bradrydzewski: I would include the filename as a metadata field:

```diff
		logrus.WithError(err).
+			WithField("path", config.License).
			Fatalln("main: invalid or expired license")
```

tphoney:
Hi @pnovotnak you have not made the suggested change from @bradrydzewski. Thanks for your work but we are going to close this issue. If you still want this change, rebase and we can re-review it. 
Thanks you for your patience

PR #2639, opened 2019-03-28
adds DRONE_RUNNER_ANNOTATIONS config bit

whereismyjetpack:
This PR adds the DRONE_RUNNER_ANNOTATIONS config bit to be used by the kubernetes Runner. 

This is my first attempt at go, and I apologize in advance for any conventions that may be improper. I figured starting the conversation would be better than not trying at all

This work is to help close out https://github.com/drone/drone-runtime/issues/38

### DIFF
--- a/cmd/drone-controller/config/config.go
+++ b/cmd/drone-controller/config/config.go
@@ -82,20 +82,21 @@ type (
 
 	// Runner provides the runner configuration.
 	Runner struct {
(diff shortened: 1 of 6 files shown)


bradrydzewski:
the kubernetes runtime in this repository was deprecated.  the second generation implementation of the kubernetes runner can be found at https://github.com/drone-runners/drone-runner-kube

the second iteration does not yet support global annotations, so this pull request could be opened against the new repository.

PR #2647, opened 2019-04-06
Fix typo

posquit0:
Just typo issues


bradrydzewski:
thanks, however, this is not a typo. [Unknwon](https://github.com/Unknwon) is the creator of gogs :)

PR #2649, opened 2019-04-08
Drone Job Resource Limits / Requests

gambol99:
- adding environment variables allowing use to set the limits and requests on the batch job
- namely as it was inheriting out default limits and getting oom at 400Mi

@note, i'm not sure if `JOB` is the right description for this? ..


geekgonecrazy:
I can see this being super useful.  👍  

Job feels appropriate IMO

bradrydzewski:
we do something similar for Nomad [1] and we use `RequestMemory` and `RequestCPU`. I recommend using the same variables for Kubernetes for consistency reasons.

[1] https://github.com/drone/drone/blob/master/scheduler/nomad/nomad.go#L89:L94

bradrydzewski:
this capability is available in the new kubernetes runner home, which can be found at https://github.com/drone-runners/drone-runner-kube

geekgonecrazy:
Nice I like the organization into runners 👍 looks like i'll be switching my drone install before long :)

PR #2661, opened 2019-04-12
Fix automated legacy matrix convert

tboerger:
We got to move the envsubst call behind the legacy convert, otherwise
the matrix convert gets broken because it fails to replace the matrix
variables as they don't exist anymore.

### DIFF
--- a/operator/runner/runner.go
+++ b/operator/runner/runner.go
@@ -184,23 +184,24 @@ func (r *Runner) Run(ctx context.Context, id int64) error {
 	// TODO mutate the yaml
 	//
 
-	y, err := envsubst.Eval(string(m.Config.Data), func(name string) string {
+	// this code is temporarily in place to detect and convert
+	// the legacy yaml configuration file to the new format.
+	y, err := converter.ConvertString(string(m.Config.Data), converter.Metadata{
+		Filename: m.Repo.Config,
+		Ref:      m.Build.Ref,
+	})
+
+	if err != nil {
+		return err
+	}
+
+	y, err = envsubst.Eval(y, func(name string) string {
 		env := environ[name]
 		if strings.Contains(env, "\n") {
 			env = fmt.Sprintf("%q", env)
 		}
 		return env
 	})
-
-	// this code is temporarily in place to detect and convert
-	// the legacy yaml configuration file to the new format.
-	y, err = converter.ConvertString(y, converter.Metadata{
-		Filename: m.Repo.Config,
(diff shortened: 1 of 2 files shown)


bradrydzewski:
the only concern I have is order of operations. The unmarshal could fail if a `${variable}` is used in a field that is not of type string. thinking ...

tboerger:
Could you explain a little bit more? The legacy matrix convert doesn't replace variables except the matrix variables which is the right behavior, after that envsubst replaces all remaining variables. 

The current behavior replaces the matrix variables before they get converted, that way not any matrix pipeline can get replaced.

The test suite for the matrix conversion within drone-yaml is working totally fine, but it is not calling envsubst at all, that's why the library itself is not affected by that.

operator/runner/runner.go
```diff
@@ -184,24 +184,25 @@ func (r *Runner) Run(ctx context.Context, id int64) error {
...
-		if strings.Contains(env, "\n") {
-			env = fmt.Sprintf("%q", env)
-		}
-		return env
-	})
-
 	// this code is temporarily in place to detect and convert
 	// the legacy yaml configuration file to the new format.
-	y, err = converter.ConvertString(y, converter.Metadata{
+	y, err := converter.ConvertString(string(m.Config.Data), converter.Metadata{
```
bradrydzewski: can avoid the extra string conversion:

```diff
-y, err := converter.ConvertString(string(m.Config.Data), converter.Metadata{
+y, err := converter.Convert(m.Config.Data, converter.Metadata{
```

bradrydzewski:
the only concern I have is the hypothetical situation where the matrix parameters include a non-string value like this:

```
matrix:
- IMAGE: alpine:3.8
  PULL: true
```

which is interpolated like this:

```
image: ${IMAGE}
pull: ${PULL}
```

In the above example, the yaml will fail to parse because string literal `${PULL}` cannot be unmarshaled into type boolean, which is why interpolation needs to happen before.  But perhaps this is only a hypothetical situation and not a real-world problem?

tboerger:
At least I can say that our crazy owncloud pipeline seem to start correctly :)

bradrydzewski:
> At least I can say that our crazy owncloud pipeline seem to start correctly :)

fair enough, I think it is more of a hypothetical issue and has a low probability of causing any issues. I am good to merge once we have an entry in the changelog :)

tboerger:
Apply changelog and comment within https://github.com/drone/drone/pull/2674

PR #2684, opened 2019-04-22
Refresh Gitea Oauth2 token

techknowlogick:


jonasfranz:
@techknowlogick that's the correct URL

techknowlogick:
Thanks :)

cmd/drone-server/inject_login.go
```diff
@@ -172,5 +172,13 @@ func provideRefresher(config config.Config) *oauth2.Refresher {
 			Source:       oauth2.ContextTokenSource(),
 		}
 	}
+	if config.Gitea.ClientID != "" {
```
tboerger: Maybe add a switch instead of another if?

techknowlogick: Updated

cmd/drone-server/inject_login.go
```diff
@@ -164,13 +164,21 @@ func provideStashLogin(config config.Config) login.Middleware {
...
 			ClientID:     config.Bitbucket.ClientID,
 			ClientSecret: config.Bitbucket.ClientSecret,
 			Endpoint:     "https://bitbucket.org/site/oauth2/access_token",
 			Source:       oauth2.ContextTokenSource(),
 		}
+	case config.Gitea.Server != "":
+		return &oauth2.Refresher{
+			ClientID:     config.Gitea.ClientID,
+			ClientSecret: config.Gitea.ClientSecret,
+			Endpoint:     config.Server.Addr + "/login/oauth/access_token",
```
bradrydzewski: this should be `config.Gitea.Server`

cmd/drone-server/inject_login.go
```diff
@@ -164,13 +164,21 @@ func provideStashLogin(config config.Config) login.Middleware {
...
-	if config.Bitbucket.ClientID != "" {
+	switch {
+	case config.Bitbucket.ClientID != "":
 		return &oauth2.Refresher{
 			ClientID:     config.Bitbucket.ClientID,
 			ClientSecret: config.Bitbucket.ClientSecret,
 			Endpoint:     "https://bitbucket.org/site/oauth2/access_token",
 			Source:       oauth2.ContextTokenSource(),
 		}
+	case config.Gitea.Server != "":
```
bradrydzewski: I recommend changing this to `config.Gitea.ClientID != ""` since we only want to enable the refresher if oauth is configured.

cmd/drone-server/inject_login.go
```diff
@@ -164,13 +164,21 @@ func provideStashLogin(config config.Config) login.Middleware {
 // provideRefresher is a Wire provider function that returns
 // an oauth token refresher for Bitbucket.
```
oliver006: Change comment to `... refresher for Bitbucket and Gitea` ?

techknowlogick: updated. Thanks :)

cmd/drone-server/inject_login.go
techknowlogick: updated, thanks :)

cmd/drone-server/inject_login.go
techknowlogick: good catch. I've updated, thanks :)

techknowlogick:
My non-scientific testing, by setting lifetime of oauth2 token to 100seconds then forcing drone to do api calls (by activating repos, and sync'ing repo list) after lifetime is over shows that drone will request a new token, and then use it.

bradrydzewski:
awesome, merging this now so it can be included in our 1.1.0 release tomorrow

PR #2686, opened 2019-04-24
Add DRONE_PROMETHEUS_ANONYMOUS_ACCESS configuration option

JanBerktold:


bradrydzewski:
:wave: hey there, have you considered creating a machine account for prometheus access? https://docs.drone.io/administration/user/machine/

bradrydzewski:
actually, this url is better https://docs.drone.io/administration/server/metrics/

JanBerktold:
Hey @bradrydzewski, we have considered that but we already have Prometheus setup to scrape all Nomad jobs (with a specific tag) for metrics and having to change the central prometheus instance for a single job does not seem ideal and would cause challenges. Our load balancers ensures that no non-internal user can hit the /metrics endpoint to prevent the metrics from leaking to the outside, having unauthenticated access within our system is therefore the preferred option.

metric/handler.go
```diff
@@ -26,13 +27,15 @@ var errAccessDenied = errors.New("Access denied")
 type Server struct {
 	metrics http.Handler
 	session core.Session
+	config  config.Config
 }
 
 // NewServer returns a new metrics server.
-func NewServer(session core.Session) *Server {
+func NewServer(session core.Session, config config.Config) *Server {
```
bradrydzewski: we typically try to avoid passing the entire config into different packages. So in this case I recommend the following for consistency with our other packages:

```
-func NewServer(session core.Session, config config.Config)
+func NewServer(session core.Session, anonymous bool)
```

metric/handler.go
```diff
@@ -26,13 +27,15 @@ var errAccessDenied = errors.New("Access denied")
 type Server struct {
 	metrics http.Handler
 	session core.Session
+	config  config.Config
```
bradrydzewski: explained below, we can just store the boolean

```
type Server struct {
  metrics http.Handler
  session core.Session
- config  config.Config
+ anonymous bool
}
```

metric/handler.go
bradrydzewski: note that this will require modifying https://github.com/drone/drone/blob/master/cmd/drone-server/inject_server.go and creating a provider function. For example:

```diff
// wire set for loading the server.
var serverSet = wire.NewSet(
	manager.New,
-	metric.NewServer,
	api.New,
	web.New,
+	provideMetrics,
	provideRouter,
	provideRPC,
	provideServer,
	provideServerOptions,
)
```

bradrydzewski:
gotcha ... I have no problem with optionally opening up the prometheus endpoint.  I've added some minor comments around the implementation, primarily to ensure consistency with other packages and how we pass around configuration parameters.

JanBerktold:
@bradrydzewski Sounds good, working on addressing those right now and also adding a test case. I am unclear on how to inject the bool into metric.NewServer - I haven't worked with wire before but it fails to `inject InitializeApplication: no provider found for bool` which makes sense. Should I just make it an exported field and set it afterwards?

bradrydzewski:
you will need to modify this file:
https://github.com/drone/drone/blob/master/cmd/drone-server/inject_server.go

configure to load the metrics server from a custom function:

```diff
// wire set for loading the server.
var serverSet = wire.NewSet(
	manager.New,
-	metric.NewServer,
	api.New,
	web.New,
+	provideMetrics,
	provideRouter,
	provideRPC,
	provideServer,
	provideServerOptions,
)
```

in the same file, implement the custom function:

```
func provideMetrics(session core.Session, config config.Config) *metrics.Server {
  return metrics.New(session, config.Prometheus.AnonymousAccess)
}
```

something like that ... :)

JanBerktold:
@bradrydzewski Thanks for the pointers! Sorry for that, haven't worked with wire before and I was a bit blind. :) 

I have addressed the feedback and added a test. It appears that the CI build is failing, but that seems unrelated to my changes. I have confirmed this by pulling the master branch and running the tests locally - they also fail off the unmodified branch.

bradrydzewski:
huh strange, last push to master was passing https://cloud.drone.io/drone/drone/127. I will check out the logs and see what I can find.

bradrydzewski:
looks maybe like a race condition in one of the unit tests which is why this is intermittent ... I see a nil pointer. I will take care of this and push an update to master, and re-run these tests.

JanBerktold:
I spoke too soon - Looks like I had a compilation issue because I didn't regenerate the file. (a git pre-commit hook would be nice).

Unclear on how that could cause a runtime issue though.

bradrydzewski:
> Unclear on how that could cause a runtime issue though.

the failed test came from a package untouched by your change, and looks like a race condition, which would explain the intermittent errors. I will handle it separately.

JanBerktold:
Looks like we're good here :)

cmd/drone-server/config/config.go
```diff
@@ -162,6 +162,11 @@ type (
 		Text   bool `envconfig:"DRONE_LOGS_TEXT"`
 	}
 
+	// Prometheus provides the prometheus configuration.
+	Prometheus struct {
+		EnableAnonymousAccess bool `envconfig:"DRONE_PROMETHEUS_ANONYMOUS_ACCESS" default:"false"`
```
tboerger: If you name it AnonymousAcces you don't have to set the envconfig tag.

bradrydzewski:
@JanBerktold we received your information request, but the email bounced back. It looks like the mailing list could be restricted. I would be happy to chat, but just need a way to get in touch :)  Perhaps you can shoot a message to brad at drone.io?

badsmoke:
why is this not documented?

I have been looking for it for ages

PR #2687, opened 2019-04-26
Use branch form value as fallback for last build

tboerger:
The current version of the Drone CLI is still using the branch form
value to filter the latest build information while the server is
listening only for the ref form value. With this fix we are falling back
to branch form value if it gets defined, formats it as a ref and hands
it to the further functions.


tboerger:
For reference the the current Go library: https://github.com/drone/drone-go/blob/8dcd49ee509e4753547bb738b4c38e4da3b4f72e/drone/client.go#L236-L238

Issue #2693, opened 2019-05-03
Drone for Kubernetes not deleting local volume when finished


chrisghill:
This issue is discussed here: https://discourse.drone.io/t/drone-leaving-directories-behind-in-kubernetes/4395

To summarize: the code exists for the controller drone-job to remove the `/tmp/drone/<namespace>` directory after a build is finished, but the drone job isn't created with a volume mount for the `/tmp/drone` directory, so it is ineffective.

I'm mostly creating this issue to track a branch and link a PR.

PR is here: https://github.com/drone/drone/pull/2694

bradrydzewski:
closing now that  #2694 is merged

PR #2694, opened 2019-05-03
Fix orphaned /tmp/drone directories in kubernetes

chrisghill:
Details on this issue here: https://discourse.drone.io/t/drone-leaving-directories-behind-in-kubernetes/4395

Issue here: https://github.com/drone/drone/issues/2693

I did what I could to copy the format/structure from here: https://github.com/drone/drone-runtime/blob/master/engine/kube/util.go#L137-L183

### DIFF
--- a/scheduler/kube/kube.go
+++ b/scheduler/kube/kube.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+        "path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


scheduler/kube/kube.go
```diff
@@ -104,6 +105,24 @@ func (s *kubeScheduler) Schedule(ctx context.Context, stage *core.Stage) error {
...
+        mount := v1.VolumeMount{
+		Name:           "local",
+		MountPath:      filepath.Join("/tmp", "drone"),
+	}
+        mounts = append(mounts, mount)
+
+	var volumes []v1.Volume
+        source := v1.HostPathDirectoryOrCreate
+        volume := v1.Volume{
+		Name:            "local",
```
bradrydzewski: is local volume the right storage type? we are using host volume mounts with HostPathDirectoryOrCreate. For reference https://github.com/drone/drone-runtime/blob/master/engine/kube/util.go#L137

chrisghill: @bradrydzewski Is that not what I'm doing right below here in the next 5 lines?  I just named it "local", but we can name it whatever we want.  I tried to replicate the same functionality as the runtime.  I specify the source as v1.HostpathDirectoryOrCreate on 116 and use that as the source on line 122.

bradrydzewski: sorry, I misread the code and thought it was creating a `local` mount instead of a `hostPath` mount. That part looks good. One question I have .... does `local` need to be a unique name? What will happen if two pipelines run on the same machine and try to create `local` volumes?

chrisghill: Good point on the potential name collision.  I don't know if it would cause an issue or not, but I agree we should probably use a more unique or at least explicitly named volume.  The name in the runtime code is obfuscated by the engine spec struct so I wasn't able copy that.  I thought about maybe using the namespace name.  Any other suggestions?

chrisghill: @bradrydzewski I've updated the volume name to be unique.  Please take a look.

bradrydzewski:
hey there, sorry for the delay on my end, but appreciate you pushing this through. thanks!

PR #2699, opened 2019-05-08
Add EphemeralDisk to nomad env configuration

surzycki:
When using nomad as a scheduler, the ephemeral disk allocated to each step is fixed at 300mb.  In most cases this is enough, but there are many in which a large disk is necessary (building docker images, compiling with webpack, etc)

### DIFF
--- a/cmd/drone-server/config/config.go
+++ b/cmd/drone-server/config/config.go
@@ -145,6 +145,7 @@ type (
 		ImagePull   bool     `envconfig:"DRONE_NOMAD_IMAGE_PULL"`
 		Memory      int      `envconfig:"DRONE_NOMAD_DEFAULT_RAM" default:"1024"`
 		CPU         int      `envconfig:"DRONE_NOMAD_DEFAULT_CPU" default:"500"`
+		Disk        int      `envconfig:"DRONE_NOMAD_DEFAULT_DISK" default:"300"`
 	}
 
 	// License provides license configuration
--- a/cmd/drone-server/inject_scheduler.go
+++ b/cmd/drone-server/inject_scheduler.go
@@ -100,6 +100,7 @@ func provideNomadScheduler(config config.Config) core.Scheduler {
 		// LimitCompute:     config.Nomad.CPU,
 		RequestMemory:    config.Nomad.Memory,
 		RequestCompute:   config.Nomad.CPU,
+		EphemeralDisk:    config.Nomad.Disk,
 		CallbackHost:     config.RPC.Host,
 		CallbackProto:    config.RPC.Proto,
 		CallbackSecret:   config.RPC.Secret,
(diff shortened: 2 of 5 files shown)


bradrydzewski:
note that the `drone-controller` container that is scheduled through nomad does not actually perform and build tasks (building docker images, compiling with webpack, etc). The `drone-controller` container launches docker containers out-of-band (e.g. outside of the nomad scheduler) for each step in your pipeline.

surzycki:
@bradrydzewski understood, it is more the nomad job that is scheduled that defaults to 300mb if you don't specify it.  I was running into an issue where I was building a docker image and it was crapping out due to this limitation, nomad allocates the resources that you specify and it schedules based on those resources, cpu, memory, and disk;)

I have been running this fix on our severs for a couple weeks now, and it does the job !

bradrydzewski:
> I was running into an issue where I was building a docker image and it was crapping out due to this limitation [...] I have been running this fix on our severs for a couple weeks now, and it does the job

I think I am still confused because the disk that is allocated for the `drone/controller` container has no bearing on any of the pipeline steps, which are launched as separate docker container __outside__ of the nomad scheduler.

tboerger:
Why are you talking about a controller? For my understanding it just adds a configurable disk size for the drone pipelines in case you need more than the default space for your nomad pipelines.

bradrydzewski:
the reason I am asking about `drone/controller` is because this is the container that is scheduled to execute via nomad. The controller container then uses the host machine docker socket to spawn additional container, out of band.  This pull request sets the disk size for `drone/controller` which has no bearing on the amount of disk available to pipeline containers since these are scheduled outside of nomad.

tboerger:
Ah, I thought you are launching the pipeline steps via nomad.

surzycki:
My understanding of how this works is the following;

1. Drone schedules a Nomad job
2. Within that job, the pipeline is run
3. That Nomad job has a certain number of params that help nomad schedule the job on the server (reserved CPU, reserved memory, and reserved DISK) based on the server resources

These reservations that nomad has limits the environment in which anything can run, so that you can't run a job that is CPU intensive on a nomad client that sucks all the CPU (or memory or DISK).  This would defeat the purpose of the 'smart' scheduling that nomad uses.

bradrydzewski:
> Within that job, the pipeline is run

this is incorrect. All pipeline steps are run _outside_ of the job. Only the controller process, responsible for launching docker containers, collecting logs, etc is run within the nomad job.

surzycki:
I am not sure this is the case, otherwise why was I getting an out of ephemeral disk space error on some of builds. I added this configuration for the disk constant, I can see that I now have control over the ephemeral disk (with this change, moved from 300mb to 1Gb) and I no longer get errors that I am now out ephemeral disk? I change it back, to 300mb and error appear.  

Maybe there is another explanation, but I don't see it immediately and the solution appeared to fix the problem

bradrydzewski:
I am certain this is how the implementation works. You can see in the below code that we schedule the `drone/controller` container and mount the host machine docker socket, which the controller uses to spawn additional containers outside of the nomad process. It is very likely to be another explanation.

```diff
+	volume := "/var/run/docker.sock:/var/run/docker.sock"
	if stage.OS == "windows" {
		volume = "////./pipe/docker_engine:////./pipe/docker_engine"
	}

	task := &api.Task{
		Name:      "stage",
		Driver:    "docker",
		Env:       env,
		Resources: &api.Resources{},
		Config: map[string]interface{}{
			"image":      internal.DefaultImage(s.config.DockerImage),
			"force_pull": s.config.DockerImagePull,
+			"volumes":    []string{volume},
		},
	}
```

[1] https://github.com/drone/drone/blob/master/scheduler/nomad/nomad.go

PR #2701, opened 2019-05-09
handle errors that raise in goroutine

lpf32:
terminate program when errors raise in goroutine


goerzh:
I run drone-server when DRONE_SERVER_PORT is 80 or other invalid arguments, the HTTP server is down and raises error, but the drone-server is still alive.  I fix this bug that stops drone-server when errors raise in goroutine.

marko-gacesa:
I'm sorry. We are not going to merge this PR. Thank you.

PR #2718, opened 2019-05-30
Use Proxy.Addr instead of Server.Addr for external links.

manuelluis:
Fix that DRONE_SERVER_PROXY_HOST is not used for external links in gitea, github etc..


bradrydzewski:
the code is written as intended. The proxy address is only used for webhooks. The intended use case if when your github server (for example) sits outside the firewall and you need webhooks delivered to an alternate address.  Login redirects, on the other hand, happen within the browser on your host machine (e.g. your laptop) and are therefore not the target use case for this field.

PR #2721, opened 2019-06-05
Remove basic auth as an option for Gitea

### LINKED ISSUE #2704: Gitea token API change breaking Gitea-Drone authentication ("Login Failed. Forbidden")
I have been experiencing this issue: https://github.com/go-gitea/gitea/issues/6883 which I believe is caused by a recent Gitea commit: ( https://github.com/go-gitea/gitea/commit/46373e765702a203bfcb576c9d3639314ef4965f#diff-e8b760d6f593a855db95ca635708d685L9470 ) which has changed the create-token API response.

Specifically, the key called 'sha1' (containing the token) has seemingly changed name to 'token'.

This means that the following code is no longer correct:

https://github.com/drone/go-login/blob/0e45116bc8185ecedcc4e6d6f7e70b4a12b75623/login/gogs/handler.go#L43

However, this also represents a breaking change between Gogs and Gitea authentication - thus it would be necessary to separate out a Gitea-specific routine in go-login (which currently uses the same code for Gogs and Gitea).

Should Gitea instead be asked to revert this change to restore compatibility?

techknowlogick:
I suspect this may increase support communication. If this is the case feel free to direct users to the gitea discord where we can help them switch to OAuth2.

Fixes #2704


techknowlogick:
@bradrydzewski now that Gitea OAuth support has been in Drone since 1.1.0 would it be possible to review this PR for 1.4.0?

bradrydzewski:
as soon as I have the docs updated to include oauth (which should be soon) I will merge this right away. sorry for the delay!

techknowlogick:
no worries, I completely understand. I know docs aren't public yet, but lmk if I can help with them.

bradrydzewski:
thanks and sorry for the delay.  the docs have also been updated accordingly.

PR #2723, opened 2019-06-06
update go-scm to v1.5.0

techknowlogick:


bradrydzewski:
oops, looks like I beat you to this in a separate commit :)
sorry about that

PR #2732, opened 2019-06-15
Add status badge for individual stages

PatWie:
This shows for builds consisting of multiple pipelines the status
badge for a requested stage.


PatWie:
To give some context on this change. I frequently wished that the badge shows the status for individual pipelines (different versions) as in

- https://github.com/PatWie/cuda-design-patterns
- https://github.com/PatWie/tensorflow-cmake

If one stage fails all are displayed as a failure. Hope this PR fixes this. 
Disclaimer: I blindly applied the changes and added the tests.

PatWie:
Added unit-tests. I have no idea where the error for `linux-arm64` comes from. 

Given the fact, that I just removed an unnecessary test compared to the run in https://cloud.drone.io/drone/drone/168/3/2 this is strange: If a subset of tests does not pass while the superset with one more test passes.

tu-nv:
+1. Any update on this PR?

PatWie:
As there was no feedback, I haven't continued to work on this. I would rebase on HEAD, when I would update drone on my machine.

mattmattox:
Is this still being worked on? If not, I would happy to take over this feature.

PatWie:
I am using woodpecker-ci. There won't be an update from my side.

alikhil:
Hi, @bradrydzewski ! What do you think about this feature?

bot2-harness:

PR #2735, opened 2019-06-18
Add support for v1.PodSpec.ImagePullSecrets on Kubernetes native Drone server

dmlambea:
When using a private registry with authentication in Kubernetes, jobs get stuck in "ImagePullBackoff" state. This PR adds support for ImagePullSecrets when creating the Kubernetes Job, so the job can start.

To use the feature, set the env. var. DRONE_KUBERNETES_IMAGE_PULL_SECRETS to the name of the k8s secret for the private registry when deploying the Kubernetes native Drone server.

### DIFF
--- a/cmd/drone-server/config/config.go
+++ b/cmd/drone-server/config/config.go
@@ -124,14 +124,15 @@ type (
 
 	// Kubernetes provides kubernetes configuration
 	Kubernetes struct {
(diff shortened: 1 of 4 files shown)


bradrydzewski:
the kubernetes runtime in this repository was deprecated.  the second generation implementation can be found at https://github.com/drone-runners/drone-runner-kube

Issue #2685, opened 2019-04-24
Include Action in when clause


bradrydzewski:
We should start tracking the action associated with a webhook, for example pull request opened/closed/syncronized and branch created/updated. We already store the action field in the database, so the changes would be primarily to:

1. ~~update the go-scm libraries to ensure actions are being set~~
2. ~~update the yaml when section to support actions~~
3. ~~update the yaml trigger section to support actions~~
4. update the /trigger package to evaluate actions
5. update the /runner package to evaluate actions

We should also consider aliasing `push` events as `branch` events to use in conjunction with `created` and `updated` events.

Example update to `when` clause:

```
when:
  event: [ pull_request ]
  action: [ opened ]
```

Example update to `trigger` clause:

```
trigger:
  event: [ pull_request ]
  action: [ opened ]
```

aasm-moura:
Any workaround til we get this feature?

lucianjon:
@bradrydzewski This would be pretty useful for us, any concerns if I were to pick up this work?

bradrydzewski:
@lucianjon yes, I had originally hoped to ship this in 1.2 but I ran out of time. Some of the todos in the list were completed. I will update the issue to reflect which are completed this week so we can come up with a plan.

bradrydzewski:
@lucianjon master now has the updated yaml structure which includes the `action` field in the when clause and in the `trigger` clause. In some cases the action clause is being populated, including pull request opened and synchronized actions.

Next steps are:

1. Update the trigger logic to support action
  https://github.com/drone/drone/blob/master/trigger/trigger.go#L255:L272
2. Update the when logic to support action (might require updated to drone/drone-yaml)
  https://github.com/drone/drone/blob/master/operator/runner/runner.go#L260:L270

lucianjon:
@bradrydzewski Thanks for that, I'll take a look when I have a chance this week

lucianjon:
@bradrydzewski I've created https://github.com/drone/drone-yaml/pull/48 and https://github.com/drone/drone/pull/2741

lucianjon:
I just need to do some more testing around the when conditionals

bradrydzewski:
@lucianjon I'm so sorry 😞we completely missed your pull request and implemented a subset of the functionality. Your pull request was actually much more complete (and better) so I was wondering if you would be willing to rebased and merge. If not I completely understand.

bradrydzewski:
[...]

PR #2741, opened 2019-06-25
Add support for Action conditions in pull request events

lucianjon:
Done as a part of https://github.com/drone/drone/issues/2685. Requires the work done in https://github.com/drone/drone-yaml/pull/48

In the process I found the `trigger/trigger_test.go` tests we're panicing. It seemed to be coming from the yaml unmarshaller not understanding the test data and generating nil value pipelines. Change it to something that worked but would be curious if there's a different way to fix it as I didn't fully understand all the unmarshalling logic.

Adds the rest of the work needed to process a `pull_request` event's `action`. Useful if you only want to trigger a pipeline on the opening of a pull request. 

Has been tested locally with a personal github repo and this drone configuration:

```
---
kind: pipeline
name: "only on open"

steps:
- name: echo
  image: golang
  commands:
  - echo "this only triggers on open"

trigger:
  event:
  - pull_request
  action:
  - opened
---
kind: pipeline
name: "all pr events"

steps:
- name: echo
  image: golang
  commands:
  - echo "this should always trigger"

trigger:
  event:
  - pull_request
```

### DIFF
--- a/.github/security.md
+++ b/.github/security.md
@@ -16,8 +16,8 @@ your report. After the initial reply to your report, the security team will
 endeavor to keep you informed of the progress towards a fix and full
 announcement, and may ask for additional information or guidance.
 
-Report security bugs in third-party packages to the person or team maintaining
-the module.
+Report security bugs in third-party software to the person or team maintaining
+that software.
 
 ## Disclosure Policy
 
@@ -28,7 +28,7 @@ involving the following steps:
   * Confirm the problem and determine the affected versions.
   * Audit code to find any potential similar problems.
   * Prepare fixes for all releases still under maintenance. These fixes will be
-    released as fast as possible to npm.
+    released as fast as possible to DockerHub.
 
 ## Comments on this Policy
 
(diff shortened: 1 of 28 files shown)


bradrydzewski:
Looks really solid. To hook everything up to the when clause you will also want to pass the Action to the filter func args here https://github.com/drone/drone/blob/master/operator/runner/runner.go#L261

lucianjon:
Cheers, yeah was just working on that too. It should be working now, have tested it with this pipeline: 

```
kind: pipeline
name: test
steps:
- name: echo only
  image: golang
  when:
    event:
    - pull_request
    action:
    - opened
  commands:
  - echo "this only triggers on open"
- name: all pr events
  image: golang
  when:
    event:
    - pull_request
  commands:
  - echo "this triggers on all"
```

lucianjon:
@bradrydzewski I think the last thing to do to get the build passing is releasing a new drone-yaml to get access to the `Action` field in `SkipData`

PR #2749, opened 2019-07-04
config scheduler's host aliases via ENV

bgdsh:
Hi,
this PR is for demo the feature I want in https://discourse.drone.io/t/if-its-possible-to-add-hostaliases-on-the-job-pod/5042 
if it is not qualified to merge, please help to implement it yourself or give me some advice, then I can do the relevant changes.

### DIFF
--- a/cmd/drone-server/config/config.go
+++ b/cmd/drone-server/config/config.go
@@ -132,6 +132,7 @@ type (
 		ServiceAccountName string `envconfig:"DRONE_KUBERNETES_SERVICE_ACCOUNT"`
 		PullPolicy         string `envconfig:"DRONE_KUBERNETES_IMAGE_PULL" default:"Always"`
 		Image              string `envconfig:"DRONE_KUBERNETES_IMAGE"`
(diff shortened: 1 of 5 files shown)


bgdsh:
This feature is not only useful for me, it is also useful for the ones who are using github enterprise version.

rocketbitz:
This is also extremely useful for me. Anything holding up this getting merged in besides the `go.sum` conflict? Willing to contribute if needed.

bradrydzewski:
the first, experimental iteration of the kubernetes runner was deprecated back in april. The second experimental iteration now lives here:
https://github.com/drone-runners/drone-runner-kube

also the kubernetes runner was re-architected and it is likely this change is no longer required, since the new implementation no longer uses jobs.

rocketbitz:
Thanks for the clarification @bradrydzewski. Does that mean I can change `DRONE_KUBERNETES_IMAGE=drone/drone-runner-kube:latest` if I want to run the new one, or are there other steps? Thanks in advance.

bradrydzewski:
it is a completely separate installation and has its own dedicated yaml syntax
https://kube-runner.docs.drone.io

PR #2751, opened 2019-07-08
Fix gitignore for Intellij GoLand import

tiggerlee2:


bradrydzewski:
You can create a global .gitignore file that is specific to your environment. See https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

PR #2754, opened 2019-07-09
Promote a set of labels to all Kubernetes-native Drone Controller jobs on creation

dmlambea:
This PR adds a preconfigured set of labels on every drone-controller job at creation time. By allowing the jobs to be labelled, Kubernetes' PodPreset objects can inject additional configurations such as global environment vars and mount points.

My use case is to inject kubeconfig files onto drone-controllers' jobs, so the controller can authenticate against the apiserver by using client certificates.


bradrydzewski:
the kubernetes runtime in this repository was deprecated.  the second generation implementation can be found at https://github.com/drone-runners/drone-runner-kube

PR #2772, opened 2019-08-03
Add id column in migrations table for MySQL Group Replication

rluisr:
Need PK each table on MySQL Group Replication.
```
2019-08-03T19:22:34.446530Z 80836 [ERROR] [MY-011542] [Repl] Plugin group_replication reported: 'Table migrations does not have any PRIMARY KEY. This is not compatible with Group Replication.'
```

Run drone and add id primary key, It’s worked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@
 .env.*
 release/
 scripts/
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 .env.*
 release/
 scripts/
+.idea/
--- a/go.sum
+++ b/go.sum
@@ -176,6 +176,7 @@ github.com/hashicorp/go-multierror v1.0.0 h1:iVjPR7a6H0tWELX5NxNe7bYopibicUzc7uP
 github.com/hashicorp/go-multierror v1.0.0/go.mod h1:dHtQlpGsu+cZNNAkkCN/P3hoUDHhCYQXV3UM06sGGrk=
 github.com/hashicorp/go-retryablehttp v0.0.0-20180718195005-e651d75abec6 h1:qCv4319q2q7XKn0MQbi8p37hsJ+9Xo8e6yojA73JVxk=
 github.com/hashicorp/go-retryablehttp v0.0.0-20180718195005-e651d75abec6/go.mod h1:fXcdFsQoipQa7mwORhKad5jmDCeSy/RCGzWA08PO0lM=
+github.com/hashicorp/go-rootcerts v1.0.0/go.mod h1:K6zTfqpRlCUIjkwsN4Z+hiSfzSTQa6eBIzfwKfwNnHU=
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@
 .env.*
 release/
 scripts/
+.idea/
```
tboerger: You should ignore IDE specific files and folders in your own global gitignore.

rluisr: ok, fixed.

store/shared/migrate/mysql/ddl_gen.go
```diff
@@ -186,6 +186,7 @@ func selectCompleted(db *sql.DB) (map[string]struct{}, error) {
```
bradrydzewski: this file is automatically generated and should not be manually updated. Otherwise the next time we generate this file your changes will be lost. You need to send a pull request to update the generator, and then you need to regenerate this file using `go generate`

store/shared/migrate/mysql/ddl_gen.go
```diff
@@ -186,6 +186,7 @@ func selectCompleted(db *sql.DB) (map[string]struct{}, error) {
 
 var migrationTableCreate = `
 CREATE TABLE IF NOT EXISTS migrations (
+ id INTEGER PRIMARY KEY AUTO_INCREMENT,
  name VARCHAR(255)
```
bradrydzewski: If a primary key is required could we just remove the unique constraint from the name column, and use a primary key constraint instead?

```diff
CREATE TABLE IF NOT EXISTS migrations (
 name VARCHAR(255)
-,UNIQUE(name)
+,PRIMARY KEY(name)
)
```

go.sum
```diff
@@ -176,6 +176,7 @@ github.com/hashicorp/go-multierror v1.0.0 h1:iVjPR7a6H0tWELX5NxNe7bYopibicUzc7uP
 github.com/hashicorp/go-multierror v1.0.0/go.mod h1:dHtQlpGsu+cZNNAkkCN/P3hoUDHhCYQXV3UM06sGGrk=
```
bradrydzewski: please exclude this file from the pull request

store/shared/migrate/mysql/ddl_gen.go
rluisr: yes. I think no need `id` if `name` is unique.

store/shared/migrate/mysql/ddl_gen.go
rluisr: sorry I don't know the file where should I modify.

rluisr: I fond.
https://github.com/drone/drone-migrate/blob/master/migrate/db/mysql/ddl_gen.go

rluisr:
I recreate PR for excluding go.sum when this [PR](https://github.com/drone/drone-migrate/pull/17) is merged.

PR #2792, opened 2019-08-16
Add support for primary store of GCS

Lucretius:
This adds support for using Google Cloud Storage as the main logging store.  For credentials, GCP uses an environment variable `GOOGLE_APPLICATION_CREDENTIALS` which points to a JSON file on disk.  These service account credentials are created via the Google Cloud portal.  It will need to either be volume-mounted to the container or inside the container itself.  The only thing that actually needs to be passed in is the `DRONE_GCS_BUCKET` env var.  I was able to successfully test this out running locally via ngrok.

I checked a few times that the go modules were minimal, the `cloud.google.com/go` package seems to also update several of the other modules - resulting in a noisier `go.sum` than for the other PR's `Azure` provider.  I also tried just running `go install <path-to-drone-server>` and seem to add the same deps.  If there's another way I should be performing module changes (VS Code's Go plugin is doing this all for me), let me know and I can try another approach.

This will definitely have conflicts in the `inject_store` and `config` files with my other PR for Azure.


tboerger:
You should use go fmt to format the code properly

Lucretius:
> You should use go fmt to format the code properly

Yeah, I used the Github visual editor last night to fix the conflicts between master and my branch - I'll pull it down and give it a pass to fix the formatting issues.

tboerger:
The latter makes most sense as the initialize doesn't return errors

store/logs/gcs.go
```diff
@@ -0,0 +1,86 @@
...
+	bucket, err := g.getBucket(ctx)
+	if err != nil {
+		return err
+	}
+	obj := bucket.Object(fmt.Sprintf("%d", step))
+	return obj.Delete(ctx)
+}
+
+func (g *gcsStore) getBucket(ctx context.Context) (*storage.BucketHandle, error) {
+	if g.bucket != nil {
```
tboerger: You can drop that, getBucket is anyway called once.

store/logs/gcs.go
```diff
@@ -0,0 +1,86 @@
...
+		return g.bucket, nil
+	}
+	client, err := storage.NewClient(ctx)
+
+	if err != nil {
+		return nil, err
+	}
+	g.bucket = client.Bucket(g.bucketName)
```
tboerger: Just return that and drop the bucket attribute

Lucretius:
The reason I opted to keep the bucket pointer attribute was so that we would only ever generate one storage client and one bucket handle per log store.  If we drop the pointer, every time we retrieve the bucket we end up throwing away and recreating the same bucket handle and client over and over again.  It seemed more efficient to keep that all in the data structure, instantiate it once and just reference the pointer whenever we needed.  Maybe I am misunderstanding - is there a reason you'd prefer the other way?

bradrydzewski:
I just realized that Google Cloud Storage supports the S3 protocol [1].  Perhaps instead we re-use the S3 implementation but use Google-specific environment variables instead of S3 variables?

```
    client = boto3.client("s3", region_name="auto",
                          endpoint_url="https://storage.googleapis.com",
                          aws_access_key_id=google_access_key_id,
                          aws_secret_access_key=google_access_key_secret)
```

[1] https://cloud.google.com/storage/docs/migrating#migration-simple

tboerger:
You should take https://github.com/Lucretius/drone/blob/master/store/logs/s3.go as reference, there it's also always initialized.

bradrydzewski:
my previous comment aside, if you are going to use a global bucket reference it should probably be created in `NewGCSEnv`.  The current implementation is not threadsafe and should otherwise use a mutex or sync.Once. But I think in the `NewGCSEnv` is probably best.

Lucretius:
@bradrydzewski That's actually a really good idea - and requires no additional changes.  I tried it out with an HMAC key on the existing master branch and it just worked.  Drone is able to read/write to the storage account.

Just need to set the same env vars that are required for AWS with the changes you mentioned and it works as-is.  I guess the only work to be done is documentation in that case.  If you'd like to go down that route, I can close this PR.

PR #2801, opened 2019-08-24
Fix typo

crossworth:


crossworth:
Yes, https://github.com/drone/drone/search?q=LicensePeronal&unscoped_q=LicensePeronal

Goland as well says that there is no other references for this const.

tboerger:
Ok

PR #2809, opened 2019-08-30
Trusted environment switch

vyckou:


bradrydzewski:
thanks!  I will take a look and provide feedback.

bradrydzewski:
First of all, thanks for the pull request.  I have been thinking of how we might want to implement this, and I am thinking we may want to change this at the runner level.  We could set `DRONE_RUNNER_TRUST_ALWAYS` (for example) and the runner would always run the linter under the assumption that the repository is trusted.

vyckou:
OK will change to this approach.

But wouldn't it be a bit misleading? As in database you would have trusted / untrusted settings and then simply would ignore that setting on runner(s)

Also just popped the idea, that this type of flag would actually make the default setting of newly enabled repo as trusted and still would allow to set the repo(s) as untrusted

Issue #2793, opened 2019-08-17
Drone missing nsswitch.conf


matt-simons:
When pulling drone:1.1.0 there seems to be a regression from when this was last fixed in #1306

Drone no longer seems to respect /etc/hosts.

bradrydzewski:
I cannot remember the exact reasons, but the the nsswitch was removed back in 2016

tboerger:
Drone is based on alpine, if I remember correctly Müsli libc didn't rely on nsswitch.conf

matt-simons:
If I understand [this](https://giedrius.blog/2018/11/22/etc-nsswitch-conf-and-etc-hosts-woes-with-the-alpine-and-others-docker-image-and-golang/) correctly, it is golang's net package which assumes there will always be an nsswitch, which isn't always the case.

Which I suppose leaves 3 options:

1. Add an nsswitch anyway so that drone will read /etc/hosts (though as @bradrydzewski suggests this may cause other issues)
2. Bake support for reading /etc/hosts into drone itself
3. Status quo and require the end user to mount an nsswitch.conf file at runtime

It would be ideal if the drone images supported Docker linking and Kubernetes hostAlias out of the box.

jstxu:
I can confirm that running drone with the flag `-v /etc/nsswitch.conf:/etc/nsswitch.conf` makes it work. 

So I've made a PR: https://github.com/drone/drone/pull/2820

jstxu:
I wanted to add the `nsswitch.conf` because I needed to setup the gitea service ip on the `/etc/hosts`. 

If this is your case, just using the steps above won't be enough, as even though the Drone server will access the correct IP, the pipeline slaves won't because those dockers won't have the `nsswitch.conf` either.

So the solution I've found is to use the host dns server for the containers. You can
specify the dns servers in the docker networks, but if you don't want to specify the `--network` flag on the drone slaves, you have to modify the `bridge0` network. To do so you have to add this configuration to the `/etc/docker/daemon.json`

```json
{
  "dns": ["172.17.0.1"]
}
```

Usually systemd machines have a `systemd-resolv` service listening on
`127.0.0.53:53`, the problem is that you can't access that address from within
the container, so you would have to make systemd-resolv listen to the inet
address of the bridge0 network, or use another dns server that forwards the dns
queries from the docker network to the systemd one or the `/etc/hosts` file.

```bash
$ ifconfig docker0
[...]

PR #2820, opened 2019-09-10
Added nsswitch.conf to Linux Dockerfiles

jstxu:
Added `nsswitch.conf` to the Linux Dockerfiles to solve https://github.com/drone/drone/issues/2793.

It's really similar to an old PR, https://github.com/drone/drone/pull/1306

Cheers


bradrydzewski:
we decided to add the nsswitch (so thanks for the feedback) however we decided to use a slightly different approach based on what some other docker images are doing [1]

```
RUN [ ! -e /etc/nsswitch.conf ] && echo 'hosts: files dns' > /etc/nsswitch.conf
```

this is now fixed in master and will be part of the 1.6 release.

[1] https://github.com/docker-library/docker/pull/84/files#diff-1e3e470e2d59f5fd7cbb742d609f8b2dR10

PR #2826, opened 2019-09-11
feat(repo): allow internal project visibility by default

JordanSussman:
This partially solves the issue #2825. However, I ran into issues when testing this on an existing drone instance/database where existing repositories (unactivated) are already written to the database with the relevant visibility, and this code does **not** update the database in that scenario. Creating net-new repos in the scm or utilizing a new-net database makes the code work as expected.

Open to suggestions on resolving the database for existing instances.

### DIFF
--- a/cmd/drone-server/config/config.go
+++ b/cmd/drone-server/config/config.go
@@ -173,7 +173,8 @@ type (
 
 	// Repository provides the repository configuration.
 	Repository struct {
-		Filter []string `envconfig:"DRONE_REPOSITORY_FILTER"`
+		Filter     []string `envconfig:"DRONE_REPOSITORY_FILTER"`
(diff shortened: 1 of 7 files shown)


bradrydzewski:
> However, I ran into issues when testing this on an existing drone instance/database where existing repositories (unactivated) are already written to the database with the relevant visibility, and this code does not update the database in that scenario.

I think I actually prefer this behavior.  It is consistent with how some of our other configuration parameters work, namely the repository filter.  It also keeps things a bit more simple and we can provide a manual workaround if needed `update repos set repo_visibility = 'internal' where repo_private = false`

bradrydzewski:
this looks good to me, I have no additional feedback or requests

PR #2831, opened 2019-09-14
Add unique random stage token

frebib:
This random value is useful for referring to a uniquely identifiable constant across all steps in a pipeline/stage.

Currently the only way to replicate the same behaviour is to concatenate multiple `DRONE_` variables together to get a predictable value. Our use-case for this is identifying files/docker images across multiple stages in the same pipeline, allowing for concurrent runs on the same agent without naming conflicts/races

### DIFF
--- a/operator/runner/env.go
+++ b/operator/runner/env.go
@@ -15,6 +15,7 @@
 package runner
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"regexp"
 	"strings"
@@ -43,6 +44,8 @@ func agentEnviron(runner *Runner) map[string]string {
 }
 
 func repoEnviron(repo *core.Repository) map[string]string {
+	h := sha256.New()
+	fmt.Fprint(h, repo.ID, repo.UserID, repo.Slug)
 	return map[string]string{
 		"DRONE_REPO":            repo.Slug,
 		"DRONE_REPO_SCM":        repo.SCM,
@@ -56,6 +59,7 @@ func repoEnviron(repo *core.Repository) map[string]string {
 		"DRONE_GIT_SSH_URL":     repo.SSHURL,
 		"DRONE_REPO_VISIBILITY": repo.Visibility,
 		"DRONE_REPO_PRIVATE":    fmt.Sprint(repo.Private),
+		"DRONE_REPO_TOKEN":      fmt.Sprintf("%x", h.Sum(nil)),
 
 		//
 		// these are legacy configuration parameters for backward
@@ -71,6 +75,8 @@ func repoEnviron(repo *core.Repository) map[string]string {
 }
 
 func stageEnviron(stage *core.Stage) map[string]string {
+	h := sha256.New()
+	fmt.Fprint(h, stage.RepoID, stage.BuildID, stage.ID)
 	return map[string]string{
 		"DRONE_STAGE_KIND":       "pipeline",
 		"DRONE_STAGE_NAME":       stage.Name,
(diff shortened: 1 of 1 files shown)


bradrydzewski:
the `Build.ID` is a globally unique value for a single build that would be the same for all stages. This could be made available via a `BUILD_ID` environment variable.

frebib:
The point here is to have a value _different_ for each stage/pipeline in a build, but the same in each step in the pipeline- that way I can have two pipelines running in the same build concurrently, but having different id/token values so they don't try to create the same file/image/whatever on the host. 
The terminology here is particularly confusing, which doesn't help matters.

frebib:
I think perhaps that the scope of this PR is perhaps a little narrow. Would it be better if I implemented this functionality for each level of a build for consistency? Variables such as:
```
DRONE_BUILD_TOKEN
DRONE_STAGE_TOKEN
DRONE_STEP_TOKEN
```

bradrydzewski:
would this be the stage name?

frebib:
Yes and no; In my particular use-case, I'd need a value unpredictable enough to not cause a conflict. A good example of where Drone already does this is the network names, or container names. Using the `BUILD_ID` would work in some cases, but given that it is just a serial number, the chances of collision are very high. 

Specifically I need a value for naming stored docker images across multiple steps in a single pipeline- currently we only run one pipeline per agent, and the image names are fixed. By using a random value (per pipeline) I gain two things: implicit naming (I don't have to copy-paste a name across the .drone.yml) and being able to run the same build twice on the same docker instance/agent.

One example: (the instances of `test-dev` would be replaced implicitly by the variable)
https://github.com/spritsail/drone-docker-test/blob/master/.drone.yml

bradrydzewski:
I am not opposed to creating a new environment variable, but I am opposed to creating columns and storing a new value in the database unless absolutely necessary.  There should be a way to generate such a hash from existing values.

frebib:
Oh for sure, this value doesn't need to be stored. Instead would you prefer I generate this value as a hash of existing variables inside `operator/runner/runner.go`? That seems the next-best approach that doesn't extend the struct

bradrydzewski:
If we are using existing values to create a deterministic hash, we should only need to update operator/runner/env.go

frebib:
I changed up how this works now, please lmk what you think.

I ran a simple POC showing it doing the right thing:
* `.drone.yml`
https://git.spritsail.io/frebib/drone-test/src/branch/env/.drone.yml

* Two runs showing hash determinism
https://drone.nerdhouse.io/frebib/drone-test/32
https://drone.nerdhouse.io/frebib/drone-test/33

bradrydzewski:
can we limit this change to the `buildEnviron` function? all the environment variable code is moving to the runner-go library, which means everything in this pull request needs to be backed out. Keep this code isolated is therefore preferred.

```
func buildEnviron(build *core.Build) map[string]string {
...

	h := sha256.New()
	fmt.Fprintln(h, build.ID)
	env["DRONE_BUILD_HASH"] = fmt.Sprintf("%x", h.Sum(nil))

...
```

also I feel like the hash of the `build.ID` should be just as unique as the hash of all the other combined fields, unless you are running multiple Drone instances on the same machines (which it sounds like you are not).

frebib:
I have left in several values to ensure that the hash values don't collide with each other `BUILD_TOKEN` == `STAGE_TOKEN` (`1 == 1`), I hope you think that's ok.

I also think perhaps there should be a hash for the `Stage` too for completeness, but I'm not entirely sure it's particularly useful

PR #2834, opened 2019-09-16
Fix Dropped operator Errors

alrs:
This PR picks off two dropped errors, one in `operator/manager` and the other in `operator/runner`.

### DIFF
--- a/operator/runner/runner.go
+++ b/operator/runner/runner.go
@@ -204,7 +204,7 @@ func (r *Runner) Run(ctx context.Context, id int64) error {
 		return env
 	})
 	if err != nil {
-		return err
+		return r.handleError(ctx, m.Stage, err)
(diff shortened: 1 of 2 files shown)


operator/runner/runner.go
```diff
@@ -203,6 +203,9 @@ func (r *Runner) Run(ctx context.Context, id int64) error {
 		}
 		return env
 	})
+	if err != nil {
+		return err
```
bradrydzewski: for this particular error we should do the following to ensure it is reported back to the server and visible int he user interface:

```
return r.handleError(ctx, m.Stage, err)
```

alrs: Done, @bradrydzewski . Do you guys prefer that I squash commits, or leave this as-is?

PR #2839, opened 2019-09-18
Add plugins/acr to PrivilegedMode

domgoer:


domgoer:
Added [drone-cli#153](https://github.com/drone/drone-cli/pull/153)

tboerger:
Thanks

PR #2863, opened 2019-10-25
Add DRONE_COMMIT_TIMESTAMP to runner environment

ivanvc:
Hi,

Our code coverage tool (Codeclimate) requires to access the commit timestamp, which is not accessible right now from the Runner environment. This is a tiny addition just to expose it as an environment variable to the runners.

P.S. I couldn't find a contributing guide, so I just made sure to not break the format, and that all unit tests are passing locally, by running `go test ...`, but didn't add one as these variables don't have unit tests. I just made sure that `build.Timestamp` existed and was accessible.

### DIFF
--- a/operator/runner/env.go
+++ b/operator/runner/env.go
@@ -100,7 +100,7 @@ func buildEnviron(build *core.Build) map[string]string {
 		"DRONE_COMMIT_AUTHOR_EMAIL":  build.AuthorEmail,
 		"DRONE_COMMIT_AUTHOR_AVATAR": build.AuthorAvatar,
 		"DRONE_COMMIT_AUTHOR_NAME":   build.AuthorName,
-		"DRONE_COMMIT_TIMESTAMP":     build.Timestamp,
+		"DRONE_COMMIT_TIMESTAMP":     fmt.Sprint(build.Timestamp),
 		"DRONE_BUILD_NUMBER":         fmt.Sprint(build.Number),
 		"DRONE_BUILD_EVENT":          build.Event,
 		"DRONE_BUILD_ACTION":         build.Action,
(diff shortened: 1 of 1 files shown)


bradrydzewski:
thanks for sending a pull request !  note that the timestamp of the commit is not captured at the moment and is a zero value, which needs to be resolved in github.com/drone/go-scm

ivanvc:
:open_mouth: oh ok. Is that a limitation, or an open issue? Happy to work on that too.

bradrydzewski:
thanks, I went ahead and updated the push webhook to show what is needed:
https://github.com/drone/go-scm/commit/08514fd8b9d133854197976f8f5b4a7eb34f20a8#diff-7b308f8bd592c43997207fe05c964c88L175

if you look at the following files:
```
scm/driver/github/testdata/webhooks/*.golden
```

and look for empty dates:

```
0001-01-01T00:00:00Z
```

you can probably find examples of instances where we came across a data string but did handle the value. I recall there being some concerns around nil values, but my memory is hazy.

ivanvc:
After reviewing the golden files, and the actual webhooks, it looks like the timestamp of the commit is only available for the webhook events: push branch, push and push tag. In any other case this value is not provided.

I did look for empty dates, but they're either for deletion of refs, or the create and update timestamp for the repository, not part of the `head_commit`. For this reason, I guess the environment variable should be named in the `DRONE_COMMIT` "namespace". But maybe needs to be documented in the documentation site (which don't know how it is built). 

Is there anything I can do to make this commit into master? Should I add it to the `CHANGELOG.md`?

ashwilliams1:
This codebase no longer is responsible for setting these variables. This functionality has moved to a shared library at https://github.com/drone/runner-go

Issue #1878, opened 2016-12-05
Support additional pull request actions (closed, merged, etc)


tonglil:
As discussed on Gitter.

> "Figure out all possible events drone might want to handle that it doesn't handle today"

# Use case
When a PR is opened, deploy the app somewhere, and when the pr is closed, tear it down.

# Related code

> for github the parse code is at https://github.com/drone/drone/blob/master/remote/github/parse.go#L77:L92

> right now we only have a pull_request event, so we would need a pull_request_closed event type, which would get added to https://github.com/drone/drone/blob/master/model/const.go#L3

- New database field for the flag
- Ability to toggle on in the UI

bradrydzewski:
I'm interested in other event types that we might want to add to ensure that we don't need to make any additional design changes or abstractions.

Here is a list of all Bitbucket pull request event types:

```
Created
Updated
Approved
Approval removed
Merged
Declined
Comment created
Comment updated
Comment deleted
```

Here are GitHub pull request event types:

```
assigned
unassigned
labeled
unlabeled
opened
edited
closed
reopened
```

Here are some other GitHub events that we don't handle today:

```
create (tag or branch created)
delete (tag or branch deleted)
release
```

It is also important to note that drone assumes it can clone a pull request (or tag) as part of the pipeline process. So drone probably can't handle a tag deleted event because it will have nothing to clone, and won't be able to fetch the .drone.yml for the tag sha. **NOTE** this might also be the case for pull requests being closed, so we will need to verify.

bradrydzewski:
there was a thread in the discourse forum:
http://discourse.drone.io/t/support-for-github-event-closed-on-issues/372/6

I will copy my comments here

bradrydzewski:
Copied from discourse thread.
http://discourse.drone.io/t/support-for-github-event-closed-on-issues/372/6

### Database Changes

First step will be to decide how we want to represent and store this data in our Go code and in our database. We will need to store the action type, as you pointed out above.

```diff
create table builds (
---
+build_action TEXT
---
);
```

I would suggest that action values for pull requests should be the following:

```text
enum {
  open,
  merged,
  declined,
  synchronized,
}
```

The GitHub `closed` type would be converted to either `merged` or `declined`. I think in this case it is better to adopt the Bitbucket semantics so that we can store this information in a single field, as opposed to the GitHub approach which uses two fields (`action string, merged bool`).

### Yaml Changes

[...]

PR #2884, opened 2019-11-22
Added close_pull_request event trigger, to support cleanups of PR deployments.

fhp:
I have added the `close_pull_request` event trigger to make it possible to cleanup a PR environment with the github hooks.

### DIFF
--- a/core/event.go
+++ b/core/event.go
@@ -16,11 +16,12 @@ package core
 
 // Hook event constants.
 const (
-	EventCron        = "cron"
-	EventCustom      = "custom"
-	EventPush        = "push"
-	EventPullRequest = "pull_request"
-	EventTag         = "tag"
-	EventPromote     = "promote"
-	EventRollback    = "rollback"
+	EventCron             = "cron"
+	EventCustom           = "custom"
+	EventPush             = "push"
+	EventPullRequest      = "pull_request"
+	EventClosePullRequest = "close_pull_request"
+	EventTag              = "tag"
+	EventPromote          = "promote"
+	EventRollback         = "rollback"
 )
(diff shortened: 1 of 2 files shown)


bradrydzewski:
Please see the _Avoid Breaking Changes_ section of [this comment](https://github.com/drone/drone/issues/1878#issuecomment-287598789).

core/event.go
```diff
@@ -16,11 +16,12 @@ package core
...
-	EventPush        = "push"
-	EventPullRequest = "pull_request"
-	EventTag         = "tag"
-	EventPromote     = "promote"
-	EventRollback    = "rollback"
+	EventCron             = "cron"
+	EventCustom           = "custom"
+	EventPush             = "push"
+	EventPullRequest      = "pull_request"
+	EventClosePullRequest = "close_pull_request"
```
bradrydzewski: Instead of creating a new event type, this value should be stored in the `Action` field.

amq:
Could be behind a config variable.

ashwilliams1:
There is a proposal here that seems perfectly fine to me:
https://github.com/drone/drone/issues/1878#issuecomment-287598789

The additional action types could be opt-in and only executed when the trigger section is configured accordingly. For example:
 
```
trigger:
  pull_request:
  - declined
  - closed
  - merged
```

tboerger:
@ashwilliams1 if you replace `pull_request` with `action` drone is already prepared :)

andrewhowdencom:
I had a go at implementing this. It's not tested, but looking for feedback:

https://github.com/drone/drone/pull/2902

ashwilliams1:
I am going to close this in favor of #2902 which is based on this pull request, but uses the existing `action` attribute instead of adding a new event type.

PR #2902, opened 2020-01-14
Implement pull request close events

andrewhowdencom:
There is currently a feature request associated with being able to run
some sort of pipeline at the termination of some git event; specifically
in this case a "closure" of a pull request.

This allows the consumers of drone to use drone to spin up ephemeral
environments for the purpose of testing, and then later have them torn
down automatically by Drone once the pull request has been closed.

This commit implements such a solution after reading the guidance on:

  #2884

The commits describe the design path in further detail.

### DIFF
--- a/service/hook/parser/parse.go
+++ b/service/hook/parser/parse.go
@@ -236,7 +236,6 @@ func (p *parser) Parse(req *http.Request, secretFunc func(string) string) (*core
 		}
 		return hook, repo, nil
 	case *scm.PullRequestHook:
-
 		// TODO(bradrydzewski) cleanup the pr close hook code.
 		if v.Action == scm.ActionClose {
 			return &core.Hook{
@@ -253,9 +252,13 @@ func (p *parser) Parse(req *http.Request, secretFunc func(string) string) (*core
 				}, nil
 		}
 
-		if v.Action != scm.ActionOpen && v.Action != scm.ActionSync {
+		switch v.Action {
+		case scm.ActionOpen, scm.ActionClose, scm.ActionSync:
+			// Valid, continue
+		default:
 			return nil, nil, nil
 		}
+
 		// Pull Requests are not supported for Bitbucket due
 		// to lack of refs (e.g. refs/pull-requests/42/from).
 		// Please contact Bitbucket Support if you would like to
@@ -264,6 +267,7 @@ func (p *parser) Parse(req *http.Request, secretFunc func(string) string) (*core
 		if p.client.Driver == scm.DriverBitbucket {
 			return nil, nil, nil
 		}
+
 		hook = &core.Hook{
 			Trigger:      core.TriggerHook, // core.TriggerHook,
 			Event:        core.EventPullRequest,
--- a/trigger/trigger.go
+++ b/trigger/trigger.go
@@ -28,8 +28,6 @@ import (
 	"github.com/drone/drone/core"
 	"github.com/drone/drone/trigger/dag"
 
-	"github.com/drone/go-scm/scm"
-
 	"github.com/sirupsen/logrus"
 )
 
@@ -296,24 +294,9 @@ func (t *triggerer) Trigger(ctx context.Context, repo *core.Repository, base *co
 		// TODO add target
 		// TODO add ref
 
-		// Insert Defaults
 		name := pipeline.Name
 		if name == "" {
 			name = "default"
-		}
-
-		// With the introduction of the "close" event in drone the default behaviour changes from "opt in to opens
-		// only" to "opt in to all".
-		//
-		// So as to preserve BC, this polyfills the previous design of "open" and "sync" events being included, but
-		// "exclude" events not. This allows users to "opt in" to the exclude behaviour.
-		if pipeline.Trigger.Event.Match(core.EventPullRequest) &&
-			len(pipeline.Trigger.Action.Include) == 0 {
-			pipeline.Trigger.Action.Include = append(
-				pipeline.Trigger.Action.Include,
-				scm.ActionOpen.String(),
-				scm.ActionSync.String(),
-			)
 		}
 
 		node := dag.Add(pipeline.Name, pipeline.DependsOn...)
--- a/trigger/trigger.go
+++ b/trigger/trigger.go
@@ -28,6 +28,8 @@ import (
 	"github.com/drone/drone/core"
 	"github.com/drone/drone/trigger/dag"
 
+	"github.com/drone/go-scm/scm"
+
 	"github.com/sirupsen/logrus"
 )
 
@@ -293,10 +295,27 @@ func (t *triggerer) Trigger(ctx context.Context, repo *core.Repository, base *co
 		// TODO add instance
 		// TODO add target
 		// TODO add ref
+
+		// Insert Defaults
 		name := pipeline.Name
 		if name == "" {
 			name = "default"
 		}
+
+		// With the introduction of the "close" event in drone the default behaviour changes from "opt in to opens
+		// only" to "opt in to all".
+		//
+		// So as to preserve BC, this polyfills the previous design of "open" and "sync" events being included, but
+		// "exclude" events not. This allows users to "opt in" to the exclude behaviour.
+		if pipeline.Trigger.Event.Match(core.EventPullRequest) &&
+			len(pipeline.Trigger.Action.Include) == 0 {
+			pipeline.Trigger.Action.Include = append(
+				pipeline.Trigger.Action.Include,
+				scm.ActionOpen.String(),
+				scm.ActionSync.String(),
+			)
+		}
+
 		node := dag.Add(pipeline.Name, pipeline.DependsOn...)
 		node.Skip = true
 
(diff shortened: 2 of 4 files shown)


andrewhowdencom:
This worked :D I've patched our drone server with it for now to see if it breaks things

service/hook/parser/parse.go
```diff
@@ -232,9 +232,17 @@ func (p *parser) Parse(req *http.Request, secretFunc func(string) string) (*core
 		}
 		return hook, repo, nil
 	case *scm.PullRequestHook:
-		if v.Action != scm.ActionOpen && v.Action != scm.ActionSync {
+		switch v.Action {
+		case scm.ActionOpen:
```
bradrydzewski: you can shorten with `case scm.ActionOpen, scm.ActionClose, scm.ActionSync:`

andrewhowdencom: Done; seems to compile. Will test it .. shortlyish.

trigger/trigger.go
```diff
@@ -278,10 +280,27 @@ func (t *triggerer) Trigger(ctx context.Context, repo *core.Repository, base *co
...
 			name = "default"
 		}
+
+		// With the introduction of the "close" event in drone the default behaviour changes from "opt in to opens
+		// only" to "opt in to all".
+		//
+		// So as to preserve BC, this polyfills the previous design of "open" and "sync" events being included, but
+		// "exclude" events not. This allows users to "opt in" to the exclude behaviour.
+		if pipeline.Trigger.Event.Match(core.EventPullRequest) &&
+			len(pipeline.Trigger.Action.Include) == 0 {
```
bradrydzewski: I would prefer we introduce a new skip function and add to the if / else block below, and then add unit tests to ensure we have coverage for all scenarios.

andrewhowdencom: I'm not quite sure what you mean. 

If you mean like `ifSkipPullRequest` or? My thinking is that's getting a little bespoke; following this conceptual model forward, there might be defaults that are changed as actions continue to be built on future.

Perhaps an alternative abstraction might be some functions that match a pipeline and fill defaults if there are otherwise none before they get passed to the matcher? Like, if you had a map of functions:

```go
// backfills/api.go
// Note: I am not familiar with this codebase; these examples may not match variable names or types directly
type DefaultBackfill func(pipline *Pipeline) 

var Backfills := []DefaultBackfill{PullRequestActions} // Can add backfills here in future
```

With a backfill impl like:

```go
func PullRequestActions(pipeline *Pipeline) {
    if ! pipeline.Trigger.Event.Match(core.EventPullRequest) {
      return
    }

    // Backfill logic here. 
}
```

That would be extensible and easily unit testable. 

Alternatively, can you explain what you mean by "new skip function"?

bradrydzewski: > If you mean like ifSkipPullRequest or? My thinking is that's getting a little bespoke; following this conceptual model forward, there might be defaults that are changed as actions continue to be built on future.

yes, I mean adding an `ifSkipEventAction` which would skip pull request close events if the trigger does not explicitly enable this event and action.

bradrydzewski:
```
  trigger:
    event:
      include:
        - pull_request
    action:
      include:
        - closed
```

> Which worked with
> - Creating the PR
> - Updating the PR
> - Closing the PR
> - Merging the PR

I would expect the above yaml to only work with pull request closed actions, and to ignore sync and create actions.  If you want to trigger on create, sync and closed events you need to enumerate all action types.

andrewhowdencom:
> I would expect the above yaml to only work with pull request closed actions, and to ignore sync and create actions. If you want to trigger on create, sync and closed events you need to enumerate all action types.

Sorry yes, I should have made that clear -- that's how it works presently. Rather, the initial comment:

> Creating the PR
> Updating the PR
> Closing the PR
> Merging the PR

Meant that those were the behaviors that I tested across both pipelines in that yaml.

trigger/trigger.go
andrewhowdencom: > The way this is currently implemented, in this pull request, is undesirable because it mutates the pipeline by modifying the trigger section. This is the sort of thing that comes back to bite me years later when I'm making changes and not expecting this to be mutated.

That makes sense. Okay, I'll reimplement it in this way. To set expectations up front, I don't know when I'll get this -- hopefully within a couple weeks, honestly i don't know.

andrewhowdencom:
There's a regression in this patch -- it somehow blocks pipelines that only include:

```
trigger:
  ref:
    include:
      - refs/tags/**
```

I do not know why yet; I dropped the commit off our server.

andrewhowdencom:
Testing criteria:

```
---
# Expected: Only trigger on tags
# OK (push to tag): 61
kind: "pipeline"
type: "kubernetes"
name: "tag-but-an-event"
steps:
  - name: "container"
    image: "busybox"
    commands:
      - echo "Pushed"        
trigger:
  event:
    - tag
---
# Expected: Only trigger on tags
# OK (push to tag): 61
kind: "pipeline"
type: "kubernetes"
name: "tag"
steps:
  - name: "container"
    image: "busybox"
    commands:
      - echo "Pushed"        
trigger:
  ref:  
    - refs/tags/**
---
# Expected: Only trigger on push to i-am-a-pull-request
# - OK (push to branch): 58
kind: "pipeline"
type: "kubernetes"
name: "i-am-a-pull-request"
steps:
  - name: "container"
    image: "busybox"
    commands:
      - echo "Pushed"        
trigger:
  ref:
    - refs/heads/i-am-a-pull-request
  event:
    - push
---
# Expected: Trigger when a PR is opened, closed and synchronized
# - OK (PR Creation): 63
# - OK (PR Sync): 65
# - OK (PR Close): 68
kind: "pipeline"
type: "kubernetes"
name: "pull-request-explicit"
steps:
  - name: "container"
    image: "busybox"
    commands:
      - echo "Pushed"        
trigger:
  event:
    - pull_request
  action:
    - closed
    - opened
    - synchronized
---
# Expected: Trigger when a PR is synchronized and closed.
# OK (PR Sync): 65
# OK (PR Closed): 68
kind: "pipeline"
type: "kubernetes"
name: "exclude-open"
steps:
  - name: "container"
    image: "busybox"
    commands:
      - echo "Pushed"        
trigger:
  event:
    - pull_request
  action:
    exclude:
      - opened
---
# Expected: Trigger when a PR is opened and synchronized
# - OK (PR Creation): 63
# - OK (PR Creation): 65
kind: "pipeline"
type: "kubernetes"
name: "no-includes"
steps:
  - name: "container"
    image: "busybox"
    commands:
      - echo "Pushed"        
trigger:
  event:
    include:
      - pull_request
---
# Expected: Trigger when a PR is closed only.
# - OK (PR Closed)
kind: "pipeline"
type: "kubernetes"
name: "closed"
clone:
  disable: true
steps:
  - name: "container"
    image: "busybox"
    commands:
      - echo "Closed"        
trigger:
  event:
    include:
      - pull_request
  action:
    include:
      - closed
```

andrewhowdencom:
Pull request is also now up to date with revised solution (See a6effe3). Will leave patched on our server for now to see how it goes, but beyond that it *should* work fine.

bradrydzewski:
thanks for providing the examples.  the logic needs to be slightly adjusted to avoid breaking the current excludes behavior. Basically if the action is closed, the pipeline should only execute when the following evaluates to true:

```
if len(when.actions.includes) > 0 && when.actions.Includes("closed")
```

finally, we should add some unit tests to confirm the above logic:
https://github.com/drone/drone/blob/master/trigger/skip_test.go

andrewhowdencom:
Re. Unit tests I agree but also have no time at the minute; 

Re. Logic I'm not sure I 100% understand. Do you mean, given a pipeline that has:

```
trigger:
  event:
    - pull_request
  action:
    exclude:
      - opened
```

It should *not* listen for the "closed" event? That is, practically it would mean:

```
trigger:
  event:
    - pull_request
  action:
    exclude:
      - opened
      - closed
```

bradrydzewski:
correct, the below example should not create a pipeline for a pipeline closed event. There may be teams already using the below syntax (for synchronization actions) and they would not be expecting to start processing pull request closed events. So we need to avoid this breaking change.

```
trigger:
  event:
    - pull_request
  action:
    exclude:
      - opened
```

andrewhowdencom:
Sounds good! That makes sense. I will tackle that as soon as I can.

andrewhowdencom:
I had another go, but haven't tested it by actually releasing it yet. I wrote some tests for it based on the yaml I had before. 

For now, I need sleep :sleeping: but will check this in a bit.

andrewhowdencom:
status update: This sat on the build server for a week and didn't break anything!

natemcdowel:
This would be a very helpful feature to have!

service/hook/parser/parse.go
```diff
@@ -232,9 +232,13 @@ func (p *parser) Parse(req *http.Request, secretFunc func(string) string) (*core
 		}
 		return hook, repo, nil
 	case *scm.PullRequestHook:
-		if v.Action != scm.ActionOpen && v.Action != scm.ActionSync {
+		switch v.Action {
+		case scm.ActionOpen, scm.ActionClose, scm.ActionSync:
```
jakehamilton, line 236: Should this also include `scm.ActionMerge` to capture PR merge events?

jakehamilton: I'm not the most knowledgeable when it comes to Go and this codebase is new for me, but I've been digging through the webhook system. Here is the piece that I'm wondering about: https://github.com/drone/go-scm/blob/master/scm/driver/bitbucket/webhook.go#L49

amq:
We desperately need this.

andrewhowdencom:
The conflicts on this are non-trivial iirc, and I have limited time to resolve them. When we next upgrade our build server we'd look at it, but there's no massive reason to do that at the minute so it might be a while.

If someone wants to carry this, please feel free ♥

andrewhowdencom:
Revised; not really tested yet though. Someone should consider carrying this. For now, I deployed it on an internal CI server.

PR #2904, opened 2020-01-17
fixes #2903 accumulated delay and cron jobs alignement

### LINKED ISSUE #2903: Detected accuracy problems on cron jobs with "@every XX" with XX period time
Hi to all drone followers.

First I would like to be grateful with drone owners and collaborators for this great software.

As previously discused in https://discourse.drone.io/t/triggered-builds-by-cron-have-high-deviation-against-expected-time/6612/4 , we have detected some accuracy problems with builds triggered from cron with expressions `@every XX`  ( continuous period without no important where  it exact begins) 

The 2 main discussion issues was.

## Accumulated delay

 on jobs needed to have a stable cadence (with acceptance of some jitter) it could be a problem as there is lack of data.

**With current drone version**

image

En the previous image cron jobs should be triggered each 10min but all cron-jobs has been accumulating time until bypass one period 

**Expected**

image

This is the expected behaviour, where there is one execution on each time bucket

##  job alignment

Right now drone changes the triggering time depending on the DRONE_CRON_INTERVAL if drone-server (for maintenance or for incidence) has been stopped more than one "XX period"  on restart all cron jobs will be triggered first at the same time.  This fact could have impact on not much efficient resource consumption distributión and could overload runner system. 

Could be interesting if Drone could maintain the cron deploy initial time by the end user ( this is like delegate resource load distribucion on the end user ) , instead of changing it with this dangerous behaviour

toni-moreno:
Hi With this PR we can fix the behaviour described in #2903.

**With current drone version**

image

**With this patch**

image

### DIFF
--- a/handler/api/repos/crons/create_test.go
+++ b/handler/api/repos/crons/create_test.go
@@ -56,7 +56,7 @@ func TestHandleCreate(t *testing.T) {
 	got, want := &core.Cron{}, dummyCron
 	json.NewDecoder(w.Body).Decode(got)
 
-	ignore := cmpopts.IgnoreFields(core.Cron{}, "Next")
+	ignore := cmpopts.IgnoreFields(core.Cron{}, "Next", "Prev")
 	if diff := cmp.Diff(got, want, ignore); len(diff) != 0 {
 		t.Errorf(diff)
 	}
--- a/trigger/cron/cron.go
+++ b/trigger/cron/cron.go
@@ -46,6 +46,9 @@ type Scheduler struct {
 
 // Start starts the cron scheduler.
 func (s *Scheduler) Start(ctx context.Context, dur time.Duration) error {
+	// first execution will  update only cron info, won't trigger jobs
+	s.run(ctx, true)
+
 	ticker := time.NewTicker(dur)
 	defer ticker.Stop()
 
@@ -54,15 +57,22 @@ func (s *Scheduler) Start(ctx context.Context, dur time.Duration) error {
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-ticker.C:
-			s.run(ctx)
+			start := time.Now()
+			s.run(ctx, false)
+			elapsed := time.Since(start)
+			logrus.Infof("cron: job processed in [%s]", elapsed)
+			//check if processing time  greater than cron interval
+			if elapsed.Seconds() > dur.Seconds() {
+				logrus.Warnf("cron: job sched has been take [%s] more than scheduled time [%s] ", elapsed, dur)
+			}
 		}
 	}
 }
 
-func (s *Scheduler) run(ctx context.Context) error {
+func (s *Scheduler) run(ctx context.Context, isfirst bool) error {
 	var result error
 
-	logrus.Debugln("cron: begin process pending jobs")
+	logrus.Debugf("cron: begin process pending jobs First[%t]", isfirst)
 
 	defer func() {
 		if err := recover(); err != nil {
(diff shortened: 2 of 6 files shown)


handler/api/repos/crons/create_test.go
```diff
@@ -56,7 +56,7 @@ func TestHandleCreate(t *testing.T) {
 	got, want := &core.Cron{}, dummyCron
 	json.NewDecoder(w.Body).Decode(got)
 
-	ignore := cmpopts.IgnoreFields(core.Cron{}, "Next")
+	ignore := cmpopts.IgnoreFields(core.Cron{}, "Next", "Prev")
```
toni-moreno: Now Prev Field will change on each cron job creation

trigger/cron/cron_test.go
```diff
@@ -44,9 +44,6 @@ func TestCron(t *testing.T) {
 
 	before := time.Now().Unix()
 	checkCron := func(_ context.Context, cron *core.Cron) {
-		if got, want := cron.Prev, int64(2000000000); got != want {
-			t.Errorf("Expect Next copied to Prev")
-		}
```
toni-moreno: Prev won't be anymore copied from Next to avoid alignment

trigger/cron/cron.go
```diff
@@ -46,12 +46,21 @@ type Scheduler struct {
...
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case <-time.After(dur):
+		case <-ticker.C:
+			start := time.Now()
 			s.run(ctx)
+			elapsed := time.Since(start)
+			logrus.Infof("cron: job processed in [%s]", elapsed)
```
tboerger: You could use the structured logger with some fields

trigger/cron/cron.go
```diff
@@ -91,17 +100,33 @@ func (s *Scheduler) run(ctx context.Context) error {
...
+		prev := time.Unix(job.Prev, 0)
+		next := sched.Next(prev)
+
+		//If past in time, retreive the next valid time until next time will be greater than now
+		for next.Before(now) {
+			prev = next
+			next = sched.Next(next)
+			// if next still under now this scheduled  will be skipped , else will be executed !
+			if next.Before(now) {
+				logger.Warnf("skipped cron job [%s]  scheduled in the past at : [%s]  NOW [%s]", job.Name, prev, now)
```
tboerger: Too much spaces?

trigger/cron/cron.go
```diff
@@ -91,17 +100,33 @@ func (s *Scheduler) run(ctx context.Context) error {
...
+			}
+		}
+
+		logger.Infof("cron job [%s]   scheduled  at : [%s] has been triggered at [%s] ", job.Name, prev, now)
```
tboerger: Same here

trigger/cron/cron.go
toni-moreno: Hi @tboerger thank you for the quick reply. 
What exactly would you refer? Are you talking aboult spaces in the log message?

tboerger: Yes, just refering to the duplicated spaces within the log message, so really nothing critical.

toni-moreno:
Hi @tboerger I've rebased the PR and also added a few fixes. 
a) skip first execution , f not skipped all old scheduled jobs will be triggered at once first time , after first execution each one will be triggered with its own timing, not aligned.
b)  when next scheduled time is equal to now(), I was scheduled twice.

Now it works much better than previous PR

magicoder10:
Can this PR be merge?! Cron has been repeatedly triggering multiple builds simultaneously for me, which wasted lots of my time to add logs and debug Drone's source code. Also wasted me hours trying to migrate from Drone CI to Github actions

bot2-harness:

PR #2907, opened 2020-01-20
Improve the logging for event matching

andrewhowdencom:
While developing:

  https://github.com/drone/drone/pull/2902

It was difficult to verify which events the server was receiving and
how. This commit was applied to the server to make transparent what
events the solver was solving for and immediately cleared up the issue.

It stands to reason it might be useful for other users debugging their
pipelines.

### DIFF
--- a/trigger/trigger.go
+++ b/trigger/trigger.go
@@ -16,6 +16,7 @@ package trigger
 
 import (
 	"context"
+	"fmt"
 	"runtime/debug"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


bradrydzewski:
logger supports `Infof` so that you don't need to use `fmt`, however, our convention is to place this data in a field as opposed to inside the message directly.

andrewhowdencom:
lol I looked right past the logs that were already there. My bad!

tboerger:
Next time you can simply add the requested changes and append that commit, while merging we are able to squash the commits ;)

PR #2915, opened 2020-01-30
Fix incorrect return value from toMap function

mcnicholls:
Github deployments can include a `payload` object that is read by drone and the resulting key/value pairs should be injected into the build step as environment variables.

This fix correctly returns the constructed map of key/values from the `payload` object so they can be injected into the builds environment later.


bradrydzewski:
good catch, thanks!

PR #2919, opened 2020-02-12
Allow overriding .drone.yml path globally

hatsuyuki15:
### Context for this change:  
We are using starlark for all of our pipelines.   
Currently, we have to manually set the configuration file to `.drone.star` everytime we activate a repository.  
We want drone to use `.drone.star` as default configuration file for newly created repos.

### Change in this PR:
- Add a configuration `DRONE_YAML_FILE` that allows overidding the default `.drone.yml`.
- `DRONE_YAML_FILE` is default to `.drone.yml` for backward compatibility

### DIFF
--- a/cmd/drone-server/config/config.go
+++ b/cmd/drone-server/config/config.go
@@ -300,6 +300,7 @@ type (
 
 	// Yaml provides the yaml webhook configuration.
 	Yaml struct {
+		File       string `envconfig:"DRONE_YAML_FILE" default:".drone.yml"`
 		Endpoint   string `envconfig:"DRONE_YAML_ENDPOINT"`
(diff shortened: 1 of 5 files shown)


tboerger:
You could even do that already today with a global webhook plugin.

hatsuyuki15:
I certainly didn't think of that!
That's really a clever way to use webhook to solve the problem.
Still, I think it'd be better to extract out the hard coded ".drone.yml"
and have a more simple & reliable (no network callback) solution rather
than relying on the  webhook.

ashwilliams1:
This is a recommended use case for a configuration extension. You can learn more about creating extensions here: https://docs.drone.io/extensions/configuration/

Issue #2921, opened 2020-02-18
webhooks on build complete event


vyckou:
Hi, 

In documentation it is stated , that webhooks are emitted on following events:
```
User is created
User is deleted
Repository is activated
Repository is de-activated
Build is created
Build is updated
Build is completed
```
ref: https://docs.drone.io/webhooks/overview/

while in the code https://github.com/drone/drone/blob/1ddf7963feb98ae7fed7dd2834389441f2401c77/core/webhook.go#L29
there is no such type as `completed` 
Any plans to update documentation or implement `completed` action?

ashwilliams1:
> Build is completed

a webhook is emitted when a build is complete.  the action type is updated and the build status will no longer be pending or running.  this is how you know it is completed.

vyckou:
Unfortunately that is not true:
There is a webhook content, which arrived when pipeline run ended:
All steps have status `success`, but build status is still `running`
`{
    "event": "build",
    "action": "updated",
    "repo": {
        "id": 1625,
        "uid": "2639",
        "user_id": 6,
        "namespace": "DRON",
        "name": "app-cicd-test",
        "slug": "DRON/app-cicd-test",
        "scm": "",
        "git_http_url": "https://syst-stash.bitbucket.server.net/scm/dron/app-cicd-test.git",
        "git_ssh_url": "ssh://git@syst-stash.bitbucket.server.net:7999/dron/app-cicd-test.git",
        "link": "https://syst-stash.bitbucket.server.net/projects/DRON/repos/app-cicd-test/browse",
        "default_branch": "master",
        "private": true,
        "visibility": "private",
        "active": true,
        "config_path": ".drone.yml",
        "trusted": true,
        "protected": false,
        "ignore_forks": false,
        "ignore_pull_requests": false,
        "timeout": 60,
        "counter": 169,
        "synced": 1568632832,
        "created": 1568632832,
        "updated": 1573563025,
        "version": 173,
        "build": {
            "id": 1305,
            "repo_id": 1625,
            "trigger": "fila",
            "number": 169,
            "status": "running",
            "event": "push",
            "action": "",
            "link": "",
            "timestamp": 0,
            "message": "readme.md edited online with Bitbucket",
            "before": "",
            "after": "1abbca556daecc3dc61a63d111696ee85106d17b",
            "ref": "refs/heads/e-f-g-test",
            "source_repo": "",
            "source": "e-f-g-test",
            "target": "e-f-g-test",
            "author_login": "be1177",
[...]

PR #2923, opened 2020-02-18
Added webhook type of build completed event

vyckou:
Currently when the build/stage is completed, "updated" type of the webhook is fired.
"updated" type of web hook is not being removed, not to break any existing integrations and additional "completed" web hook is introduced, which would help to identify for the webhook receivers, when actually the build ended

Issue: https://github.com/drone/drone/issues/2921

### DIFF
--- a/core/webhook.go
+++ b/core/webhook.go
@@ -27,11 +27,12 @@ const (
 
 // Webhook action types.
 const (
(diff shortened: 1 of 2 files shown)


bot2-harness:

Issue #2823, opened 2019-09-10
Segmentation Fault on ARM devices


KN4CK3R:
Drone fails on ARM devices with a segmentation fault (see https://discourse.drone.io/t/drone-1-3-0-silently-fails-on-armv7/5417/3 and https://discourse.drone.io/t/run-drone-linux-arm-on-raspberry-pi-3-with-error-exit-code-132/5590/ ) The last working version was 1.2.3. I compiled all commits since the v1.2.3 tag and it stops working after the merge of the pull request #2788 (@Lucretius). I got it working again after this change:
```diff
diff --git "a/cmd/drone-server/inject_store.go" "b/cmd/drone-server/inject_store.go"
index c34aebfb..8357b415 100644
--- "a/cmd/drone-server/inject_store.go"
+++ "b/cmd/drone-server/inject_store.go"
@@ -92,12 +92,12 @@ func provideLogStore(db *db.DB, config config.Config) core.LogStore {
 		return logs.NewCombined(p, s)
 	}
 	if config.AzureBlob.ContainerName != "" {
-		p := logs.NewAzureBlobEnv(
+		/*p := logs.NewAzureBlobEnv(
 			config.AzureBlob.ContainerName,
 			config.AzureBlob.StorageAccountName,
 			config.AzureBlob.StorageAccessKey,
 		)
-		return logs.NewCombined(p, s)
+		return logs.NewCombined(p, s)*/
 	}
 	return s
 }
```
I don't know why the if check doesn't work as expected but `NewAzureBlobEnv` gets called which causes the segmentation fault.

Lucretius:
Sorry to hear my PR is causing issues.  Definitely did not test my code on a raspberry pi or an ARM device.  Did you try logging out what exactly what it thinks the container name is?  Since it’s appearing to skip the S3 check, I wonder if something is getting set somehow for AzureBlob?

KN4CK3R:
I tried to output the values with `fmt.println` but it didn't print the text. I haven't worked with Go until now so that may be wrong. If you could give me code for logging I can test it on a raspberry.

Lucretius:
I believe the project uses [logrus](https://github.com/sirupsen/logrus) for its logging, output via fmt.Println may be suppressed.  Try also logging just outside of that if-statement.  If nothing is being written out and it's simply ignoring the if-check maybe  it comes down to how ARM process instructions from compiled Go code? 

If the logging doesn't work see if changing the two if checks into an if-elseif-else checks has any effect.

[...]

PR #2936, opened 2020-03-11
Bump to Go 1.14

jeffwidman:
Bump to `go` `1.14`.

Previously this pinned to `1.13.0`, but I think it's sufficient to pin only to the minor version and not worry about pinning the patch release.  The odds are higher that no one will remember to bump the patch release than that the patch upgrade would actually break something.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,12 +9,12 @@ platform:
 
 steps:
 - name: test
-  image: golang:1.13.0
+  image: golang:1.14
   commands:
   - go test ./...
   
 - name: build
-  image: golang:1.13.0
+  image: golang:1.14
   commands:
   - sh scripts/build.sh
   environment:
@@ -48,7 +48,7 @@ platform:
 
 steps:
 - name: build
-  image: golang:1.13.0
+  image: golang:1.14
   commands:
   - sh scripts/build.sh
   environment:
@@ -86,7 +86,7 @@ platform:
 
 steps:
 - name: build
-  image: golang:1.13.0
+  image: golang:1.14
   commands:
   - sh scripts/build.sh
   environment:


mavimo:
@jeffastorey IMHO also the patch version should be pinned otherwise a build is not completely reproducible, since the base image should change out of our control.. is like having the `go.sum` that do not include the package sha 😄

jeffwidman:
> IMHO also the patch version should be pinned otherwise a build is not completely reproducible, since the base image should change out of our control.. 

I'm aware of the theoretical benefits of doing that... I mentioned it in my PR description. 

However, as mentioned, practically I've observed that people forget to bump pins _far_ more than they struggle with corner cases... as long as it's exposed somewhere exactly what version was used when the build broke, then it's always possible to pin the patch level temporarily if needed while bug hunting.

This is simply my personal opinion, so if the maintainers prefer pinning the patch version I am happy to update the PR.

KN4CK3R:
This would fix #2823 again. The error is back and drone doesn't work on Raspberry Pis at the moment.

```shell
$ docker run ... golang:1.13.0 go build -ldflags "-extldflags \"-static\"" ... ./cmd/drone-server
$ ./drone-server
Segmentation fault
$ docker run ... golang:1.14.0 go build -ldflags "-extldflags \"-static\"" ... ./cmd/drone-server
$ ./drone-server
license: ""
...
```

tboerger:
@bradrydzewski I think that should be safe to merge, right?

jeffwidman:
nudge...

jeffwidman:
I noticed this is no longer needed as it was implemented on `master` by @bradrydzewski ... IMO it discourages rather than encourages community contributions when rather than responding to a PR that someone took the time to open and even nudge on you instead implement the change directly.

PR #2958, opened 2020-04-29
dot is valid for username

okian:
Currently Drone doesn't allow (.) in username but thats valid (at least in gitlab)
example: https://gitlab.com/hadi.farnoud
Screenshot from 2020-04-30 00-59-45

### DIFF
--- a/core/user.go
+++ b/core/user.go
@@ -94,7 +94,7 @@ func (u *User) Validate() error {
 	switch {
 	case !govalidator.IsByteLength(u.Login, 1, 50):
 		return errUsernameLen
-	case !govalidator.Matches(u.Login, "^[a-zA-Z0-9_-]+$"):
+	case !govalidator.Matches(u.Login, "^[.a-zA-Z0-9_-]+$"):
 		return errUsernameChar
 	default:
 		return nil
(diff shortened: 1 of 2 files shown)


core/user.go
```diff
@@ -94,7 +94,7 @@ func (u *User) Validate() error {
 	switch {
 	case !govalidator.IsByteLength(u.Login, 1, 50):
 		return errUsernameLen
-	case !govalidator.Matches(u.Login, "^[a-zA-Z0-9_-]+$"):
+	case !govalidator.Matches(u.Login, "^[.a-zA-Z0-9_-]+$"):
```
tboerger, line 97: Not sure about that, not should the dot be escaped?

okian: It doesn't need to be escaped inside brackets. also i add a test for it and it didn't break any previous cases.

tboerger:
@bradrydzewski I guess this would be safe to merge?

okian:
Any update?

hadifarnoud:
I'm also waiting on this guys @tboerger @bradrydzewski

tphoney (approved):

tphoney:
Thanks for you patience everyone and special thanks to @okian for the fix and tests. 🎆

PR #2966, opened 2020-05-12
Do not ignore Gitea TagHook

CirnoT:
Gitea no longer sends push webhook for tag events, as per https://github.com/go-gitea/gitea/pull/11082


ashwilliams1:
Is this a breaking change from Gitea?

CirnoT:
It was not an intention of that PR to introduce a breaking change, Drone just happened to rely on that behavior.

ashwilliams1:
> It was not an intention of that PR to introduce a breaking change, Drone just happened to rely on that behavior.

The previous behavior was the same as GitHub and GitLab, but after this pull request it no longer behaves in the same way. Are you sure this was not a regression?

lafriks:
Changes in Gitea has been reverted

CirnoT:
Closing as per https://github.com/go-gitea/gitea/pull/11397

PR #2973, opened 2020-05-29
 to fixed: call to commits.FindRef returns 'not found' when using gitea

18981119465:
to fixed: call to commits.FindRef returns 'not found' when using gitea


bradrydzewski:
This does not require a code change for Drone, we just need to update the go-scm dependency to include [this patch](https://github.com/drone/go-scm/pull/61/files). Also please try to avoid updating all dependencies, and instead only send a pull request that updates go-scm. This can be done with `go get -u github.com/drone/go-scm`

PR #2982, opened 2020-06-21
feat: delete scm webhook when disable repo.

dakele123:
when disable repo, drone delete scm webhook.

for this issue: https://discourse.drone.io/t/disabling-repo-on-drone-does-not-delete-webhook/5146

### DIFF
--- a/handler/api/api.go
+++ b/handler/api/api.go
@@ -183,7 +183,7 @@ func (s Server) Handler() http.Handler {
 			).Post("/", repos.HandleEnable(s.Hooks, s.Repos, s.Webhook))
 			r.With(
 				acl.CheckAdminAccess(),
-			).Delete("/", repos.HandleDisable(s.Hooks, s.Repos, s.Webhook))
+			).Delete("/", repos.HandleDisable(s.Users, s.Hooks, s.Repos, s.Webhook))
 			r.With(
 				acl.CheckAdminAccess(),
 			).Post("/chown", repos.HandleChown(s.Repos))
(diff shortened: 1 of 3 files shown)


bradrydzewski:
thanks for the pull request.  a few comments

1. deleting the webhook should probably come after the database update
2. if deleting the webhook fails it should output warning logs, but should not fail the API call. There are valid scenarios where we expect deleting the webhook to fail (for example, repo was deleted in github)
3. the repo owner should be used when making the API call. see https://github.com/drone/drone/blob/master/handler/api/repos/builds/create.go#L52. If returning the owner from the database fails, it should not fail the api call, and deleting the webhook should be skipped.

dakele123:
> thanks for the pull request. a few comments
> 
> 1. deleting the webhook should probably come after the database update
> 2. if deleting the webhook fails it should output warning logs, but should not fail the API call. There are valid scenarios where we expect deleting the webhook to fail (for example, repo was deleted in github)
> 3. the repo owner should be used when making the API call. see https://github.com/drone/drone/blob/master/handler/api/repos/builds/create.go#L52. If returning the owner from the database fails, it should not fail the api call, and deleting the webhook should be skipped.
> 
> also please be sure to sign the CLA

thanks for reply.

i will check the code and
1.  move delete func after update the database updated.
2. delete the return statement when webhook delete failed.
3. refer to the link and check the repo owner.
4. signed the CLA.

if there is any mistakes in comprehension, let me know.

bot2-harness:

PR #2998, opened 2020-07-20
Add configurable timeout for config, convert, and validate plugins

mach6:
Also run 'go fmt' on changed files

--- 
Justification for this PR - See 2. in https://github.com/bitsbeats/drone-tree-config/issues/26

Though the issue was specific to a `config` plugin, I implemented the timeout option for `validate` and `convert` also. Projecting a bit -- it seemed useful to do so.

### DIFF
--- a/cmd/drone-server/config/config.go
+++ b/cmd/drone-server/config/config.go
@@ -98,7 +98,7 @@ type (
 	}
 
 	Cleanup struct {
-		Disabled  bool         `envconfig:"DRONE_CLEANUP_DISABLED"`
+		Disabled bool          `envconfig:"DRONE_CLEANUP_DISABLED"`
 		Interval time.Duration `envconfig:"DRONE_CLEANUP_INTERVAL"         default:"24h"`
 		Running  time.Duration `envconfig:"DRONE_CLEANUP_DEADLINE_RUNNING" default:"24h"`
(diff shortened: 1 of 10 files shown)


mach6:
> Great work, now it's just missing some retry logic :P

Perhaps another PR. :)

mach6:
@tboerger thanks for the feedback and review.

mach6:
@tboerger and reviewers: it turns out there is another 1 minute timeout in `drone-go` that always wins -- https://github.com/drone/drone-go/blob/master/plugin/internal/client/client.go#L96 ... As a result, this PR will have no impact until the `drone-go` timeout can be influence by the config. I will submit a change to `drone-go` and update this PR soon.

mach6:
This PR is linked with the following proposed change in `drone-go` - https://github.com/drone/drone-go/pull/52  

If it helps to validate these changes, the following config plugin can be used;
https://github.com/mach6/drone-delay-config

mach6:
@tboerger @bradrydzewski Should I go ahead and add a TIMEOUT config option for all of the extensions?  Also, safe to assume this will be merged once a new version of `drone-go` is released? Anything you need from me?

mach6:
@bradrydzewski anything you need from me on this PR? Are we waiting for a new release of `drone-go` perhaps?

PR #3009, opened 2020-08-13
fix some typing error

OhBonsai:


bradrydzewski:
First, thanks for finding this!  unfortunately if we fix the name of the json field it will break all of the runners, which currently rely on this tag being incorrect.  We will probably have to wait until Drone 2.0 to correct this mistake to avoid breaking things.

PR #3023, opened 2020-09-23
fix go-lint's warning

colynn:


marko-gacesa:
Thank you for your effort, however this change does not improve the project in any way. I'm not saying lint warnings should be ignored, but you could have placed a real godoc comment rather than just two dots :)

PR #3030, opened 2020-10-27
Add path support different than root

### LINKED ISSUE #475: Relative url path
I can't find anything in the documentation, nor the sourcecode about a way to put drone.io under example example.com/ci/, instead of using a whole dedicated domain. It looks like all the links in the html template files uses hardlinking, so I guess this is not currently possible.

Is there a plan to implement a more dynamic base url? Its important for many projects.

jlsalvador:
Drone expected that everything will be served on the web root. This patch
enables support from another path than root. `HandleIndex` will inject the
value for `window.BASE_URL` used by the frontend Drone-UI.

New environment variable `DRONE_SERVER_PATH`, by default "/".
Replaced some hard-coded endpoints.

Fixes #475


ashwilliams1:
Hi there, we have received a few different pull requests for this feature in the past, however, we have politely declined to merge such features. If you plan on using a shared server to host Drone, instead of sub paths, we officially recommend using subdomains. Using subdomains has become much easier over the past few years thanks to lets encrypt and acme. Alternatively you could setup a dedicated VM for just a few dollars a month.  Ultimately this is not a maintenance burden we want to bear going forward, given that there are other viable options for installing and running Drone. I am closing this because I do not want you to invest any more time in something that would not get merged.

jlsalvador:
> Hi there, we have received a few different pull requests for this feature in the past, however, we have politely declined to merge such features. If you plan on using a shared server to host Drone, instead of sub paths, we officially recommend using subdomains. Using subdomains has become much easier over the past few years thanks to lets encrypt and acme. Alternatively you could setup a dedicated VM for just a few dollars a month. Ultimately this is not a maintenance burden we want to bear going forward, given that there are other viable options for installing and running Drone. I am closing this because I do not want you to invest any more time in something that would not get merged.

Hello Ash. I'm not going to hide my pity about your decision, but it's fair enough. If there are not space here to improve or discuss around this feature, please close this other Pull Request that is related. https://github.com/drone/drone-ui/pull/344
Have a nice day.

PR #3038, opened 2020-11-18
fix(trigger) : Change order of skip triggers

beaucoralk:
Do skip triggers after the base hook object has been filled with commit message when it is empty.

It should fix the issue that the [skip ci] didn't work for recent Bitbucket version because the commit message isn't on the push event but the hash still being in it.

linked to : https://github.com/drone/go-scm/pull/84

### DIFF
--- a/trigger/trigger.go
+++ b/trigger/trigger.go
@@ -94,21 +94,6 @@ func (t *triggerer) Trigger(ctx context.Context, repo *core.Repository, base *co
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


beaucoralk:
I've not realy tested it, but the goal is to do skip triggers after the commit message has been filled with the "After" property of the hook (Bitbucket push hook has been modified in https://github.com/drone/go-scm/pull/84 to add the hash commit).

PR #3041, opened 2020-11-23
Adds ExtVar in Jsonnet

Upgreydd:
```
External Variables
std.extVar(x)
If an external variable with the given name was defined, return its string value. Otherwise, raise an error.
```
Allows you to access environment variables at jsonnet level. Thanks to that you can do more advanced logic


tboerger:
Same here as for your other PR, AFAIK the operator package have been deprecated.

PR #3042, opened 2020-11-23
Adds TLA (top-level-arguments) in Jsonnet

Upgreydd:
```
Top-level arguments
Alternatively the same code can be written using top-level arguments, where the whole config is written as a function. How does this differ to using external variables?

Values must be explicitly threaded through files
Default values can be provided
A config with top-level arguments can also be imported as a library and called as a function with parameters passed in
Generally, top-level arguments are the safer and easier way to parameterize an entire config, because the variables are not global and it's clear what parts of the config are dependent on their environment. However, they do require more explicit threading of the values into other imported code.
```
Thanks to that in code you can use all environment variables as jsonnet arguments - for example:
```
function(DRONE_BRANCH) {
  "branch": DRONE_BRANCH
}
```
will return
```
{ "branch": "master" }
```

### DIFF
--- a/operator/runner/env.go
+++ b/operator/runner/env.go
@@ -83,7 +83,7 @@ func stageEnviron(stage *core.Stage) map[string]string {
 	}
 }
 
-func buildEnviron(build *core.Build) map[string]string {
+func BuildEnviron(build *core.Build) map[string]string {
 	env := map[string]string{
 		"DRONE_BRANCH":               build.Target,
 		"DRONE_SOURCE_BRANCH":        build.Source,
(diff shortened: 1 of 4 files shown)


tboerger:
Generally it's a good idea, but I think the operator package is deprecated as it's related to the old kubernetes operator.

Upgreydd:
@tboerger please share more details. As I see in code it's still used by runner.

tboerger:
That's something up to @bradrydzewski, he knows which packages are not deprecated and he knows which types haven't been exported inentionally.

PR #3047, opened 2020-12-20
Update go-chi version to go mod compatible

lafriks:


6543:
@bradrydzewski anything holding this back?!?

marko-gacesa:
Thanks @lafriks. We don't see how this change could benefit the project in any way. However, recently we started a proposal site https://github.com/drone/proposal where you can express this problem and where we can also discuss it. Thank you for your patience.

PR #3050, opened 2020-12-30
stash needs to get ref via branch, cant via heads same as bitbucket

bakito:
As proposed in https://discourse.drone.io/t/panic-or-undefined-error-when-create-build-from-bitbucket-repository/8549/6 using the workaround from bitbucket also in stash solves the issue, that new builds can not be created without providing a commit.


eoinmcafee00:
Hey @bakito, thanks for this. If you wish to pull the latest, it should be fixed. I'll close this PR.

PR #3052, opened 2021-01-07
Support branch "create" action

quantonganh:
https://github.com/drone/drone/issues/2685

Discussion: https://gitter.im/drone/drone?at=5ff5c66acd31da3d9a961645

### DIFF
--- a/service/hook/parser/parse.go
+++ b/service/hook/parser/parse.go
@@ -163,6 +163,10 @@ func (p *parser) Parse(req *http.Request, secretFunc func(string) string) (*core
 				Sender:       v.Sender.Login,
 			}
 		}
+		// When pushing to a new branch, the SHA of the most recent commit is empty
+		if v.Before == emptyCommit {
+			hook.Action = core.ActionCreate
+		}
 		repo = &core.Repository{
 			UID:       v.Repo.ID,
 			Namespace: v.Repo.Namespace,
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -75,3 +75,5 @@ require (
 replace github.com/h2non/gock => gopkg.in/h2non/gock.v1 v1.0.14
 
 go 1.13
+
+replace github.com/drone/go-scm => github.com/quantonganh/go-scm v1.8.1-0.20210107022054-cb9eef77a93f
```
techknowlogick, line 79: This should be removed prior to merging (but after blocking PR is merged)

quantonganh: Of course. It's temporary to make the test passed.

quantonganh:
As @bradrydzewski [pointed out](https://github.com/drone/go-scm/pull/92#issuecomment-755852095), no need to add a new field to know if a branch has been created. We can do that by checking if the before commit is empty. Updated my PR.

tphoney:
Hi @quantonganh i am unsure of this newly added behaviour, as it changes what existing drone users would expect. To move this forward in an open way i would suggest adding a proposal here https://github.com/drone/proposal. So we can design with the rest of the community what the behaviour should be and if the community as a whole needs this functionality.

For now i am going to close this PR. Thank you for your submission and the time and effort you have put in.

Issue #2056, opened 2017-05-24
Control PR build through comments


donny-dont:
Drone should provide a way to control PR builds through the comment system of the remote.

For example a contributor could specify in the comments to `do: build` and then Drone would kick off a build.

Potential use cases
* Gating a PR build so it is not run until a maintainer approves a run
* Run different configurations such as turning a feature flag on/off

Examples in the wild
* [Jenkins Plugin](https://github.com/jenkinsci/ghprb-plugin)
* [GitLab Integration](https://gitlab.kitware.com/cmake/cmake/merge_requests/798)

donny-dont:
Looking to formalize this into a feature

bradrydzewski:
do you think there should be a formal notation / language similar to `[ci skip]`?

```sh
[drone approve]           # approve the gated build
[drone deploy production] # deploy to production
```

or should we let individual projects define regular expressions or matching criteria in the yaml?

```
triggers:
  - match: deploy to production
    action: deploy
    environment:
      - FOO=bar
      - BAZ=boo
```

in terms of turning feature flags on / off would this be done by injecting environment variables into the build? I included an environment section in the above example for this purpose.

mohanraj-r:
Looking to switch to drone from Jenkins, and this is the feature that is holding me back. 

I guess options for the following would make sense
* auto-build on PR create : boolean
 - auto-build on new commits pushed to PR : boolean 
    - not sure if this needs to be a separate option or rolled in with the PR create option
* auto-merge/close PR on build success: boolean
  - not sure if user config about the remote branch to merge to would be needed for this option
* build-trigger phrase: regex
  - build-trigger user whitelist: list of users who can use the trigger phrase to trigger builds
* PR-merge phrase: regex
  - merge-trigger user whitelist: list of users who can use the trigger phrase to merge PRs

Optionally:
* ability to specify actions after merge (e.g. test and/or build after merge) 
  - But I guess this could be specified as trigger actions on the target branch being merged into?
* I am not sure if it would be useful to have control over status updates from drone on the PR (e.g. control over msg string and customizing using env/build vars).

bradrydzewski:
> auto-build on PR create : boolean
> auto-build on new commits pushed to PR : boolean

fyi drone already does these

mohanraj-r:
Thanks for clarification @bradrydzewski 
Didn't know that drone supports PR triggers - just found http://docs.drone.io/github.html#hooks

> and these can be done via plugins today

[...]

PR #3053, opened 2021-01-09
Add keyword trigger condition

kdumontnu:
This enables a new trigger for keywords in the commit message or title. The usage is as follows:

```
trigger:
  keyword:
  - "*[RUN PIPELINE]*"
```
Should only run if the keyword is matched.

or

```
trigger:
  keyword:
    exclude:
    - "*[SKIP DEPLOY]*"
```
Should not run if the keyword is matched.

I considered simplifying the notation by appending `*` to each end of the keyword strings, but it would give the user less control on matching specific strings.

Note: The way this is currently implemented follows the same syntax as other triggers, but requires adding the `Keyword` condition to `drone-yaml`. However, I see that the GitHub repository is currently archived, so I'm unable to open a PR. 

On further inspection it looks like drone-yaml `Conditions` essentially expects each argument to be a "path" in the matching operation. Passing an arbitrary string (message or title) will probably result in odd behavior. The solution here appears to be to simply create a separate Match function for Condition that uses go `strings.Contains` and remove the wildcard operators above. Again, this is in drone-yaml, which appears to be private.

### DIFF
--- a/trigger/skip.go
+++ b/trigger/skip.go
@@ -53,6 +53,10 @@ func skipCron(document *yaml.Pipeline, cron string) bool {
 	return !document.Trigger.Cron.Match(cron)
 }
 
+func skipKeyword(document *yaml.Pipeline, message string, title string) bool {
+	return !(document.Trigger.Keyword.Match(message) || document.Trigger.Keyword.Match(title))
+}
+
 func skipMessage(hook *core.Hook) bool {
 	switch {
 	case hook.Event == core.EventTag:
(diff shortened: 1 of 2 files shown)


kdumontnu:
Just saw [this issue](https://github.com/drone/drone/issues/2056), which appears closely related to this. Seems like there was no consensus on the proper syntax, so I propose to leave it up to the user which strings to match.

tphoney:
Hey @kdumontnu this seems to have went a little stale, there are a few things i would like to raise. It is unclear how this would work and it needs a little bit more explanation. I would suggest creating a proposal for this feature here https://github.com/drone/proposal first . So we can gather feedback on this would work. I am closing the issue for now, due to staleness not signing the CLA and failing tests.

PR #3060, opened 2021-01-26
fix/change gitlab status udpate logic

zr9:
Gitlab statuses may transition in their CI
in case if status is already set Gitlab will return err
which will be shown as warning in drone logs

Those patch add logic to check if status we trying to set is different from the one in Gitlab

### DIFF
--- a/service/status/status.go
+++ b/service/status/status.go
@@ -81,6 +81,26 @@ func (s *service) Send(ctx context.Context, user *core.User, req *core.StatusInp
 		return err
 	}
 
+	if s.client.Driver == scm.DriverGitlab {
+		/*
+		NOTE:
+			Gitlab statuses may transition in their CI
+			in case if status is already set Gitlab will return err
+		*/
+		got, _, err := s.client.Repositories.ListStatus(ctx, req.Repo.Slug, req.Build.After, scm.ListOptions{});
+
+		if err == scm.ErrNotSupported {
+			return nil
+		}
+
+		if err != nil {
+			return err;
+		}
+
+		if len(got) > 0 && got[0].State == convertStatus(req.Build.Status) {
+			return nil
+		}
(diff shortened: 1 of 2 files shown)


bradrydzewski:
we typically encapsulate most vendor-specific logic in the [github.com/drone/go-scm](https://github.com/drone/go-scm) repository.  However, before we make that decision, It would be interesting to know if this could be simplified to read the contents of the error message so that we could avoid a second API call.  For example, like this:

```diff
	_, _, err = s.client.Repositories.CreateStatus(ctx, req.Repo.Slug, req.Build.After, &scm.StatusInput{
		Title:  fmt.Sprintf("Build #%d", req.Build.Number),
		Desc:   createDesc(req.Build.Status),
		Label:  createLabel(s.name, req.Build.Event),
		State:  convertStatus(req.Build.Status),
		Target: fmt.Sprintf("%s/%s/%d", s.base, req.Repo.Slug, req.Build.Number),
	})
	if err == scm.ErrNotSupported {
		return nil
	}
+	if err != nil && strings.Contains(err.Error(), "<some gitlab error message string>") {
+		return nil
+	}
	return err
```

zr9:
i thought about `go-scm` reasons to place it in status was next:
1. There is already github logic in status file, so i made similar way
2. It will require update of `CreateStatus` so if it used in other places that will double the request amount

For read error, yup will work, that was my first idea, however the issue with it, so it masks error, not fixes, since normally it still occurs just will be hidden from drone/user

Gustavotd2 (commented):
K

tphoney:
Hi @zr9  , it does look like putting this change into go-scm is a much cleaner solution and it means more people can avail of the fix.
I am closing this pull request. Thank you for your time and effort on this.

PR #3062, opened 2021-02-04
bumped aws-go-sdk version

prune998:
The outdated version of aws-go-sdk (`github.com/aws/aws-sdk-go v1.15.57`) does not support native `assumeRole` and requires to set the AWS credentials in the environment before running `drone-server`.
The latest version can natively use the `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` env variables to assume a role.

This support is needed when running drone-server in an AWS EKS cluster.

on the side:
A TON of other packages used by Drone are also outdated and should require a review. (Note the latest `github.com/Azure/azure-storage-blob-go` package is not compatible because of an API change).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Azure/azure-storage-blob-go v0.7.0
 	github.com/Azure/go-autorest/autorest/adal v0.8.3 // indirect
 	github.com/asaskevich/govalidator v0.0.0-20180315120708-ccb8e960c48f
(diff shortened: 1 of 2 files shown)


bradrydzewski:
thanks, regarding dependencies, we do not attempt to track the latest version for each dependency.  we update dependencies if there are bugs, security fixes or new features that we need to take advantage of, but otherwise we prefer to keep our dependency tree static and stable to reduce the potential for regressions.

PR #3064, opened 2021-02-07
add trigger environment variable

zc2638:
When the `Promote` event is triggered, it is expected that the trigger person can be obtained for further judgment


zc2638:
@bradrydzewski

tboerger (approved):

bradrydzewski:
environment variables are set using the github.com/drone/runner-go library.  The code in this pull request is actually deprecated and will be removed, so there is no need to update.  I see you already made a pr to runner-go.  We will move the discussion to https://github.com/drone/runner-go/pull/4.

PR #3071, opened 2021-02-26
[feature] allow starlark loading in trusted repo

jmcelwain:
We tested switching to the starlark plugin by using `drone-cli` to template our `.drone.yml` in a pull request before merging and switching over entirely to starlark. Unfortunately, this exposes a gap between `drone-cli` and `drone` server, which does not support module loads. For us, this is the primary reason we want to use starlark as our `.drone.yml` has grown to an unpleasant size for a monolithic piece of configuration.

This PR makes an assumption we don't want to let users load modules in un-trusted repos. I'm not sure that matters in practice, but assumed that might be one reason for not implementing the loader func originally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- support for relative module loading in starlark scripts.
+
 ### Breaking
 - removed deprecated kubernetes integration in favor of official kubernetes runner.
 - removed deprecated nomad integration in favor of official nomad runner.
--- a/plugin/converter/starlark/starlark.go
+++ b/plugin/converter/starlark/starlark.go
@@ -53,6 +53,9 @@ var (
 	// ErrCannotLoad indicates the starlark script is attempting to
 	// load an external file which is currently restricted.
 	ErrCannotLoad = errors.New("starlark: cannot load external scripts")
+
+	// ErrLoadCycle indicates the starlark strip has a cyclic dependency
+	ErrLoadCycle = errors.New("starlark: cycle in load dependency graph")
 )
 
 // New returns a conversion service that converts the
@@ -82,9 +85,17 @@ func (p *starlarkPlugin) Convert(ctx context.Context, req *core.ConvertArgs) (*c
 		return nil, nil
 	}
 
+	// if we're in a trusted repo, we can load relative files
+	var load func (_ *starlark.Thread, module string) (starlark.StringDict, error)
+	if req.Repo.Trusted {
+		load = makeLoad()
+	} else {
+		load = noLoad
+	}
+
 	thread := &starlark.Thread{
 		Name: "drone",
-		Load: noLoad,
+		Load: load,
 		Print: func(_ *starlark.Thread, msg string) {
 			logrus.WithFields(logrus.Fields{
 				"namespace": req.Repo.Namespace,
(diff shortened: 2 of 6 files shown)


bradrydzewski:
hey there, this feature is available using the following extension:
https://github.com/drone/drone-convert-starlark

we may eventually bring this capability into Drone core, but we probably want a more robust snippet implementation that also supports jsonnet and yaml includes. So I do not think we are ready to implement this feature yet until we have an RFC that covers more use cases.

jmcelwain:
@bradrydzewski  Totally understood -- would love the more complex implementation. We're pushing up against the limits of maintainability right now and are looking for solutions.

Unfortunately we are already using a convert plugin (meltwater/drone-convert-pathschanged), and I don't believe it's possible  to use multiple per the docs here: https://docs.drone.io/extensions/conversion/.

Would you accept a PR that allows multiple convert plugins to be chained?

jmcelwain:
Blah, just found at that even using the starlark conversion on the server seems to disable our other external conversion plugin, which is a huge bummer. Will close this but feeling like we're hitting some unfortunate limitations.

bradrydzewski:
You are correct that Drone cannot invoke more than 1 conversion plugin, however, this does not mean Drone cannot satisfy your use case.  Looking at this problem from a different angle, instead of chaining multiple plugins in Drone, you can create a plugin that chains together multiple plugins.

__Step 1__ create a custom plugin that combines two existing plugins:

```
func New(plugin1, plugin2 converter.Plugin) converter.Plugin {
	return &Plugin{ plugin1, plugin2 }
}

type Plugin struct {
	plugin1 converter.Plugin
	plugin2 converter.Plugin
}

func (p *plugin) Convert(ctx context.Context, req *converter.Request) (*drone.Config, error) {
	out, err := p.plugin1.Convert(ctx, req)
	if err != nil {
		return out, err
	}
	if out != nil {
		req.Config = out
	}
	return p.plugin2.Convert(ctx, req)
}
```

__Step 2__ import that starlark and paths changed plugins

```
import (
	plugin1 "github.com/drone/drone-convert-starlark/plugin"
	plugin2 "github.com/meltwater/drone-convert-pathschanged/plugin"
)
```

__Step 3__ pass the starlark and paths changed plugins to your custom plugin:

```
New(
	plugin1.New(...),
	plugin2.New(...),
)
```

The reason we do not chain together plugins in Drone is because there are edge cases that we have not solved for yet.  This may not sound like a big deal on the surface, however, there are tends of thousands of active Drone installations and at this scale, even small edge cases can become a large support burden for us.

jmcelwain:
Thanks @bradrydzewski this is super helpful. We'll give it a shot. We love Drone for not being overwrought, it's just hard not to sometimes feel like your use case is the upper limit :).

I know there are other feature requests for filtering on change sets, but starlark is very attractive to us as a solution. It would be incredible if the diff was passed to starlark and we could just implement all this logic ourselves.

jmcelwain:
@bradrydzewski That worked great! Thanks again for the advice.

bradrydzewski:
Great, glad to hear you got this working.

> It would be incredible if the diff was passed to starlark and we could just implement all this logic ourselves.

This is the end goal.  There are two interesting extension that use diffs in slightly different ways, both of which have merit [1][2].  I think at this point we are waiting to see which extension emerges as the community standard, so that we can adopt a similar approach natively in Drone.

[1] https://github.com/bitsbeats/drone-tree-config
[2] https://github.com/meltwater/drone-convert-pathschanged

wez:
I'm just discovering drone now (perhaps rather late to the party?), and the ability to use starlark (done!) and also filter out work based on changed paths is top of my list.

@tychedelia is there any chance that your combined plugin is still around somewhere and can be shared?

tychedelia:
> @tychedelia is there any chance that your combined plugin is still around somewhere and can be shared?

Unfortunately not. This was at a previous gig and we were packaging Drone ourselves, which is I think how we included our custom plugin. Highly recommend Drone though! I miss using it regularly. 

P.S. thanks for wezterm ✨

wez:
Thanks for responding!  It turns out that you can achieve this now by setting another env var on the drone server to allow mixing multiple plugins. I commented over on https://github.com/harness/drone/pull/2994#issuecomment-1722669455 

> P.S. thanks for wezterm ✨

❤️

PR #3074, opened 2021-03-02
The 'is failing' and 'is passing' are past tense statuses

gaima8:
StatusFailing and StatusPassing are sent after a build has happened.


tphoney:
Hi @gaima8 first of all thank you for the pr, i really appreciate the time that you put into this. Unfortunately we are not going to accept this submission, mostly because it could affect existing behaviour.  In more detail, the terminology used for build messages is 'standard' across most CI providers. Secondly and more importantly, changing messages like this can affect other automation that relies on these messages. 
Thanks again for your work.

PR #3076, opened 2021-03-18
Ignore skip trigger for promote events

username72764274:
In the current implementation, a build can not be promoted if its
message contains the skip ci directive. According to the drone
documentation, the skip directive should be ignored for promotion
events. We added a missing switch case to the skipMessage function
so that promotion events are not skipped.


tphoney:
Hey @username72764274  cool user name BTW 
Functionality like this has already been merged here https://github.com/drone/drone/commit/e3f09c867dfbc3eaaa8fa64bdef41c092be11222

Thanks for the PR and your time. Much appreciated

PR #3077, opened 2021-04-30
fixing a small typo in the compose readme. 

eoinmcafee00:
Fixing typo in docker/compose/README.md. Client secret was missing, client Id repeated.
Also updated the gitignore file to include Jetbrain & Mac specific folders/files.


tboerger:
The gitignore looks pretty much insane. Imho this should only be added to your global gitignore and not being part of the project ignore.

PR #3079, opened 2021-05-15
Fix Gitea branch builds

Vici37:
Drone prefixes a branch build ref with `ref/heads/`, which results in
go-scm constructing a URL to Gitea's commits API as
`.../commits/ref%2Fheads%2fmy-branch`, which gets a 404 response from
Gitea. Removing the prefix and only using `my-branch` for the full ref
now correctly has drone getting the commit details for the branch and
builds are built successfully.

Tested with Gitea 1.13.6 (released March 24, 2021).

### DIFF
--- a/service/commit/commit.go
+++ b/service/commit/commit.go
@@ -78,7 +78,7 @@ func (s *service) FindRef(ctx context.Context, user *core.User, repo, ref string
 	})
 
 	switch s.client.Driver {
-	case scm.DriverBitbucket:
+	case scm.DriverBitbucket, scm.DriverGitea:
 		ref = scm.TrimRef(ref)
 		branch, _, err := s.client.Git.FindBranch(ctx, repo, ref) // wont work for a Tag
 		if err != nil {


bradrydzewski:
👋  hey there, thanks for sending a patch.  I believe this change should be made in the go-scm library.  We already do something similar for the FindTag function in Gitea.  See https://github.com/drone/go-scm/blob/master/scm/driver/gitea/git.go#L39

Vici37:
I can also see this change being put directly into go-scm's gitea driver, since it looks like there's prior art for it to anticipate a `ref/heads/` prefix and to trim it off when [looking for tags](https://github.com/drone/go-scm/blob/master/scm/driver/gitea/git.go#L39).

Vici37:
Ha! Thanks for the response @bradrydzewski ! Yeah, I can transfer this change over to that repo and make a PR there.

Vici37:
Created new [PR on drone/go-scm](https://github.com/drone/go-scm/pull/108/files) with this fix, will close this PR.

PR #3080, opened 2021-05-16
Cleanup unwanted pkgs from go.mod/go.sum

guessi:
It's simply execute `go mod tidy`.


tphoney:
Hey @guessi thanks for the contribution, however we only make changes to the go.mod file when we have to eg security fix / need a new version of library for a feature. This should not affect your ability to work on drone/drone locally. 

Thanks again.

PR #3081, opened 2021-05-20
Feature/template config

eoinmcafee00:
**_What has been done:_**
- Database created for templates
- Store created for querying db
- Core types created
- API for CRUD operations
- unit tests 
- parser for starlark
- converter for starlark
- tech qa comments


store/shared/migrate/mysql/ddl_gen.go
```diff
@@ -658,3 +662,18 @@ CREATE TABLE IF NOT EXISTS latest (
 var createIndexLatestRepo = `
 CREATE INDEX ix_latest_repo ON latest (latest_repo_id);
 `
+
+//
+// 015_create_table_template.sql
+//
+
+var createTableTemplate = `
+CREATE TABLE IF NOT EXISTS template (
```
marko-gacesa, line 671: It seams the table names are names in plural.

```suggestion
CREATE TABLE IF NOT EXISTS templates (
```

handler/api/template/create.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+type templateInput struct {
+	Name string `json:"name"`
+	Data []byte `json:"data"`
+}
+
+// HandleCreate returns an http.HandlerFunc that processes http
+// requests to create a new template.
+func HandleCreate(secrets core.TemplateStore) http.HandlerFunc {
```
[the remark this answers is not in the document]
eoinmcafee00: great catch @marko-gacesa

bradrydzewski:
@eoinmcafee00 let's group import names in all files [1]

example before:

```
import (
	"bytes"
	"context"
	"encoding/json"
	"github.com/drone/drone/core"
	"github.com/drone/drone/handler/api/errors"
	"github.com/drone/drone/mock"
	"github.com/go-chi/chi"
	"github.com/golang/mock/gomock"
	"github.com/google/go-cmp/cmp"
	"net/http"
	"net/http/httptest"
	"testing"
)
```

after:

```
import (
	"bytes"
	"context"
	"encoding/json"
	"net/http"
	"net/http/httptest"
	"testing"

	"github.com/drone/drone/core"
	"github.com/drone/drone/handler/api/errors"
	"github.com/drone/drone/mock"

	"github.com/go-chi/chi"
	"github.com/golang/mock/gomock"
	"github.com/google/go-cmp/cmp"
)
```

[1] https://github.com/golang/go/wiki/CodeReviewComments#imports

plugin/converter/parser/starlark_test.go
```diff
@@ -0,0 +1,92 @@
+package parser
```
bradrydzewski, line 1: missing license header in file

plugin/converter/parser/starlark.go
```diff
@@ -1,26 +1,9 @@
-// Copyright 2019 Drone IO, Inc.
```
bradrydzewski, line 1: revert license header in file

bradrydzewski:
@eoinmcafee00 looks great. I do have a few minor comments with regards to database.  First is that we need to create a unique index on the template table that includes namespace and name fields;  this prevents a user from creating two templates with the same name, for the same namespace / org.  The second is that we should create a standard index on the template namespace field;  generally speaking we always want to index fields (or groups of fields) that are part of our select statement.

plugin/converter/starlark.go
```diff
@@ -0,0 +1,56 @@
...
+	// skip this plugin by returning zero values.
+	switch {
+	case strings.HasSuffix(req.Repo.Config, ".script"):
+	case strings.HasSuffix(req.Repo.Config, ".star"):
+	case strings.HasSuffix(req.Repo.Config, ".starlark"):
+	default:
+		return nil, nil
+	}
+
+	file, _ := starlark.Parse(req, nil, nil)
```
bradrydzewski, line 52: I don't think we want to ignore the error here.  If parsing the Starlark file fails we want the error to be surfaced to the user.

plugin/converter/starlark/starlark_test.go
```diff
@@ -15,17 +15,23 @@
 package starlark
 
 import (
-	"context"
+	"github.com/drone/drone/core"
```
bradrydzewski, line 18: package import grouping

plugin/converter/template.go
```diff
@@ -0,0 +1,85 @@
...
+	if strings.HasSuffix(templateArgs.Load, ".script") ||
+		strings.HasSuffix(templateArgs.Load, ".star") ||
+		strings.HasSuffix(templateArgs.Load, ".starlark") {
+
+		file, err := starlark.Parse(req, template, templateArgs.Data)
+		if err != nil {
+			return nil, err
+		}
+		return &core.Config{
+			Data: *file,
```
bradrydzewski, line 81: change from `*file` to `file`

plugin/converter/starlark/starlark.go
```diff
@@ -146,12 +128,10 @@ func (p *starlarkPlugin) Convert(ctx context.Context, req *core.ConvertArgs) (*c
 	// this is a temporary workaround until we
 	// implement a LimitWriter.
 	if b := buf.Bytes(); len(b) > limit {
-		return nil, ErrMaximumSize
+		return nil, nil
```
bradrydzewski, line 131: why was the return `ErrMaximumSize` removed?

plugin/converter/starlark/starlark.go
```diff
@@ -55,33 +53,7 @@ var (
...
-	// skip this plugin by returning zero values.
-	switch {
-	case strings.HasSuffix(req.Repo.Config, ".script"):
-	case strings.HasSuffix(req.Repo.Config, ".star"):
-	case strings.HasSuffix(req.Repo.Config, ".starlark"):
-	default:
-		return nil, nil
-	}
-
+func Parse(req *core.ConvertArgs, template *core.Template, templateData map[string]interface{}) (file *string, err error) {
```
bradrydzewski, line 56: returning a pointer to a string in this scenario is uncommon.  I would change from `*string` to `string`.  Also I am not sure we need to name the return variables since they are not being referenced anywhere.  I would change this to the following:

```diff
-Parse(req *core.ConvertArgs, template *core.Template, templateData map[string]interface{}) (file *string, err error)
+Parse(req *core.ConvertArgs, template *core.Template, templateData map[string]interface{}) (string, error)
```

plugin/converter/starlark/starlark.go
```diff
@@ -146,12 +128,10 @@ func (p *starlarkPlugin) Convert(ctx context.Context, req *core.ConvertArgs) (*c
...
 	if b := buf.Bytes(); len(b) > limit {
-		return nil, ErrMaximumSize
+		return nil, nil
 	}
-
-	return &core.Config{
-		Data: buf.String(),
-	}, nil
+	parsedFile := buf.String()
+	return &parsedFile, nil
```
bradrydzewski, line 134: this can be simplified to:
```
	return buf.String(), nil
```

plugin/converter/starlark.go
```diff
@@ -0,0 +1,56 @@
...
+	return &core.Config{
+		Data: *file,
```
bradrydzewski, line 54: remove dereference

plugin/converter/starlark/starlark.go
```diff
@@ -16,11 +16,9 @@ package starlark
...
 import (
 	"bytes"
-	"context"
-	"errors"
-	"strings"
 
 	"github.com/drone/drone/core"
+	"github.com/drone/drone/handler/api/errors"
 	"github.com/sirupsen/logrus"
 	"go.starlark.net/starlark"
```
bradrydzewski, line 23: package grouping (github.com/drone should be separate from third party packages)

plugin/converter/starlark.go
eoinmcafee00: My bad, i'd taken that part out whilst testing and forgot to add it back. Fixed now.

plugin/converter/starlark/starlark_test.go
eoinmcafee00: Done.

plugin/converter/template.go
eoinmcafee00: Updated.

plugin/converter/starlark/starlark.go
eoinmcafee00: Not sure how i missed that, fixed.

plugin/converter/starlark/starlark.go
eoinmcafee00: Good to know. Updated.

plugin/converter/starlark.go
eoinmcafee00: Done

PR #3085, opened 2021-06-03
add jsonnet support for build templates

eoinmcafee00:


plugin/converter/jsonnet/jsonnet.go
```diff
@@ -0,0 +1,57 @@
...
+		jsonentFileName = template.Name
+	} else {
+		jsonnetFile = req.Config.Data
+		jsonentFileName = req.Repo.Config
+	}
+	// map external inputs
+	if len(templateData) != 0 {
+		for k, v := range templateData {
+			if s, ok := v.(string); ok {
+				key := "input." + k
```
bradrydzewski, line 31: for now, you could probably do this:

```
for k, v := range templateData {
  key := fmt.Sprintf("input." + k)
  val := fmt.Sprint(v)
  vm.ExtVar(key, val)
}
```

the `fmt.Sprint` command will handle conversion of all types to a string

plugin/converter/template.go
```diff
@@ -19,6 +19,7 @@ import (
 	"errors"
 
 	"github.com/drone/drone/core"
+	"github.com/drone/drone/plugin/converter/jsonnet"
 	"github.com/drone/drone/plugin/converter/starlark"
 
 	"regexp"
```
bradrydzewski, line 25: one minor edit:

```diff
import (
  "context"
  "errors"
+ "regexp"
+ "strings"

  "github.com/drone/drone/core"
  "github.com/drone/drone/plugin/converter/jsonnet"
  "github.com/drone/drone/plugin/converter/starlark"

- "regexp"
- "strings"

  "gopkg.in/yaml.v2"
)
```

plugin/converter/jsonnet/jsonnet.go
eoinmcafee00: Good to know, updated.

PR #3094, opened 2021-06-16
Added "DEBIAN_FRONTEND=noninteractive" to list of default environment variables.

hwittenborn:
All Drone CI builds are run in an automated manner that doesn't accept user input.

Therefore, it would be desirable if, when installing packages with `apt`, that it *defaults* to running without user prompts.

This can currently be done if the flag is manually set in a build file, but I don't see any reason why it shouldn't be a global default (besides maybe being platform-specific to Debian systems, but Debian's quite widely used).

### DIFF
--- a/operator/runner/env.go
+++ b/operator/runner/env.go
@@ -30,6 +30,7 @@ func systemEnviron(system *core.System) map[string]string {
 		"DRONE_SYSTEM_HOST":     system.Host,
 		"DRONE_SYSTEM_HOSTNAME": system.Host,
 		"DRONE_SYSTEM_VERSION":  fmt.Sprint(system.Version),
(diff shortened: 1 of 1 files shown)


tphoney:
Hey @hwittenborn, many thanks for your time and effort here. Unfortunately we will not be merging this PR.
The primary concern with injecting this environment variable into all pipelines automatically is that it could unexpectedly break existing pipelines.  
This can be set globally using instead of hard-coding non-in
https://docs.drone.io/runner/docker/configuration/reference/drone-runner-environ/

hwittenborn:
No problem. The Drone environment extension seemed to serve my purpose just fine as well.

tphoney:
glad it worked out. 🎉

PR #3098, opened 2021-06-28
fix(bug): build failed with tag "oss"

guessi:
Tried to build with v2.0.2, and quickly found out it's unable to build when tag `oss` is set.

```
cmd/drone-server/inject_store.go:63:19: too many arguments in call to db.Connect

    have (string, string, int)
    want (string, string)
```

introduced by https://github.com/drone/drone/pull/3089

### DIFF
--- a/store/shared/db/conn_oss.go
+++ b/store/shared/db/conn_oss.go
@@ -26,11 +26,14 @@ import (
 )
 
 // Connect to an embedded sqlite database.
-func Connect(driver, datasource string) (*DB, error) {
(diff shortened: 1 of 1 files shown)


guessi:
another issue is that `/version` not reflect it's actual version

since the `version/version.go` not updated
https://github.com/drone/drone/blob/v2.0.2/version/version.go#L29-L30

eoinmcafee00:
Thank you @guessi . Merging now.

PR #3099, opened 2021-07-01
Add middleware function redirect slashes to fix trailing slash in url.

eoinmcafee00:


tphoney (approved):
the behaviour works. 
I am not an expert on whether this is the best option though.

PR #3103, opened 2021-07-13
Allow filtering builds by event

julienduchesne:
A user can now pass `&event=cron|push|pull_request|...`

The WHERE condition is built like `('' = :build_event OR build_event = :build_event)` to allow the same query to be used when an event is specified and when it isn't (empty string)
As such, the `List` and `ListRef` methods were also merged into one using the same pattern

This is done to support https://github.com/drone/drone-ui/pull/356. I know that the support of filtering was riding on the implementation so this PR can serve as a discussion base
Let me know what you think of this

### DIFF
--- a/core/build.go
+++ b/core/build.go
@@ -67,10 +67,7 @@ type BuildStore interface {
 	FindRef(context.Context, int64, string) (*Build, error)
 
 	// List returns a list of builds from the datastore by repository id.
-	List(context.Context, int64, int, int) ([]*Build, error)
-
-	// ListRef returns a list of builds from the datastore by ref.
-	ListRef(context.Context, int64, string, int, int) ([]*Build, error)
+	List(context.Context, int64, string, string, int, int) ([]*Build, error)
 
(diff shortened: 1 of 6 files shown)


bradrydzewski:
I recommend using a query builder here instead:

```Go
var buf strings.Builder
buf.WriteString(queryBase)
buf.Writestring("FROM builds WHERE build_repo_id = :build_repo_id")
if params.Event != "" {
  buf.WriteString("AND build_ref = :build_ref")
}
buf.WriteString(`ORDER BY build_id DESC LIMIT :limit OFFSET :offset`)
```

And modifying the signature so we have more flexibility in the future to add more filtering criteria:

```Go
func (s *buildStore) List(ctx context.Context, repo int64, opts core.BuildListOpts) ([]*core.Build, error)
```

marko-gacesa:
Hey @julienduchesne, thanks for your contribution. I'm going to work on the changes that @bradrydzewski suggested. I'm going to build them on top of your commit. When I'm done I'll create a new pull request.

PR #3105, opened 2021-07-15
Allow jsonnet imports in pipeline configuration

hhamalai:
Having large projects using e.g., mono repo managing pipeline configurations becomes quickly quite cumbersome and error-prone. Having the ability to split the single pipeline configuration into multiple jsonnet files makes it a bit easier to manage.

Implementation in this pull request is based on old drone/drone-jsonnet-config with updates.

I've tested this to work with gitea. I've an open question still which token should be used in the scm Find call, now it's using user token, but this likely should be a token bound to drone SCM user itself. But is there any preferable way to retrieve that?

### DIFF
--- a/plugin/converter/jsonnet.go
+++ b/plugin/converter/jsonnet.go
@@ -19,14 +19,18 @@ import (
 
 // Jsonnet returns a conversion service that converts the
 // jsonnet file to a yaml file.
-func Jsonnet(enabled bool) core.ConvertService {
+func Jsonnet(enabled bool, limit int, fileService core.FileService) core.ConvertService {
 	return &jsonnetPlugin{
-		enabled: enabled,
+		enabled:     enabled,
+		limit:       limit,
+		fileService: fileService,
 	}
 }
 
 type jsonnetPlugin struct {
-	enabled bool
+	enabled     bool
+	limit       int
+	fileService core.FileService
 }
 
 func (p *jsonnetPlugin) Convert(ctx context.Context, req *core.ConvertArgs) (*core.Config, error) {
@@ -40,7 +44,7 @@ func (p *jsonnetPlugin) Convert(ctx context.Context, req *core.ConvertArgs) (*co
 		return nil, nil
 	}
 
-	file, err := jsonnet.Parse(req, nil, nil)
+	file, err := jsonnet.Parse(req, p.fileService, p.limit, nil, nil)
 
 	if err != nil {
 		return nil, err
--- a/plugin/converter/jsonnet/jsonnet.go
+++ b/plugin/converter/jsonnet/jsonnet.go
@@ -2,10 +2,14 @@ package jsonnet
 
 import (
 	"bytes"
+	"context"
 	"fmt"
+	"path"
 	"strconv"
+	"strings"
 
 	"github.com/drone/drone/core"
+	"github.com/drone/drone/handler/api/errors"
 
 	"github.com/google/go-jsonnet"
 )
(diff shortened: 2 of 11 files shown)


tphoney:
Could you explain how this would work for an end user, to give some added context.

hhamalai:
Sure, for an end-user, the pipeline configuration would look like the following example https://gist.github.com/hhamalai/506b2d0f533957eff4e7324b4ba52ecd

plugin/converter/jsonnet.go
```diff
@@ -8,6 +8,7 @@ package converter
 
 import (
 	"context"
+	"github.com/drone/go-scm/scm"
```
bradrydzewski, line 11: import grouping

plugin/converter/jsonnet/jsonnet.go
```diff
@@ -2,8 +2,14 @@ package jsonnet
 
 import (
 	"bytes"
+	"context"
 	"fmt"
+	"github.com/drone/drone/handler/api/errors"
```
bradrydzewski, line 7: group imports by stdlib and then internal dependencies, and then external

mavimo:
@hhamalai I'm not sure but if you use the drone-converter-plugin for jsonnet already work as you report, we use it in a multi-file configuration.

plugin/converter/jsonnet.go
```diff
@@ -19,14 +20,16 @@ import (
 
 // Jsonnet returns a conversion service that converts the
 // jsonnet file to a yaml file.
-func Jsonnet(enabled bool) core.ConvertService {
+func Jsonnet(enabled bool, client *scm.Client) core.ConvertService {
```
eoinmcafee00, line 23: We already have a service: https://github.com/drone/drone/blob/master/service/content/content.go#L49 that can be used rather than injecting the scm.Client directly and does all the heavy lifting for you.

plugin/converter/jsonnet/jsonnet.go
```diff
@@ -14,11 +20,93 @@ const repo = "repo."
...
+func Parse(req *core.ConvertArgs, client *scm.Client, template *core.Template, templateData map[string]interface{}) (string, error) {
 	vm := jsonnet.MakeVM()
 	vm.MaxStack = 500
 	vm.StringOutput = false
 	vm.ErrorFormatter.SetMaxStackTraceSize(20)
+	vm.Importer(
+		&importer{
+			repo:   req.Repo,
+			build:  req.Build,
+			limit:  10,
```
eoinmcafee00, line 105: This needs to be 0 by default unless the user specifies. Otherwise we could run into request limit issues with the users provider.

hhamalai: Added this an environment variable option, with default value 0

plugin/converter/jsonnet.go
hhamalai: Thanks, changed to use FileService which also closed the open question about authentication tokens.

plugin/converter/jsonnet/jsonnet.go
eoinmcafee00: Awesome thanks.

eoinmcafee00:
@hhamalai 
Thank you for making those changes so quickly. Can I ask one more thing, can you revert to go.mod & go.sum files, please.
Typically we don't change these unless we have a purpose for the change. Unless you have a reason for doing so, then it should just include that change(s)

hhamalai:
> @hhamalai
> Thank you for making those changes so quickly. Can I ask one more thing, can you revert to go.mod & go.sum files, please.
> Typically we don't change these unless we have a purpose for the change. Unless you have a reason for doing so, then it should just include that change(s)

Right, these were updated accidentally by my IDE, not intentionally. I reverted changes to go dependency files and checked that the jsonnet imports work as expected.

eoinmcafee00 (approved):
Approved. Once @bradrydzewski has given his approval I'll merge. Thanks.

plugin/converter/jsonnet.go
hhamalai: These should Be formatted now

plugin/converter/jsonnet/jsonnet.go
```diff
@@ -14,11 +18,89 @@ const repo = "repo."
...
+
+	// jsonnet does not cache file imports and may request
+	// the same file multiple times. We cache the files to
+	// duplicate API calls.
+	cache map[string]jsonnet.Contents
+
+	// limit the number of outbound requests. github limits
+	// the number of api requests per hour, so we should
+	// make sure that a single build does not abuse the api
+	// by imporing dozens of files.
```
james-callahan, line 35:
```suggestion
	// by importing dozens of files.
```

james-callahan (changes requested):

hhamalai:
@bradrydzewski This PR is waiting for your approval, I'm glad to discuss about these changes and provide further improvements if needed.

Cheers,

timorantalaiho:
> @bradrydzewski This PR is waiting for your approval, I'm glad to discuss about these changes and provide further improvements if needed.

That would be great! I would also benefit from getting these changes to mainline.

eoinmcafee00:
@hhamalai 
Can you resolve the conflicts, please?

hhamalai:
@eoinmcafee00 Thanks for the heads up, rebased against latest master.

bradrydzewski:
I feel like we should only load the vm importer if the `fileService` is not nil.  The main concern that I have is that `fileService` is always going to be `nil` when invoked from the template package.  If a template defines an import, the `vm.Importer` code will be used and could panic because the `fileService` is nil.

```diff
+if fileService != nil {
	vm.Importer(
		&importer{
			repo:        req.Repo,
			build:       req.Build,
			limit:       limit,
			user:        req.User,
			fileService: fileService,
		},
	)
+}
```

hhamalai:
Added tests for these cases when limit is 0 and when fileService is nil, both expecting failure when jsonnet imports are used, and success when no jsonnet imports are used.

eoinmcafee00 (approved):

pinpox:
My pipeline failes when trying to import a json file in the root of the repository.

```jsonnet
local info = import 'info.json';
local hosts = std.objectFields(info.nixosConfigurations);

local steps_hosts() = [
  {
    name: 'Build host: %s' % host,
    commands: [
      "nix build -v -L '.#nixosConfigurations.%s.config.system.build.toplevel'" % host,
    ],
  }
  for host in hosts
];

```

Is there something I have to manually enable for this to work with drone?

This is what I see in the web UI of drone:
```
RUNTIME ERROR: couldn't open import "info.json": no match locally or in the Jsonnet library paths .drone.jsonnet:7:14-32 thunk <info> from <$> .drone.jsonnet:17:34-38 <std>:1293:24-25 thunk from <function <anonymous>> <std>:1293:5-33 function <anonymous> .drone.jsonnet:17:17-59 function <steps_hosts> .drone.jsonnet:(10:3)-(18:4) .drone.jsonnet:59:7-20 object <anonymous> During manifestation
```

tiagosilva-hotmart:
I'm having the same problem as @pinpox. @hhamalai can you help us?

hhamalai:
Hi,

To make sure, you've set the following environment variables for your deployment:
```
            - name: DRONE_JSONNET_IMPORT_LIMIT
              value: "5"
            - name: DRONE_JSONNET_ENABLED
              value: "true"
```

Which drone release version are you using? Also, what's the source code management backend you are using?

pinpox:
I had only set `DRONE_JSONNET_ENABLED` because `DRONE_JSONNET_IMPORT_LIMIT` is not documented anywhere. Adding the import limit seems to have solved the problem for me. 

Maybe `DRONE_JSONNET_IMPORT_LIMIT`  should be added to the [reference](https://docs.drone.io/pipeline/environment/reference/ )? Correct me if I overlooked the documentation for it, but I couldn't find it.

hhamalai:
Thanks for confirming the working setup. The documentation and the error message require a bit of improvement there. The latter might be trickier to achieve due to the JSONnet library used.

eoinmcafee00:
Hey @pinpox 
If you've noticed something missing from the docs, please feel free to raise a PR https://github.com/drone/docs to add it & I'll take look / get it merged.
Cheers,
Eoin

pinpox:
@eoinmcafee00 I'd do that if I know what the variable exactly does. Will it just set the amount of files that can be imported? Should this be documented on the jsonnet page of the docs aswell?

PR #3106, opened 2021-07-20
remove parent promote build params

zc2638:
When I try to build from historical promote, I always inherit parameters.
This makes the build a lot of mistakes.
A promote may have many parameters, and only a part of them will be used most of the time.I think reentering is more comfortable than inheriting.

### DIFF
--- a/handler/api/repos/builds/promote.go
+++ b/handler/api/repos/builds/promote.go
@@ -76,10 +76,6 @@ func HandlePromote(
 			Params:       map[string]string{},
 		}
 
-		for k, v := range prev.Params {
-			hook.Params[k] = v
-		}
-
 		for key, value := range r.URL.Query() {
 			if key == "access_token" {
 				continue


tphoney:
Hey @zc2638 thanks for the PR and this would solve your problem, but it would change the current behaviour. Breaking some users current working patterns. I wrote a proposal to better track a more full solution https://github.com/drone/proposal/issues/8 
does that align with what you were thinking ?

zc2638:
No problem, this way can also solve my problem. @tphoney

zc2638:
I think it's more appropriate in `drone.yaml` to allow build promotion to ignore existing parameters via a configuration item.
What's your opinion? @tphoney

bradrydzewski:
> I think it's more appropriate in drone.yaml to allow build promotion to ignore existing parameters via a configuration item.
What's your opinion? @tphoney

when you promote a pipeline the user-defined parameters are inherited before the yaml is parsed, which means defining this behavior in the yaml would not be possible.

zc2638:
> > I think it's more appropriate in drone.yaml to allow build promotion to ignore existing parameters via a configuration item.
> > What's your opinion? @tphoney
> 
> When you promote a pipeline the user-defined parameters are inherited before the yaml is parsed, which means defining this behavior in the yaml would not be possible.
> 
> Perhaps another approach could be to only inherit the parameters of the parent build if no additional user-defined parameters are provided in the API request (see below). However, this could still be problematic if there are users that have come to rely on the current behavior, which is very possible.
> 
> ```diff
> -		for k, v := range prev.Params {
> -			hook.Params[k] = v
> -		}
> 
> 		for key, value := range r.URL.Query() {
> 			if key == "access_token" {
> 				continue
> 			}
> 			if key == "target" {
> 				continue
> 			}
> 			if len(value) == 0 {
> 				continue
> 			}
> 			hook.Params[key] = value[0]
> 		}
> 
> 
> +		if len(hook.Params) == 0 {
> +			for k, v := range prev.Params {
> +				hook.Params[k] = v
> +			}
> +		}
> ```

In your way, it's no different from `restart` when you don't enter parameters.

I think most of the time, `promote` works in one way.Is it feasible to configure in repo settings?

tphoney:
from the proposal, `Drone ui changes - allow build promotion to edit / remove existing build parameters, and add new build parameters in the dialog box` 
image

PR #3108, opened 2021-07-22
pubsub: redis implementation

marko-gacesa:


cmd/drone-server/inject_store.go
```diff
@@ -55,6 +56,7 @@ var storeSet = wire.NewSet(
 	global.New,
 	step.New,
 	template.New,
+	provideRedisClient,
```
eoinmcafee00, line 59: Is this the right place for this? We on load stores here. Feels like it would belong more in inject_service?

eoinmcafee00:
LGTM, I'd maybe also add to the description what this change solves and why?

cmd/drone-server/inject_store.go
marko-gacesa: I'm not sure either. Brad suggested to add it to inject_store.go. But in reality it's in its own category. Maybe it would be the best to create a new file.

eoinmcafee00: Yeah I'd agree with you there.

marko-gacesa: changed... moved to a new file inject_external.go

eoinmcafee00 (approved):

cmd/drone-server/config/config.go
```diff
@@ -166,6 +167,11 @@ type (
 		EnableAnonymousAccess bool `envconfig:"DRONE_PROMETHEUS_ANONYMOUS_ACCESS" default:"false"`
 	}
 
+	// Redis provides the redis configuration.
+	Redis struct {
+		ConnectionString string `envconfig:"DRONE_REDIS_CONNECTION"`
```
bradrydzewski, line 172: I recommend we also include redis username and password as optional parameters.  If someone is using hosted redis (such as aws redis) they will need to provide the username and password to connect.

PR #3110, opened 2021-07-29
Add support for nested data objects within templates

eoinmcafee00:
Initial template design didn't support nested data values such as:

`kind: template load: test.jsonnet data: builds: - name: build1 - name: build2
`
This PR adds this support.
Ticket was raised at: https://discourse.drone.io/t/drone-not-parsing-template-data-correctly/9531

### DIFF
--- a/plugin/converter/starlark/args.go
+++ b/plugin/converter/starlark/args.go
@@ -16,10 +16,9 @@ package starlark
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/drone/drone/core"
-
-	"reflect"
 
 	"go.starlark.net/starlark"
 	"go.starlark.net/starlarkstruct"
--- a/plugin/converter/starlark/args.go
+++ b/plugin/converter/starlark/args.go
@@ -15,50 +15,98 @@
 package starlark
 
 import (
+	"fmt"
+
 	"github.com/drone/drone/core"
+
+	"reflect"
 
 	"go.starlark.net/starlark"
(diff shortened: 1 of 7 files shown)


plugin/converter/starlark/args.go
```diff
@@ -15,50 +15,98 @@
...
-
-// TODO(bradrydzewski) add build id, will always be zero value
-// TODO(bradrydzewski) add build number, will always be zero value
-// TODO(bradrydzewski) add build started, will always be zero value
-// TODO(bradrydzewski) add build finished, will always be zero value
-// TODO(bradrydzewski) add build created, will always be zero value
-// TODO(bradrydzewski) add build updated, will always be zero value
-// TODO(bradrydzewski) add build parent
-// TODO(bradrydzewski) add build timestamp
+type GoSlice struct {
```
marko-gacesa, line 28: what is this for?

eoinmcafee00: Good catch Marko, i forgot to delete this.

plugin/converter/starlark/args.go
```diff
@@ -15,8 +15,12 @@
 package starlark
 
 import (
+	"fmt"
+
 	"github.com/drone/drone/core"
 
+	"reflect"
```
marko-gacesa, line 22: order the imports, move this line just below the "fmt" import.

eoinmcafee00: done.

marko-gacesa (approved):

PR #3111, opened 2021-07-31
(maint) Readme update Add Contributor Section 

mrsantons:
Reason for change: 
Update readme to inform new users and contributors of pertinent information. 

Steps to test change: 
None, markdown change only.


.github/readme.md
```diff
@@ -1,11 +1,64 @@
...
+- [Contributing](#contributing)
+- [Code of Conduct](#code-of-conduct)
+- [Core Team](#core-team)
+- [Setup Documentation](#setup-documentation)
+- [Usage Documentation](#usage-documentation)
+- [Sample Pipeline Configuration](#sample-pipeline-configuration)
+- [Plugin Index](#plugin-index)
+
+## Community
+---
```
bradrydzewski, line 24: the `---` can be removed since these are meant to be page breaks.  Also github already adds an underline to `h2` headers via their css style sheet.

.github/readme.md
```diff
@@ -1,11 +1,64 @@
...
+
+## Contributing
+---
+We encourage you to contribute to Drone! Please check out our [Proposing Changes Guide](https://github.com/drone/proposal).
+
+## Code of Conduct
+---
+Drone follows the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md).
+
+## Core Team
```
bradrydzewski, line 36: many projects track this data, however, it is common practice to store in a separate file.  Perhaps we do something similar?  here are some examples:

<https://github.com/golang/go/blob/master/CONTRIBUTORS> <https://github.com/goharbor/community/blob/master/MAINTAINERS.md>

mrsantons: Yeah could do something like this. Where would I put this file? We can always discuss Monday at standup.

marko-gacesa (approved):

.github/readme.md
mrsantons: @bradrydzewski Could we discuss this Monday at standup so we can get this in?

tphoney (approved):

PR #3116, opened 2021-08-15
change work channel buffer

zzhzero:
fix work channel block
 If ctx done


eoinmcafee00:
Hey @zzhzero 
Thanks for the pull request. However, I believe this would introduce a breaking change to the platform, unless you can provide more of an explanation on why this change was necessary and some unit tests. I will close for now, but feel free to raise again if you want to do the above.
Cheers,
Eoin

PR #3118, opened 2021-08-17
update dependency drone/go-scm to 1.15.2 to fix gitea build problem

sesky4:
related:
#3079
#2973


eoinmcafee00:
Thanks @sesky4 
Merged.

PR #3125, opened 2021-08-26
event-stream supports timeout

zc2638:
```golang
for {
    select {
        case <-time.After(time.Hour):
            break T
        case <-time.After(pingInterval):
    }
}
```
As the method shows, ` case <-time.After(pingInterval)` Will invalidate the `timeout definition`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ release/
 scripts/*.go
 docker/**/data
 TODO*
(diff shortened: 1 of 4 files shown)


marko-gacesa (approved):

marko-gacesa:
@zc2638 thanks for this!

PR #3128, opened 2021-08-31
add check on template extension type - throw error if invalid

eoinmcafee00:
An issue has occurred where users have saved unsupported template file types.
Therefore I have added a check when a user uploads a template file, it has to be yaml, starlark or jsonnet.

### DIFF
--- a/plugin/converter/template.go
+++ b/plugin/converter/template.go
@@ -39,6 +39,7 @@ var (
 	TemplateFileRE          = regexp.MustCompile("^kind:\\s+template+\\n")
 	ErrTemplateNotFound     = errors.New("template converter: template name given not found")
 	ErrTemplateSyntaxErrors = errors.New("template converter: there is a problem with the yaml file provided")
+	errTemplateExtensionInvalid = errors.New("template extension invalid. must be yaml, starlark or jsonnet")
 )
 
 func Template(templateStore core.TemplateStore) core.ConvertService {
@@ -84,9 +85,8 @@ func (p *templatePlugin) Convert(ctx context.Context, req *core.ConvertArgs) (*c
 	case ".jsonnet":
 		return parseJsonnet(req, template, templateArgs)
 	default:
+		return nil, errTemplateExtensionInvalid
 	}
-
-	return nil, nil
 }
 
 func parseYaml(req *core.ConvertArgs, template *core.Template, templateArgs core.TemplateArgs) (*core.Config, error) {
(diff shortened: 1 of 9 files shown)


plugin/converter/testdata/yaml.template.invalid.yml
```diff
@@ -0,0 +1,6 @@
+kind: template
```
tphoney, line 1: should this not be a txt file ?

tphoney (changes requested):

plugin/converter/testdata/yaml.template.invalid.yml
eoinmcafee00: It doesn't really matter as i'm testing the converter logic at this stage, the file extension name check would happen in the handler. I only care about testing the bit of logic https://github.com/drone/drone/pull/3128/files#diff-53a3b4e02181d9a28742c3e46d4120505787eb61138ce0fade533705d460d0deR88 and with the change I made in the handler, should never get to the point where the db has an invalid template

plugin/converter/template.go
```diff
@@ -39,6 +39,7 @@ var (
 	TemplateFileRE          = regexp.MustCompile("^kind:\\s+template+\\n")
 	ErrTemplateNotFound     = errors.New("template converter: template name given not found")
 	ErrTemplateSyntaxErrors = errors.New("template converter: there is a problem with the yaml file provided")
+	errTemplateExtensionInvalid = errors.New("template extension invalid. must be yaml, starlark or jsonnet")
```
bradrydzewski, line 42: does this need gofmt ?

eoinmcafee00: ah good catch!

eoinmcafee00: updated.

plugin/converter/template.go
```diff
@@ -36,9 +36,10 @@ import (
 
 var (
 	// TemplateFileRE regex to verifying kind is template.
-	TemplateFileRE          = regexp.MustCompile("^kind:\\s+template+\\n")
-	ErrTemplateNotFound     = errors.New("template converter: template name given not found")
-	ErrTemplateSyntaxErrors = errors.New("template converter: there is a problem with the yaml file provided")
+	TemplateFileRE              = regexp.MustCompile("^kind:\\s+template+\\n")
+	ErrTemplateNotFound         = errors.New("template converter: template name given not found")
+	ErrTemplateSyntaxErrors     = errors.New("template converter: there is a problem with the yaml file provided")
```
tphoney, line 41: are all these errors local, they should start with a lowercase letter

PR #3131, opened 2021-09-01
fix templating reg expression to match if .drone.yml contains --- characters 

eoinmcafee00:
There's been an issue identified with templating. If a user has --- in their drone.yml, the regular expression checker for type template won't match. Therefore this PR updates the expression to fix it.
`regexp.MustCompilePOSIX("^kind:[[:space:]]+template[[:space:]]?\\+?")`

### DIFF
--- a/plugin/converter/template.go
+++ b/plugin/converter/template.go
@@ -36,7 +36,7 @@ import (
 
 var (
 	// templateFileRE regex to verifying kind is template.
-	templateFileRE              = regexp.MustCompilePOSIX("^kind:[[:space:]]+template[[:space:]]?\\+$")
+	templateFileRE              = regexp.MustCompilePOSIX("^kind:[[:space:]]+template[[:space:]]?+$")
 	errTemplateNotFound         = errors.New("template converter: template name given not found")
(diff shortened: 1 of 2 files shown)


plugin/converter/template.go
```diff
@@ -36,7 +36,7 @@ import (
 
 var (
 	// templateFileRE regex to verifying kind is template.
-	templateFileRE              = regexp.MustCompile("^kind:\\s+template+\\n")
+	templateFileRE              = regexp.MustCompilePOSIX("^kind:[[:space:]]+template[[:space:]]?\\+?")
```
tphoney, line 39: are you just checking for the existance for a single line ie
```kind: template``` where there can be any number of spaces between the `kind:` and `template`. Why are you matching characters after the e in template ? should you not be checking for a newline ?

eoinmcafee00: Not fully understanding you here TP...

eoinmcafee00: I've updated to remove the \\+?, might just have been a typo when he sent it across.

bradrydzewski (approved):

marko-gacesa (approved):

PR #3135, opened 2021-09-05
bugfix: drone component failed to run and the process could not be terminated normally

zc2638:
When the `drone` starts, if one of the coroutines fails to execute, it should exit the program.

Because `errgroup` is used, the program will only exit when all the coroutines have ended.

The fastest way to test is to let `drone server` listen an occupied port.

### DIFF
--- a/cmd/drone-server/main.go
+++ b/cmd/drone-server/main.go
@@ -24,8 +24,8 @@ import (
 	"github.com/drone/drone/core"
 	"github.com/drone/drone/metric/sink"
 	"github.com/drone/drone/operator/runner"
-	"github.com/drone/drone/service/canceler/reaper"
 	"github.com/drone/drone/server"
+	"github.com/drone/drone/service/canceler/reaper"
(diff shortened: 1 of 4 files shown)


bradrydzewski:
Instead of removing errorgroup, I feel like we could fix with a small change to use `errgroup.WithContext`

> [`WithContext`](https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext) returns a new Group and an associated Context derived from ctx.
> The derived Context is canceled the first time a function passed to Go returns a non-nil error or the first time Wait returns, whichever occurs first.

zc2638:
> Instead of removing errgroup, I feel like we could fix with a small change to use `errgroup.WithContext`. When the derived context is cancelled it should cancel all remaining subroutines, since all of the subroutines accept context.

I agree with you. I'll revise it.

zc2638:
@bradrydzewski Please take a look.

bot2-harness:

PR #3145, opened 2021-09-21
Add build.number variable to jsonnet and starlark

hjkatz:
# Why?

I feel that this extVar was missed or determined to not be as useful as the others. I use `build.number` in the form of `${DRONE_BUILD_NUMBER}`, but would prefer to use `extVar("build.number")` instead.

# What?

Supply `build.number` as an external variable available to the jsonnet and starlark conversion plugins.

# Docs

See: drone/docs#481


bradrydzewski:
👋  Hey there, thanks for taking the time to submit a pull request.

The starlark and jsonnet files are parsed before [1]  the build is created and the build number is incremented [2], which means due to order of operations, the build number will be zero when the script is evaluated. 

[1] <https://github.com/drone/drone/blob/master/trigger/trigger.go#L206>
[2] <https://github.com/drone/drone/blob/master/trigger/trigger.go#L339>

hjkatz:
Well dang, nvm then :)

PR #3149, opened 2021-09-24
Feature/dron 101 cards

eoinmcafee00:
Dron-101


handler/api/acl/check.go
```diff
@@ -48,6 +51,36 @@ func CheckAdminAccess() func(http.Handler) http.Handler {
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			var (
+				owner = chi.URLParam(r, "owner")
+				name  = chi.URLParam(r, "name")
+			)
+			log := logger.FromRequest(r).
+				WithField("namespace", owner).
+				WithField("name", name)
+
+			secret := []byte(os.Getenv("DRONE_INTERNAL_AUTH_SECRET"))
```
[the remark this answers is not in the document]
bradrydzewski: also add unit tests for this :)

handler/api/api.go
```diff
@@ -285,6 +294,16 @@ func (s Server) Handler() http.Handler {
 					acl.CheckAdminAccess(),
 				).Delete("/{member}", collabs.HandleDelete(s.Users, s.Repos, s.Perms))
 			})
+
+			r.Route("/cards", func(r chi.Router) {
+				r.Get("/{build}", card.HandleFindAll(s.Builds, s.Card, s.Repos))
+				r.Get("/{build}/{stage}/{step}", card.HandleFind(s.Builds, s.Card, s.Stages, s.Steps, s.Repos))
+				r.Get("/{build}/{stage}/{step}/json", card.HandleFindData(s.Builds, s.Card, s.Stages, s.Steps, s.Repos))
+				r.With(
+					acl.AuthorizeAdmin,
```
bradrydzewski, line 303: AuthorizeAdmin should be removed since it only allows a Drone system admin to remove a card.  The CheckAdminAccess will allow anyone with admin access to the repository to delete the card, which should be sufficient.

store/card/card.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		row := queryer.QueryRow(query, args...)
+		return scanRowCardDataOnly(row, out)
+	})
+	return ioutil.NopCloser(
+		bytes.NewBuffer(out.Data),
+	), err
+}
+
+func (c cardStore) CreateCard(ctx context.Context, card *core.CreateCard) error {
```
bradrydzewski, line 84: the `data` should be removed from the `Card` and instead passed in as an `io.Reader` so that we can (eventually) allow uploading and storing in S3
https://gist.github.com/bradrydzewski/0b8b67db06f9fcf8dcb718c6fdf8e658#database-calls

core/card.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type CreateCard struct {
+	Id     int64  `json:"id,omitempty"`
+	Build  int64  `json:"build,omitempty"`
+	Stage  int64  `json:"stage,omitempty"`
+	Step   int64  `json:"step,omitempty"`
+	Schema string `json:"schema,omitempty"`
+	Data   []byte `json:"data,omitempty"`
+}
```
bradrydzewski, line 46: I think we should be able to eliminate the `CreateCard` struct since the `Data` should be passed to `CreateCard` as an `io.Reader`.  

```
CreateCard(ctx context.Context, card *Card, data io.Reader) error
```

handler/api/acl/check.go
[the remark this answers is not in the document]
bradrydzewski: actually, it might be good to move this change to a separate pull request so that we can evaluate and iterate on the approach separately.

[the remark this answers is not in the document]
eoinmcafee00: Done.

store/card/card.go
eoinmcafee00: done

handler/api/api.go
eoinmcafee00: removed

core/card.go
eoinmcafee00: done,

core/card.go
```diff
@@ -0,0 +1,64 @@
...
+
+// CardStore manages repository cards.
+type CardStore interface {
+	FindCardByBuild(ctx context.Context, build int64) ([]*Card, error)
+	FindCard(ctx context.Context, step int64) (*Card, error)
+	FindCardData(ctx context.Context, id int64) (io.ReadCloser, error)
+	CreateCard(ctx context.Context, card *Card, data io.ReadCloser) error
+	DeleteCard(ctx context.Context, id int64) error
+}
```
bradrydzewski, line 50: lets make the naming more consistent with the other stores (no need to include `Card` in the method names)

```
CardStore.Find
CardStore.FindByBuild
CardStore.FindData
CardStore.Create
CardStore.Delete
```

I recommend reading this overview of Go naming conventions, where are a bit different than other languages:
https://medium.com/@lynzt/variable-naming-conventions-in-go-89fe1ef17b0a

PR #3151, opened 2021-09-29
Add drone debug env

rockmenjack:
this pull request adds DRONE_BUILD_DEBUG=true|false to allow conditionally run debug step when using drone-runner-ssh.
for example:

```yml
kind: pipeline
type: ssh
name: default

steps:
- name: debug
  commands:
  - [ "$${DRONE_BUILD_DEBUG}" == "true" ] && exit 0
  - sleep infinity # or run tmate
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [Unreleased](https://github.com/drone/drone/tree/HEAD)
+
+[Full Changelog](https://github.com/drone/drone/compare/v2.4.0...HEAD)
+
+**Implemented enhancements:**
+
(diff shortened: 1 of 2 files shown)


bradrydzewski:
👋 hey, thanks for the pull request, however, this code was deprecated and was moved here:
https://github.com/drone-runners/drone-runner-docker and here:
https://github.com/drone/runner-go

you can see we set the debug flag here:
https://github.com/drone/runner-go/blob/master/environ/environ.go#L146

if you want this functionality in drone-runner-ssh you would need to add to:
https://github.com/drone-runners/drone-runner-ssh

PR #3154, opened 2021-10-13
migrate struct from core -step to drone-go

eoinmcafee00:
Removing Step struct from core and using drone-go version. 
Update drone-go version and dependencies


tphoney:
I wouldnt move the Nullstep struct or IsDone Function, its not about moving them, its about removing duplicated structs. But that is my opinion, worth seeing what @bradrydzewski says.

PR #3159, opened 2021-10-26
refactor create / find / delete end points for cards

eoinmcafee00:
Introduce route to persist card


operator/manager/rpc2/handler.go
```diff
@@ -26,6 +26,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/drone/drone/handler/api/render"
```
marko-gacesa, line 29: Organize imports. Move this line below, where the other drone/drone imports are.

operator/manager/rpc/client.go
```diff
@@ -38,6 +39,10 @@ type Client struct {
 	client *retryablehttp.Client
 }
 
+func (s *Client) UploadCard(ctx context.Context, step int64, input *core.CardInput) error {
```
tphoney, line 42: this should have a test similar to the other functions in this package.

tphoney (changes requested):
squash the commits and have a look at the changes in go.mod

operator/manager/rpc2/handler.go
eoinmcafee00: sorted

operator/manager/rpc/client.go
eoinmcafee00: To write tests similar to the ones in the package, id have to write code to hook it up in server.go which is kinda pointless dead code?

tphoney (approved):

PR #3162, opened 2021-11-23
Fix the bug of mapping the target field of DeployHook

noahingh:
The Drone has a bug that changes the environment of GitHub deployment after a build when the trigger is promoted. It looks like the Drone maps the `target` field with the wrong field.


hanjunlee:
@d1wilko Could check this PR? Thanks

bot2-harness:

PR #3171, opened 2021-12-17
release prep v2.7.1

eoinmcafee00:


d1wilko (approved):
LGTM :)

CubicNitrous:
@eoinmcafee00 @d1wilko looks like the version string in /version was not updated with the 2.7.1 release. https://github.com/harness/drone/blob/master/version/version.go#L30 this seems to get missed frequently

PR #3178, opened 2022-01-07
(feat) ignore archive repos on sync

eoinmcafee00:
The following providers - github, gitlab & gitee if a repository has been archive it will not longer show up in drone on the next sync


tphoney (approved):

Espoir999:
gitee repo only a part is shown。I have over a hundred repositories, but only a few dozen are shown。
gitee 仓库显示不全，只显示了一部分仓库。我有100多个仓库，但是drone上只有几十个，点击sync按钮也没有效果

What is the reason for this please?
请问下这是什么原因呢

PR #3196, opened 2022-03-14
ignore nil repos in list and add better debugging

d1wilko:


service/syncer/syncer.go
```diff
@@ -67,6 +68,7 @@ func (s *Synchronizer) Sync(ctx context.Context, user *core.User) (*core.Batch,
 		if err := recover(); err != nil {
 			logger = logger.WithField("error", err)
 			logger.Errorln("syncer: unexpected panic")
+			debug.PrintStack()
```
marko-gacesa, line 71: Maybe use the same logger to output the stack trace too, just to make sure the stack trace ends up in the same place as the rest of the logs.
```suggestion
			logger.Errorf("syncer: unexpected panic\n%s\n", debug.Stack())
```

d1wilko: 👍  updated now :)

marko-gacesa (approved):

eoinmcafee00 (approved):

PR #3203, opened 2022-04-22
bump go-scm version to v1.21.0

d1wilko:


tphoney (changes requested):
@d1wilko can you put the version you are changing to, in the commit message

PR #3205, opened 2022-04-24
Fix wrong address being transferred to local http server (non https/SSL) Port was forwarded instead

hube12:
When using a localhost installation (see configuration below), the http server is binded on :port which is wrong on windows distribution for instance and do not respect the HOST directive.

## Intent
Fix a bug when using a localhost installation

Configuration:
DRONE_SERVER_PROTO=http
DRONE_SERVER_HOST=127.0.0.1:3000
DRONE_SERVER_PORT=3000

## Why this bug was first introduced

I assume that this bug was introduced due to a kubernetes bug when the instance could not be realisatically reflected with the simple HOST, however having two field like SERVER_HOST and SERVER_PORT are confusing when SERVER_HOST accept DNS name and ip_address[:port], this case should be clarified as currently only DRONE_SERVER_PORT is used when using local binding.

### DIFF
--- a/cmd/drone-server/inject_server.go
+++ b/cmd/drone-server/inject_server.go
@@ -116,7 +116,7 @@ func provideRPC2(m manager.BuildManager, config config.Config) rpcHandlerV2 {
 func provideServer(handler *chi.Mux, config config.Config) *server.Server {
 	return &server.Server{
 		Acme:    config.Server.Acme,
-		Addr:    config.Server.Port,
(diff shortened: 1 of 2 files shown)


bradrydzewski:
The `Addr` is the public address of the server and is used to create webhooks.  The port is the port to which it binds.  These are not necessarily the same thing, especially if you install Drone behind a reverse proxy. I understand the mismatch in variable names may be confusing, but this is working as expected.

PR #3206, opened 2022-04-24
Fix subrepositories by using the oldest trick in the book (url encode it when syncing)

hube12:
When subrepositories are created on gitlab for instance, those are not shown.

## Intent
Fix subrepositories not showing in the UI and not being accessible via an URL

## Fix
Encode the namespace X/Y as X%2FY in the url.
As this is done on the syncing portion, this is only done once ever and will never be duplicated.
The url is thus addressable as the namespace route as a single element.
There is no collisions that could happen as even namespace with %2F would be invalid in any SCM (Choose a group path that does not start with a dash or end with a period. It can also contain alphanumeric characters and underscores. on gitlab). 

Collisions could happen on non standard SCM but as NONE are supported in your listing I don't see that as an issue (if it would be the case, a simple hash appending to the namespace would solve it)

### DIFF
--- a/service/syncer/syncer.go
+++ b/service/syncer/syncer.go
@@ -104,14 +104,7 @@ func (s *Synchronizer) Sync(ctx context.Context, user *core.User) (*core.Batch, 
 			return nil, err
 		}
 		for _, repo := range repos {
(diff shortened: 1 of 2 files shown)


bot2-harness:

PR #3210, opened 2022-05-06
(dron-267) correctly set parent for promotion retry

tphoney:
Working 
image
vs broken 
image

### DIFF
--- a/handler/api/repos/builds/retry.go
+++ b/handler/api/repos/builds/retry.go
@@ -64,6 +64,7 @@ func HandleRetry(
 		}
 
 		hook := &core.Hook{
+			Parent:       prev.Number,
 			Trigger:      user.Login,
 			Event:        prev.Event,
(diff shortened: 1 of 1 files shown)


eoinmcafee00 (approved):

handler/api/repos/builds/retry.go
```diff
@@ -64,6 +64,7 @@ func HandleRetry(
 		}
 
 		hook := &core.Hook{
+			Parent:       prev.ID,
```
bradrydzewski, line 67: this should be `prev.Number`

bradrydzewski (changes requested):

handler/api/repos/builds/retry.go
tphoney: great catch, i had a mind fart obviously on friday afternoon.
Made the changes

PR #3234, opened 2022-06-21
Add PR_TITLE env for pr builds

alikhil:
It's an attempt to add DRONE_PULL_REQUEST_TITLE env support.

Ideas was described [here](https://community.harness.io/t/add-support-of-drone-pull-request-title-env-variable-during-drone-build/12167).

### DIFF
--- a/operator/runner/env.go
+++ b/operator/runner/env.go
@@ -112,29 +112,30 @@ func buildEnviron(build *core.Build) map[string]string {
 		// these are legacy configuration parameters for backward
 		// compatibility with drone 0.8.
 		//
-		"CI_BUILD_NUMBER":              fmt.Sprint(build.Number),
-		"CI_PARENT_BUILD_NUMBER":       fmt.Sprint(build.Parent),
-		"CI_BUILD_CREATED":             fmt.Sprint(build.Created),
-		"CI_BUILD_STARTED":             fmt.Sprint(build.Started),
-		"CI_BUILD_FINISHED":            fmt.Sprint(build.Finished),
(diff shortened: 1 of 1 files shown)


operator/runner/env.go
```diff
@@ -112,29 +112,30 @@ func buildEnviron(build *core.Build) map[string]string {
...
+		"CI_COMMIT_AUTHOR_NAME":   build.AuthorName,
+		"CI_COMMIT_AUTHOR_EMAIL":  build.AuthorEmail,
+		"CI_COMMIT_AUTHOR_AVATAR": build.AuthorAvatar,
 	}
 	if strings.HasPrefix(build.Ref, "refs/tags/") {
 		env["DRONE_TAG"] = strings.TrimPrefix(build.Ref, "refs/tags/")
 	}
 	if build.Event == core.EventPullRequest {
 		env["DRONE_PULL_REQUEST"] = re.FindString(build.Ref)
+		env["DRONE_PULL_REQUEST_TITLE"] = re.FindString(build.Title)
```
alikhil, line 138: I think it should work. Since build.Title is [taken from PR title](https://github.com/harness/drone/blob/47886d424a6d93037c94a9289121a14671c537f0/service/hook/parser/parse.go#L273)

bradrydzewski:
👋 hey there, thanks for the pull request.  This code is no longer used and was moved to a separate library.  Therefore, if you want to add a new environment variable, it would need to be added to the library, in this file <https://github.com/drone/runner-go/blob/master/environ/environ.go>

alikhil:
Oh, cool then!

PR #3241, opened 2022-07-14
support time zone

zc2638:
When I set other time zone it always doesn't take effect, tzdata is missing in alpine image.

### DIFF
--- a/docker/Dockerfile.agent.linux.amd64
+++ b/docker/Dockerfile.agent.linux.amd64
@@ -1,5 +1,5 @@
 FROM alpine:3.11 as alpine
-RUN apk add -U --no-cache ca-certificates
+RUN apk add -U --no-cache ca-certificates tzdata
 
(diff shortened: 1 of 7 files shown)


tphoney:
Do you set the timezone after with an environment variable, like this ?

```
-e TZ=Europe/London
```

zc2638:
> Do you set the timezone after with an environment variable, like this ?
> 
> ```
> -e TZ=Europe/London
> ```

Yes.

tphoney:
tested this locally, works well. 

Many thanks @zc2638  🥂

tphoney:
i added docs as well https://github.com/drone/docs/pull/577

TankNee:
I meet the same issue in latest linux amd64 docker image, I have set the TZ environment variable in my docker-compose.yml file and it doesn't work correctly.
image
The following is drone page time and my system time.
image

PR #3249, opened 2022-08-08
feat: update drone-yaml module

jimsheldon:
the latest version of the drone-yaml module adds support for ppc64le and s390x
architectures

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/drone/drone-go v1.7.2-0.20220308165842-f9e4fe31c2af
 	github.com/drone/drone-runtime v1.1.1-0.20200623162453-61e33e2cab5d
 	github.com/drone/drone-ui v2.8.2+incompatible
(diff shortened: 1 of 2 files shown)


bradrydzewski:
@sdarwin in order to merge those pull requests we would need to build and publish the docker image from ppc64le and s390x machines. We do not have the required hardware at this time.  So your best option is to build and publish custom ppc64le and s390x clone images, and then configure the runner to use those custom images: 
https://docs.drone.io/runner/docker/configuration/reference/drone-runner-clone-image/

sdarwin:
@bradrydzewski  Wow that's cool, I did not see the setting DRONE_RUNNER_CLONE_IMAGE.   Will try it out.

> We do not have the required hardware at this time.

Not the case.  Those pull requests which I sent over use buildx which is an emulation layer and allow building alternative architecture docker images.  There are two choices, either thegeeklab/drone-docker-buildx method, or the vm method.  Both use standard amd64 linux and buildx to create the docker images.

PR #3261, opened 2022-09-07
update discourse.drone.io to community.harness.io

kit101:
some urls are too old to access.


kit101:
I created a drone server in Kubernetes and got an error message, but the URL could not be accessed. So I updated them.

image

tphoney (approved):

PR #3263, opened 2022-09-12
Enhance status check label for promotions

michelangelomo:
This PR introduces a change for labels applied to status checks related to promotions, actually a generic one is used.

### DIFF
--- a/service/status/status.go
+++ b/service/status/status.go
@@ -84,7 +84,7 @@ func (s *service) Send(ctx context.Context, user *core.User, req *core.StatusInp
 	_, _, err = s.client.Repositories.CreateStatus(ctx, req.Repo.Slug, req.Build.After, &scm.StatusInput{
 		Title:  fmt.Sprintf("Build #%d", req.Build.Number),
 		Desc:   createDesc(req.Build.Status),
-		Label:  createLabel(s.name, req.Build.Event),
+		Label:  createLabel(s.name, req.Build.Event, req.Build.Deploy),
 		State:  convertStatus(req.Build.Status),
 		Target: fmt.Sprintf("%s/%s/%d", s.base, req.Repo.Slug, req.Build.Number),
 	})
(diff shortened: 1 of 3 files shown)


michelangelomo:
hi @bradrydzewski, we would like to use this change asap, could you drop a review? :pray:

tphoney:
I had a look at the change, and this seems to be additive so it shouldnt affect any existing functionality.  I will defer to Brad, as i have not looked at the lifecycle of statuses in depth

aleinside:
hi @tphoney, any feedback from @bradrydzewski? 🙏

michelangelomo:
@tphoney @bradrydzewski bump

bradrydzewski:
The `deploy` value is freeform and may contain special characters, spaces, etc.  Do we know that all providers (github, gitlab, gitea, bitbucket, stash, azure) support freeform text and no character restrictions?  If yes, then we are good to merge.  If no, I would recommend taking the slug value (using the existing slug library that we already import).

```diff
import (
	"fmt"

	"github.com/drone/drone/core"
	"github.com/drone/go-scm/scm"

+	github.com/gosimple/slug
)
```

```diff
	case core.EventPromote:
-		return fmt.Sprintf("%s/promote/%s", name, deployTo)
+		return fmt.Sprintf("%s/promote/%s", name, slug.Make(deployTo))
```

michelangelomo:
> The deploy value is freeform and may contain special characters, spaces, etc. @tphoney do we know that all providers (github, gitlab, gitea, bitbucket, stash, azure) support freeform text and no character restrictions? If yes, then we are good to merge. If no, I would recommend taking the slug value (using the existing slug library that we already import).

hey @bradrydzewski, I think your suggestion is reasonable even if may there are no special character, better safe than sorry :smile:

SharpEdgeMarshall:
Any update on this?

d1wilko (approved):
LGTM - thanks for the change @michelangelomo - apologies for the delay

PR #3276, opened 2022-10-19
Skip check for runner architecture when item has arch="any"

jnohlgard:
Allow using the special value "any" for the pipeline target platform architecture to bypass the checks for runner architecture when scheduling runs.

Use case: I have two servers, one amd64, one arm64, each with their own Drone runner instance connected to my Drone server. Some pipelines have no architecture dependent parts, for example scripts, as long as there is a docker image with the correct architecture that the runner can use.

If this is merged it will need a paragraph in the docs somewhere in the section about pipeline platform, but I didn't find the right place to edit that yet.

### DIFF
--- a/scheduler/queue/queue.go
+++ b/scheduler/queue/queue.go
@@ -175,7 +175,7 @@ func (q *queue) signal(ctx context.Context) error {
 				if w.os != item.OS {
 					continue
 				}
-				if w.arch != item.Arch {
+				if item.Arch != "any" && w.arch != item.Arch {
 					continue
 				}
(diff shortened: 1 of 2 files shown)


jnohlgard:
needs an update to drone-yaml to avoid linting errors when starting a pipeline with arch=any

colin-nolan:
Does anyone know what needs to be done to get this merged? The repo referenced by jnohlgard (https://github.com/drone/drone-yaml) is archived so presumably that doesn't need to be updated?

bot2-harness:

PR #3292, opened 2022-11-30
(bugfix) add build params to Memoize cache key

andrii-kasparevych:
This PR fixes the problem described [here](https://community.harness.io/t/bug-starlark-ignores-parameters/12901) in details.
In short, when caching the Starlark/JSONNET conversion results, the build parameters are ignored which is harmful when the parameters are used for pipeline generation.

### DIFF
--- a/plugin/config/memoize.go
+++ b/plugin/config/memoize.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"context"
+	"crypto/md5"
 	"fmt"
 
 	"github.com/drone/drone/core"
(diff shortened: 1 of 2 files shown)


bradrydzewski:
@andrii-kasparevych golang map iteration order is random which would result in different cache strings. We had previously added the `DRONE_CONVERT_PLUGIN_CACHE_SIZE` to help solve this problem;  you can set the value to `0` which would disable caching. Generally speaking there should be no problem disabling caching unless you are running > 5000 builds per hour.

andrii-kasparevych:
@bradrydzewski thanks for the hint. I think I have confused the 2 `memoise.go` files in different places: [here](https://github.com/harness/drone/blob/master/plugin/config/memoize.go) and [here](https://github.com/harness/drone/blob/master/plugin/converter/memoize.go). 

I will try disabling the caching and see if that does the trick

andrii-kasparevych:
Disabling the caching worked for us, I will close the PR then.

PR #3319, opened 2023-05-14
+ sync gitea redirecturl config from gitee for customize login redire…

fireinice:
If the domain cannot visit from 80 or 443, the redirect url needs include the port. Sync the configuration item from gitee and and DRONE_GITEA_REDIRECT_URL` according to `DRONE_GITEE_REDIRECT_URL`


tphoney (approved):
Thanks for the PR @fireinice great work !

PR #3329, opened 2023-07-04
fix scheduler queue deadlock

marko-gacesa:
Fixes a rare server deadlock issue in the scheduler/queue. The deadlock happens when a runner asks for a stage and then immediately aborts (cancels the context). If the server successfully fetched a list of stages from the DB during that short time (registering a worker and canceling the context), it will try to give the runner a stage - while locked under a global queue mutex. And the runner side will try to remove itself from the list of runners while trying to obtain the lock on the same mutex.

Steps to test:
> go test -race github.com/drone/drone/scheduler/queue


tphoney:
closing in favour of #3330

PR #3397, opened 2023-10-02
Feature: Helm kubernetes Deployment

alexohneander:
Added helm deployment so you can quickly deploy it to your Kubernetes cluster.

The deployment consists of:
 - Service
 - Deployment
 - PVC
 - Ingress
 - ServiceAccount


charts/gitness/templates/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+          volumeMounts:
+            - name: datadir
+              mountPath: /data
+          livenessProbe:
+            httpGet:
+              path: /
+              port: http
+          readinessProbe:
+            httpGet:
+              path: /
```
abhinav-harness, line 49: We can add `api/v1/system/version` as endpoint for health and readiness check

charts/gitness/templates/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+        {{- include "gitness.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "gitness.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
```
abhinav-harness, line 30: We will probably need dind-daemon to run for us to support running CI too. I tested it like this 

```        - name: dind-daemon
          image: docker:24.0.3-dind
          command: ["dockerd", "--host", "tcp://127.0.0.1:2375"]
          resources:
            requests:
              cpu: 20m
              memory: 512Mi
          lifecycle:
            postStart:
              exec:
                command:
                  - "/bin/sh"
                  - "-c"
                  - "echo '127.0.0.1 host.docker.internal' >> /etc/hosts"
          securityContext:
            privileged: true
          volumeMounts:
            - name: docker-data
              mountPath: /certs
```

also added 
```
          env:
            - name: DOCKER_HOST
              value: tcp://localhost:2375
```

in gitness container

alexohneander: @abhinav-harness Do you want to add this as 2 containers or as a second pod that can then be accessed from gitness?

charts/gitness/templates/deployment.yaml
alexohneander: I have added these probes. Of course, it is much better if we request an endpoint.

arya-harness:
It would be good to add a NOTES.txt file for post install/upgrade steps. eg. Where to access the endpoint.

charts/gitness/templates/deployment.yaml
alexohneander: @abhinav-harness I added it. The container can be deactivated via the Values files and the image can be exchanged via it.

alexohneander:
@arya-harness I have added. This is actually created but I had to add with `git add -f`.

willzhang:
thanks for your work, can you add env to helm charts, i want define password when install
```
  -e GITNESS_PRINCIPAL_ADMIN_PASSWORD=gitness \
  -e GITNESS_USER_SIGNUP_ENABLED=false \
```

and the values.yaml maybe like this
```
env: {}

extraEnvVars: []
```

alexohneander:
Why is this PR Closed?

PR #3399, opened 2023-10-02
fix-path-traversal-vulnerability

DharunKumar04:
This pull request addresses a path traversal vulnerability in the codebase, as reported by Snyk. The vulnerability arises from unvalidated user input being used as a path in the `fs.readFileSync` function, potentially allowing an attacker to read arbitrary files.

Changes Made

- Implemented input validation and sanitization to prevent path traversal attacks.
- Enforced strict file path handling to avoid unintended access to sensitive files.
- Enhanced security checks to safeguard against unauthorized access.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -28,7 +28,7 @@
     "check:all": "npm-run-all typecheck lint prettier test",
     "clean": "rm -rf dist && rm -rf node_modules/.cache",
     "fmt": "prettier --write \"./src/**/*.{ts,tsx,css,scss}\"",
-    "services": "restful-react import --config restful-react.config.js code",
+    "services": "restful-react import --config restful-react.config.js --config code",
     "postservices": "prettier --write src/services/**/*.tsx",
     "strings": "npm-run-all strings:*",
     "strings:genTypes": "node scripts/strings/generateTypesCli.mjs"
--- a/web/package.json
+++ b/web/package.json
@@ -28,7 +28,7 @@
     "check:all": "npm-run-all typecheck lint prettier test",
     "clean": "rm -rf dist && rm -rf node_modules/.cache",
     "fmt": "prettier --write \"./src/**/*.{ts,tsx,css,scss}\"",
-    "services": "restful-react import --config restful-react.config.js --config code",
+    "services": "restful-react import --config restful-react.config.js",
     "postservices": "prettier --write src/services/**/*.tsx",
     "strings": "npm-run-all strings:*",
     "strings:genTypes": "node scripts/strings/generateTypesCli.mjs"
--- a/web/scripts/swagger-transform.js
+++ b/web/scripts/swagger-transform.js
@@ -14,61 +14,50 @@
  * limitations under the License.
  */
 
-const fs = require('fs');
-const path = require('path');
-const _ = require('lodash');
-const yaml = require('js-yaml');
-const stringify = require('fast-json-stable-stringify');
-
-// Define a mapping of valid transform modules
-const transformModules = {
-  'config1': require('./src/services/config1/transform.js'),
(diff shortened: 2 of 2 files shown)


web/scripts/swagger-transform.js
tan-nhu, line 1: Thanks, @DharunKumar04, for the PR. It seems a bit over-engineered.

Inputs for this script are specified in package.json (`"services": "restful-react import --config restful-react.config.js code"`). These inputs are internal, fixed, and not user-defined, reducing the risk of exposure.

If we really want to consider this a vulnerability, and the inputs in package.json are somehow used as an attack vector, we can simply hard-code `config` as `code` in this file. No need for the `transformModules` map, as it would also require hard-coding to map to `code` anyway.

```ts
const overridesFile = path.join('src/services', 'code', 'overrides.yaml')
const transformFile = path.join('src/services', 'code', 'transform.js')
```

DharunKumar04: Thank you for your feedback, @tan-nhu . I appreciate your input on the pull request.

I understand your point about the inputs for this script being specified in package.json and the reduced risk of exposure. While the inputs are indeed internal and fixed, we wanted to ensure a robust and flexible solution that can adapt to potential changes in the future.

The use of the `transformModules` map makes the code more reusable. It enables us to add new configurations without modifying the main logic of the script, promoting code maintainability.

However, I appreciate your suggestion to simplify the code by hard-coding the '**code**' configuration directly in the script. If the consensus leans toward this simpler approach, we can certainly make that adjustment. Please feel free to provide your final thoughts on this, and we can proceed accordingly.

Once we have a clear direction, we can either continue with the current approach or implement the suggested change and close this PR.

Thank you for your valuable input and consideration.

tan-nhu: @DharunKumar04 Let's change `config` to `code`. The change should be minimal. Currently Gitness UI connects to a single backend, the script was designed to support multiple ones. AFAIK, we won't have the second set of API in the short/medium term.

Also `restful-react` is being deprecated. We will adopt another library soon or later.

By the way, if you submit another PR. Please run `prettier` against the changed file(s). Thank you!

DharunKumar04:
@tan-nhu  Thanks for your feedback and for considering the change. I've updated the **configuration as 'code'** as you suggested to simplify the script. This change aligns with the current usage pattern and reduces potential complexity.

Regarding the use of restful-react, we'll keep an eye on its deprecation and plan to transition to the new library when needed.and thanks for the input !!

I've also ensured that Prettier has been run against the changed files as requested.

If you have any further suggestions or concerns, please feel free to let me know. Your input is greatly appreciated!

tan-nhu (approved):

web/package.json
```diff
@@ -28,7 +28,7 @@
     "check:all": "npm-run-all typecheck lint prettier test",
     "clean": "rm -rf dist && rm -rf node_modules/.cache",
     "fmt": "prettier --write \"./src/**/*.{ts,tsx,css,scss}\"",
-    "services": "restful-react import --config restful-react.config.js code",
+    "services": "restful-react import --config restful-react.config.js --config code",
```
tan-nhu, line 31: Please remove `--config code`

tan-nhu (changes requested):

web/package.json
DharunKumar04: Its done @tan-nhu
Waiting for your approval .

tan-nhu (approved):

DharunKumar04:
Hi @hitesharinga, the PR has been approved by @tan-nhu. Please merge it so that we can close this. Thank you.

hitesharinga:
Hey @DharunKumar04, Thank you for the Fix. I have merged it.

PR #3407, opened 2023-10-06
fix: Set 'Secure' attribute to true for sensitive cookies

DharunKumar04:
This pull request addresses the security issue raised by **Snyk** regarding sensitive cookies in the account package not having the '`Secure`' attribute set. The 'Secure' attribute is essential for ensuring that cookies are transmitted securely over HTTPS, preventing potential man-in-the-middle attacks.

**Changes Made:**
- Added the 'Secure' attribute to cookies in the '`includeTokenCookie`' and '`deleteTokenCookieIfPresent`' functions.
- The '`Secure`' attribute is set to true when the request is made over HTTPS, ensuring that sensitive cookies are only sent over **encrypted connections**.

### DIFF
--- a/app/api/handler/account/cookie.go
+++ b/app/api/handler/account/cookie.go
@@ -26,20 +26,18 @@ func includeTokenCookie(
 	r *http.Request, w http.ResponseWriter,
 	tokenResponse *types.TokenResponse,
 	cookieName string,
+	enforceSecure bool, // Add this parameter
 ) {
-	cookie := newEmptyTokenCookie(r, cookieName)
+	cookie := newEmptyTokenCookie(r, cookieName, enforceSecure) // Include enforceSecure
 	cookie.Value = tokenResponse.AccessToken
 	if tokenResponse.Token.ExpiresAt != nil {
 		cookie.Expires = time.UnixMilli(*tokenResponse.Token.ExpiresAt)
 	}
 
-	// Set the Secure attribute to true
-	cookie.Secure = true
-
 	http.SetCookie(w, cookie)
 }
 
-func deleteTokenCookieIfPresent(r *http.Request, w http.ResponseWriter, cookieName string) {
+func deleteTokenCookieIfPresent(r *http.Request, w http.ResponseWriter, cookieName string, enforceSecure bool) {
 	// if no token is present in the cookies, nothing todo.
 	// No other error type expected here - and even if there is, let's try best effort deletion.
 	_, err := r.Cookie(cookieName)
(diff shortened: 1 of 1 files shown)


app/api/handler/account/cookie.go
```diff
@@ -33,6 +33,9 @@ func includeTokenCookie(
 		cookie.Expires = time.UnixMilli(*tokenResponse.Token.ExpiresAt)
 	}
 
+	// Set the Secure attribute to true
+	cookie.Secure = true
```
enver-bisevac, line 37: A cookie with the Secure attribute is only sent to the server with an encrypted request over the HTTPS protocol.
[MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
why we enforce true here?, bcoz in newEmptyTokenCookie there is already isSecure value.

app/api/handler/account/cookie.go
```diff
@@ -48,16 +51,21 @@ func deleteTokenCookieIfPresent(r *http.Request, w http.ResponseWriter, cookieNa
 	cookie.Value = ""
 	cookie.Expires = time.UnixMilli(0) // this effectively tells the browser to delete the cookie
 
+	// Set the Secure attribute to true
+	cookie.Secure = true
```
enver-bisevac, line 55: same as above ^^^

app/api/handler/account/cookie.go
```diff
@@ -48,16 +51,21 @@ func deleteTokenCookieIfPresent(r *http.Request, w http.ResponseWriter, cookieNa
...
+	
 	http.SetCookie(w, cookie)
 }
 
 func newEmptyTokenCookie(r *http.Request, cookieName string) *http.Cookie {
+	isSecure := r.TLS != nil // Check if the request is using HTTPS
```
enver-bisevac, line 61: 👍

DharunKumar04:
Hi @enver-bisevac 

Thank you for your feedback and the reference to the [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies) documentation on cookies.As you mentioned, a cookie with the '`Secure`' attribute is sent to the server only with an encrypted request over the HTTPS protocol.I understand the concern and would like to clarify my approach The reason for enforcing the 'Secure' attribute as true in the `includeTokenCookie` and `deleteTokenCookieIfPresent` functions is to provide an **explicit** security measure. Since the authentication token cookie lacks the 'Secure' attribute, it maybe sent in plaintext over an unencrypted HTTP connection.And attackers can use this vulnerability as an attack surface.

By setting the secure attribute to true it serves as a clear indication that these cookies should only be transmitted over HTTPS, aligning with the best practices outlined in the MDN documentation.And also explicitly setting '`Secure`' within these functions, we adopt a defensive coding approach. This ensures that even if the **isSecure** variable in the **newEmptyTokenCookie** function were to change in the future due to code evolution, our security measure remains intact.

If you believe it's unnecessary for our codebase or if there are better ways to handle this, please let me know, or we can close this pull request.Thanks :)

app/api/handler/account/cookie.go
johannesHarness, line 61: Not sure how `r.TLS` is fixing anything specific, could you elaborate?
One scenario I'm thinking of improving is gitness behind a reverse proxy with TLS offloading - in that case `r.TLS` would be `nil` and `r.URL.Proto` would be http.

So an improvement would be to check for the availability of a `X-Forwarded-Proto` header (otherwise fall back to `r.URL.Proto` or `r.TLS`) to determine whether `cookie.Secure` should be set?

cc @enver-bisevac

enver-bisevac: I agree @johannesHarness

DharunKumar04:
Thank you, @enver-bisevac and @johannesHarness, for your valuable feedback and for engaging in this discussion. I'd like to clarify why I initially used `r.TLS` and how we can adapt the code for better flexibility while addressing the concerns raised.

The reason for using `r.TLS` in the code was to ensure that the '`Secure`' attribute of cookies aligns with the actual security context of the incoming request. When `r.TLS `is **non-nil**, it indicates that the request is made over an encrypted HTTPS connection. In this scenario, setting `cookie.Secure` to true is in line with security best practices, as we want to enforce the use of secure cookies over HTTPS connections.

And thank you, @johannesHarness, for your valuable input and raising the scenario related to reverse proxies with **TLS offloading**. Your feedback highlights an important consideration that we need to account for in our code.

You correctly pointed out that in such a scenario, `r.TLS` might be nil, and `r.URL.Proto` could be "`http`" potentially leading to **`cookie.Secure = false`** even when the caller is using HTTPS. To address this, I propose the following improvement:

```
func newEmptyTokenCookie(r *http.Request, cookieName string, enforceSecure bool) *http.Cookie {
    isSecure := false

    // Check for the X-Forwarded-Proto header
    if proto := r.Header.Get("X-Forwarded-Proto"); proto == "https" {
        isSecure = true
    } else if r.TLS != nil || r.URL.Scheme == "https" {
        isSecure = true
    }

    // Enforce 'Secure' based on the provided flag
    if enforceSecure {
        isSecure = true
    }

    return &http.Cookie{
        Name:     cookieName,
        SameSite: http.SameSiteStrictMode,
        HttpOnly: true,
        Path:     "/",
        Domain:   r.URL.Hostname(),
        Secure:   isSecure,
    }
}
```

I'd also like to acknowledge and apologize for any inconvenience caused by the previous implementation of **`cookie.Secure = true`** without considering that  this will actually break gitness running on **http** i removed those implementation in my latest commit

PR #3408, opened 2023-10-07
podman host is different

ppphp:


johannesHarness:
@vistaarjuneja I agree - we should be driving the value from `GITNESS_URL_CONTAINER` and push that down.

As it is right now, the host used by containers is hard-coded in multiple places. In case the user has some custom setup, neither might work, and they can't change it.

Thus, I think we should set the `GITNESS_URL_CONTAINER` based on the container runtime (if it's not explicitly provided by the user) - and then all other values we should derive from that. In other words, we shouldn't set the host / protocol anywhere (as it is right now)

app/pipeline/manager/client.go
```diff
@@ -103,7 +113,7 @@ func (e *embedded) Detail(ctx context.Context, stage *drone.Stage) (*client.Cont
 		Netrc:   ConvertToDroneNetrc(details.Netrc),
 		System: &drone.System{
 			Proto: e.config.Server.HTTP.Proto,
-			Host:  "host.docker.internal",
+			Host:  getContainerHost(),
```
johannesHarness, line 116: as mentioned, derive both the `proto` and the `host` from `e.config.URL.Container`, and move the `getContainerHost` logic to `backfillURLs` in `cli/server/config.go`

ppphp: > as mentioned, derive both the `proto` and the `host` from `e.config.URL.Container`, and move the `getContainerHost` logic to `backfillURLs` in `cli/server/config.go`

fixed now

app/pipeline/runner/runner.go
```diff
@@ -45,7 +45,7 @@ func NewExecutionRunner(
 ) (*runtime2.Runner, error) {
 	// For linux, containers need to have extra hosts set in order to interact with
 	// the gitness container.
-	extraHosts := []string{"host.docker.internal:host-gateway"}
+	extraHosts := []string{config.URL.Container + ":host-gateway"}
```
vistaarjuneja, line 48: Hi @ppphp , this will break since `config.URL.Container` is a full URL (eg `http://host.docker.internal:3000`) and not `host.docker.internal`. You can check `https://github.com/harness/gitness/blob/main/cli/server/config.go#L89` this is where we create the container URL.
We should only have `host.docker.internal` or `host.containers.internal` here

app/pipeline/manager/client.go
```diff
@@ -103,7 +103,7 @@ func (e *embedded) Detail(ctx context.Context, stage *drone.Stage) (*client.Cont
 		Netrc:   ConvertToDroneNetrc(details.Netrc),
 		System: &drone.System{
 			Proto: e.config.Server.HTTP.Proto,
-			Host:  "host.docker.internal",
+			Host:  e.config.URL.Container,
```
vistaarjuneja, line 106: Same here

app/pipeline/runner/runner.go
johannesHarness: you can add a new method to the `url.Provider` (similar to `GetAPIHostname() string`) where you return the container host.

ppphp: thank you for your advice, but I pushed a simple way. If new commit is wrong, please close it~

vistaarjuneja:
hi @ppphp this looks good to me, there are just some minor lint issues. Would be great if you could fix those and update the PR.

ppphp:
> hi @ppphp this looks good to me, there are just some minor lint issues. Would be great if you could fix those and update the PR.

merge main now, sorry for my late

PR #3411, opened 2023-10-11
Update readme.md

Luwa-Tech:
In the "Community and Support" section, i replaced the line that mentions the `community Slack`  with a note explaining that `a @harness.io address is required to join the Slack community`


bradrydzewski:
the Slack invitation link expired which is what caused the problem.  the team generated a new link, which will allow joining Slack without a harness address.  this change should be reflected here https://docs.gitness.com/support

PR #3417, opened 2023-10-19
Feature: Helm kubernetes Deployment 

alexohneander:
Added helm deployment so you can quickly deploy it to your Kubernetes cluster.

The deployment consists of:
 - Service
 - Deployment
 - PVC
 - Ingress
 - ServiceAccount


alexohneander:
@willzhang @arya-harness @abhinav-harness 
Accidentally closed the pull request during the last commit and could not reopen it.
It is the same request as #3397 with the additional env vars.

abhinav-harness (approved):

charts/gitness/templates/hpa.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
```
abhinav-harness, line 4: HPA isn't supported due to the code's lack of a distributed architecture. Can we remove this file?

willzhang:
I found that the Docker Dind section has been deleted. Is this not needed, or is there any other solution？

```
docker:
  enabled: true
  image: docker:24.0.3-dind
  resources: {}
    # limits:
    #   cpu: 100m
    #   memory: 128Mi
    # requests:
    #   cpu: 100m
    #   memory: 128Mi
```

or can we mount host docker.sock or containerd.sock ? Uncertain whether gitness supports containerd

```
host:
  docker:
    sock: /var/run/docker.sock
  containerd:
    sock: /var/run/containerd/containerd.sock
volumes:
```
Perhaps in the k8s environment, gigness directly communicates with the k8s API and runs jobs by creating multiple work pods？

bradrydzewski:
good point, we would want a `docker:dind` container added for pipeline execution, similar to what we have in the official Drone help chart.  See https://github.com/drone/charts/blob/master/charts/drone-runner-docker/templates/deployment.yaml#L40

adding @jimsheldon as a reviewer (he is the creator of the Drone chart)

charts/gitness/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "gitness.fullname" . }}
+  labels:
+    {{- include "gitness.labels" . | nindent 4 }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  replicas: {{ .Values.replicaCount }}
```
arya-harness, line 9: currently multiple pods of gitness are not possible. We should hardcode this to 1.

charts/gitness/values.yaml
```diff
@@ -0,0 +1,107 @@
+# Default values for gitness.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: harness/gitness
+  pullPolicy: IfNotPresent
```
arya-harness, line 9: PullPolicy are advised be 'Always' if you are using latest tag. Please add a comment for users as well

alexohneander:
@willzhang  @bradrydzewski I made a mistake yesterday. We still need the dind and that definitely has to go back in. I'll adjust it.

charts/gitness/values.yaml
willzhang: > 如果您使用最新标签，建议 PullPolicy 设置为“始终”。还请用户添加评论

we have offline environment,  maybe should not `always` policy.

And maybe should with fixed labels and clear versions, this is the standard for production environments.

alexohneander: Done :)

charts/gitness/templates/hpa.yaml
alexohneander: Is removed :)

charts/gitness/templates/deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+            {{- end }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+        {{- if .Values.dind.enabled }}
+        - name: dind
+          image: {{ .Values.dind.image }}
+          command: ["dockerd", "--host", "tcp://127.0.0.1:2375"]
+          resources:
+            requests:
+              cpu: 20m
```
arya-harness, line 64: We should allow resources to be configurable(maybe something like dind.resources ), depending upon scale of usages.
@abhinav-harness any suggestion on how the dind resources should increase if we increase the resources on gitness container?

alexohneander: @arya-harness I have added the possibility. Have left it empty as default. Then the user can reduce or extend it himself.

willzhang (commented):
review

alexohneander:
Is there anything missing or can I tweak, I would love to do that. 😃

Lite5h4dow:
there is no index.yaml file for using these templates as a helm chart.

PR #3451, opened 2024-01-08
removing duplicated metrics for servers and swapping to pattern mathi…

maxknee:
…ng, since at scale we're seeing our metrics endpoint swell to over a MB


maxknee:
@vistaarjuneja can you merge this as well?

PR #3459, opened 2024-01-16
Enable http metrics

maxknee:
Adding config option to enable http metrics


vistaarjuneja (approved):

vistaarjuneja:
looks like the previous PR is leading to a high memory usage in Drone v2.22. Merging this to get the issue fixed in 2.23, thanks @maxknee

PR #3495, opened 2024-04-03
enable the settings link in order to delete empty repository

SchoIsles:
After create or import an empty repository, only the `files` link is enabled.
if the user wants to delete it, he must initiate a commit to make it no longer empty, and then he can enter the settings page to delete it.

Can we allow users to enter the settings page of an empty repository to delete it directly?


johannesHarness:
Thank you @SchoIsles for the contribution, that's great to see!
This seems to have been already added recently and will be released in the upcoming version :) 

For reference, here's the commit:
https://github.com/harness/gitness/commit/9b561a33cda1c8260aac4eb3c80e0da7a717e6ae

PR #3600, opened 2024-12-12
fix: [AH-784]: Updated URL provider to update casing for space names

arvind-choudhary-h:


app/url/provider.go
```diff
@@ -246,6 +246,9 @@ func (p *provider) RegistryURL(_ context.Context, params ...string) string {
 	}
 
 	segments := []string{u.Path}
+	if len(params) > 0 {
```
shivagowda, line 249: looks good.
params[0] shouldn't contain any space, but to make it robust, can consider:
```
if len(params) > 0  {
    params[0] = strings.ToLower(strings.TrimSpace(params[0]))
}
```

shivagowda (approved):

app/url/provider.go
arvind-choudhary-h: thanks. spaces will not be there in the account/space in any flow, it's safe.

PR #3613, opened 2025-01-28
Update deployment ID to bigint

maxknee:
We're using github deployments and the ID is out of range: 

```
{"message":"pq: value \"2148060640\" is out of range for type integer"}
```

### DIFF
--- a/store/shared/migrate/postgres/ddl_gen.go
+++ b/store/shared/migrate/postgres/ddl_gen.go
@@ -200,6 +200,10 @@ var migrations = []struct {
 		name: "create-new-table-cards",
 		stmt: createNewTableCards,
 	},
+	{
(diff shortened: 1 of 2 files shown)


maxknee:
I'll make those changes

vistaarjuneja:
hey @maxknee , curious how you tested this and on which version of postgres? On testing manually, this command doesn't seem to work for ALTER column and we need to instead set them separately (in postgres version 14):
```
ALTER TABLE builds 
ALTER COLUMN build_deploy_id SET DATA TYPE BIGINT,
ALTER COLUMN build_deploy_id SET NOT NULL,
ALTER COLUMN build_deploy_id SET DEFAULT 0;
```

dewan-ahmed:
Hi @maxknee please make changes based on the feedback. If there's no activity, we'll close the PR.

PR #3645, opened 2025-09-19
fix: pass standalone prop to ReviewerSelect to prevent usergroup API calls

ysicing:
image


abhinav-harness:
please fix the lint checks

PR #3655, opened 2025-12-28
OLEKSANDR69 Create SECURITY.md for security policy

sacha699612-hub:
Added a security policy document outlining supported versions and vulnerability reporting.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


rajpratham1 (changes requested):
I wouldn't approve this PR in its current state.

The file is the default GitHub template and still contains placeholder text instead of the project's actual security policy. Adding a SECURITY.md is valuable, but publishing placeholder information could confuse users about which versions are supported and how vulnerabilities should be reported.

PR #3667, opened 2026-02-24
Adding support for individual stage restart via new endpoint

ioan-tataru:
**The issue:** https://github.com/drone/proposal/issues/28

**The goal:** more granular restarts so that flaky tests or internal failures don't require a full build restart

**Implementation details:**
Existing endpoint, extended with optional params:

POST /api/repos/{owner}/{name}/builds/{number}?stage=[pipeline-name]&cascade=[true|false]
- stage (optional) — name of a single pipeline to retry. If omitted, the whole build is retried (previous behavior).
- cascade (optional, only valid with stage) — when true, also retry every pipeline that transitively depends on stage. Defaults to false.

A 400 is returned if stage does not match any pipeline in the referenced build.

Tested via Go unit tests.


basedwdev:
this seems like a very good change that can help bring some costs down

PR #3688, opened 2026-05-09
fix: handle wildcard CODEOWNERS file matches

immanuwell:
Small CODEOWNERS matcher fix.

Patterns like `a.*` and `a.**` should match the file `a.` too. Before this, the matcher only hit the child-path form, so `a.` slipped through. tiny papercut, but kinda confusing ngl.

Repro before the fix:

```text
match("a.*", "a.") = false but wanted true
match("a.**", "a.") = false but wanted true
```

Fix is to check the normalized file pattern first, then fall through to the existing descendant match logic.

Verified with:

```bash
go test ./app/services/codeowners -count=1
```

I looked for related CODEOWNERS/doublestar issues and PRs, didn't find a matching one.

### DIFF
--- a/app/services/codeowners/service.go
+++ b/app/services/codeowners/service.go
@@ -628,12 +628,6 @@ func findReviewerInList(email string, uid string, reviewers []*types.PullReqRevi
 // Match matches a file path against the provided CODEOWNERS pattern.
 // The code follows the .gitignore syntax closely (similar to github):
 // https://git-scm.com/docs/gitignore#_pattern_format
-//
-// IMPORTANT: It seems that doublestar has a bug, as `*k/**` matches `k` but `k*/**` doesnt (incorrect)'.
-// Because of that, we currently match patterns like `test*` only partially:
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This looks like a solid fix. The PR removes the workaround comments for the doublestar limitation and instead performs an explicit file match before applying the directory matching logic. That makes wildcard patterns like a.* and a.** behave correctly while simplifying the implementation. The accompanying tests also re-enable the previously skipped cases, which is a good regression check.

GitHub Approval Comment

Nice improvement! This makes the matching behavior much closer to what users would expect from CODEOWNERS patterns.

I like that the implementation now performs an explicit file match before applying the existing directory matching logic. That removes the need for the previous workaround and allows wildcard file patterns such as a.* and a.** to work correctly without affecting folder matching behavior.

It's also great to see the previously disabled test cases re-enabled. Having regression tests for these wildcard scenarios gives confidence that the behavior won't accidentally break again in future changes.

Overall, this is a clean and focused fix with good test coverage.

PR #3682, opened 2026-04-16
Fix push protection bypass for plain API content writes

shaun0927:
## Summary

This PR restores the intended push-protection bypass behavior for API content writes.

The current pre-receive path treats plain API content writes as bypassed, even though the operation model still distinguishes:
- `GitOpTypeAPIContent`
- `GitOpTypeAPIContentBypassRules`
- `GitOpTypeGitPush`

This change restores bypass only for the explicit bypass operation and direct git pushes.

Fixes #3681

## Root cause

`app/api/controller/githook/pre_receive.go` widened `AllowBypass` during push-protection evaluation, so `GitOpTypeAPIContent` and `GitOpTypeAPIContentBypassRules` became equivalent in practice for push rules.

That no longer matched the rest of the operation-type model in:
- `app/api/controller/util.go`
- `app/api/controller/repo/commit.go`
- `types/enum/githook.go`

## Changes included

1. Restore scoped bypass behavior in `checkPushProtection()`:
   - allow bypass for `GitOpTypeAPIContentBypassRules`
   - allow bypass for `GitOpTypeGitPush`
   - do not bypass for plain `GitOpTypeAPIContent`
2. Add focused regression tests for both paths:
   - plain API content is not bypassed
   - explicit bypass API content is bypassed

## Validation

```bash
go test -v ./app/api/controller/githook -run 'TestCheckPushProtection_' -count=1 -timeout 20s
```

Validated locally:
- before the fix:
  - `TestCheckPushProtection_DoesNotBypassForPlainAPIContent` failed
  - `TestCheckPushProtection_DoesBypassForAPIContentBypassRules` passed
- after the fix:
  - both tests pass

## Scope

This PR is intentionally narrow:
- one controller-level bypass decision restored
- regression coverage added for the affected behavior only

PR #3713, opened 2026-08-14
fix: enforce push protection for plain API content writes

### LINKED ISSUE #3681: [BUG] Plain API content writes bypass push protection in pre-receive
## Summary

Plain API content writes are currently treated as bypassed during pre-receive push-protection evaluation.

The repository already distinguishes these operation types:
- `GitOpTypeAPIContent`
- `GitOpTypeAPIContentBypassRules`
- `GitOpTypeGitPush`

But `checkPushProtection()` now passes `AllowBypass=true` for all of them, which collapses that distinction for push-protection rules.

## Problem

In `app/api/controller/githook/pre_receive.go`, `checkPushProtection()` controls whether push rules are evaluated as bypassed.

At the moment, a plain API content operation (`GitOpTypeAPIContent`) is treated the same as an explicit bypass operation (`GitOpTypeAPIContentBypassRules`).

That means a bypassable push rule can be marked as bypassed even when the caller did **not** request bypass semantics.

## Why this looks incorrect

The operation-type model still explicitly separates plain API writes from explicit bypass writes:
- `app/api/controller/util.go`
- `app/api/controller/repo/commit.go`
- `types/enum/githook.go`

So the current pre-receive behavior no longer matches the contract implied by those operation types.

## Minimal reproduction

I added the following focused regression tests locally in `app/api/controller/githook/pre_receive_test.go`:
- `TestCheckPushProtection_DoesNotBypassForPlainAPIContent`
- `TestCheckPushProtection_DoesBypassForAPIContentBypassRules`

Running:

```bash
go test -v ./app/api/controller/githook -run 'TestCheckPushProtection_' -count=1 -timeout 20s
```

Current result on `main`:
- `TestCheckPushProtection_DoesNotBypassForPlainAPIContent` fails
- `TestCheckPushProtection_DoesBypassForAPIContentBypassRules` passes

Observed failing state for the plain API content case:
- `bypassable=true`
- `bypassed=true`

Expected for `GitOpTypeAPIContent`:
- `bypassable=true`
- `bypassed=false`

## Expected behavior

`AllowBypass` in `checkPushProtection()` should be limited to operation types that explicitly support bypass:
- `GitOpTypeAPIContentBypassRules`
- `GitOpTypeGitPush`

Plain `GitOpTypeAPIContent` should not be treated as bypassed.

## Impact

This can allow plain API content mutations (for example, commit-file style API writes) to bypass bypassable push-protection rules even when bypass was not explicitly requested.

abhisek343:
Fixes #3681

## Summary

- restrict push-protection bypass to direct Git pushes and API content operations that explicitly request bypass semantics
- keep plain API content writes subject to push-protection rules
- add regression coverage for plain API content, explicit API bypass, direct Git pushes, and other operation types

## Root cause

checkPushProtection passed AllowBypass: true for every pre-receive operation, so plain API content writes were treated like explicit bypass operations.

## Validation

- go test ./app/api/controller/githook (pass)
- go test ./app/services/protection (pass)
- make test (pass; excludes the repository’s documented registry conformance packages)
- git diff --check (pass)

The frontend dependencies were installed and yarn build completed successfully to provide the embedded web/dist assets required by the full Go test target.

### DIFF
--- a/app/api/controller/githook/pre_receive.go
+++ b/app/api/controller/githook/pre_receive.go
@@ -287,12 +287,9 @@ func (c *Controller) checkPushProtection(
 ) ([]types.RuleViolations, *repoSettingsViolations, error) {
 	pushProtection := c.protectionManager.FilterPushProtection(protectionRules)
 
-	// TODO: Once push rule violations are returned to the API layer,
-	// allow bypass only for operations that explicitly support it.
-	// Specifically:
-	//   - GitOpTypeAPIContentBypassRules: API operations that intentionally bypass rules
-	//   - GitOpTypeGitPush: direct git pushes from clients
-	allowBypass := true
+	// Only direct git pushes and API operations that explicitly request bypass
+	// semantics may bypass push protection rules.
+	allowBypass := allowPushProtectionBypass(in.OperationType)
 
 	pushVerifyOut, _, err := pushProtection.PushVerify(
 		ctx,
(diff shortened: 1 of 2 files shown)


abhisek343:
Closing this in favor of #3682. I missed the existing PR and the maintainer feedback around the API violation-propagation prerequisite. Thanks, and apologies for the duplicate.

